PR #41, opened 2026-03-18
This should reflect the body of work that WT has done on his fork

imajes:


obviouslyallie:
I think this PR should be changed in several places. Its probably a problem now that's getting worse every time develop has an update.

- Split into smaller PRs would be significantly easier. 
- The readme/docs/ci items don't need to be included. 
- Needs to target develop and not main.

zachyale:
At this point this changeset needs to be completely overhauled. I've spoken to WT and can implement the necessary B/E changes in some more targeted PRs if you want to hand this off, but as it stands this PR is effectively just a draft.

Closing until it can be reworked and is in a place where it's ready for review

PR #66, opened 2026-03-19
chore(ci): rebrand and streamline the release pipeline

imajes:
- CI/CD:
  - replace the old `develop`/`master` pipelines with explicit validation, nightly, preview, and release workflows
  - switch stable releases to `main`, keep `develop` for nightly builds, and make preview image builds manual-only
  - update the reusable migration check so callers pass explicit refs and validation works correctly across PRs, pushes, and release flows

- Packaging:
  - consolidate container packaging onto the root `Dockerfile` and remove `Dockerfile.ci`
  - stop mutating tracked source files during image builds by building frontend and backend assets inside the container pipeline
  - remove vendored `unrar` binaries and build the upstream CLI from source during the image build
  - keep runtime compatibility for existing Booklore deployments by preserving env vars, ports, mount paths, and legacy internal user defaults

- Deploy Surface:
  - move deployment examples under `deploy/` and Docker runtime assets under `packaging/docker`
  - rebrand image, chart, and helper-script references from Booklore to Grimmory on the build and release surfaces only
  - update Compose, Helm, and Podman examples to reflect Grimmory image names and the stable/nightly tag model

- Docs:
  - add the implementation plan under `docs/plans/ci-cd-rebrand-main-nightly.md`
  - refresh README, contributing guidance, PR template, and operational docs to match the new branch, image, and upgrade story
  - document the Booklore-to-Grimmory image swap path so existing self-hosted installs can upgrade by changing only the image reference

- Validation:
  - pass `git diff --check`
  - parse all workflow YAML successfully
  - pass `shellcheck` on the touched shell scripts

## Summary by CodeRabbit

* **New Features**
  * Added reusable CI workflows (PR validation, semantic PR title check, test suite, preview, nightly, semantic stable, release preview) and image publish flows.

* **Documentation**
  * Rebranded Booklore → Grimmory across docs, README, and deploy examples; added CI/CD and semantic-release plans and a new CHANGELOG.

* **Chores**
  * Consolidated Docker build, updated images/entrypoint, tightened dockerignore, added release tooling/scripts, deploy manifests, and Dependabot config.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -33,19 +33,7 @@ RUN set -eux; \
     jar_path="$(find build/libs -maxdepth 1 -name '*.jar' ! -name '*plain.jar' | head -n 1)"; \
     cp "$jar_path" /workspace/booklore-api/app.jar
 
-FROM alpine:3.22 AS unrar-build
-
-ARG UNRAR_VERSION=7.2.4
-ARG UNRAR_SHA256=b02e571a33af7711cd803080500370dc1d28eea82b2032480819d27462ad8b31
-
-RUN apk add --no-cache build-base curl tar
-
-WORKDIR /tmp
-
-RUN curl -fsSL "https://www.rarlab.com/rar/unrarsrc-${UNRAR_VERSION}.tar.gz" -o unrarsrc.tar.gz && \
-    echo "${UNRAR_SHA256}  unrarsrc.tar.gz" | sha256sum -c - && \
-    tar -xzf unrarsrc.tar.gz && \
-    make -C unrar
+FROM linuxserver/unrar:7.1.10 AS unrar-layer
 
 FROM eclipse-temurin:25-jre-alpine
 
@@ -59,7 +47,7 @@ LABEL org.opencontainers.image.title="Grimmory" \
       org.opencontainers.image.documentation="https://grimmory.org/docs/getting-started" \
       org.opencontainers.image.version=$APP_VERSION \
       org.opencontainers.image.revision=$APP_REVISION \
-      org.opencontainers.image.licenses="GPL-3.0" \
+      org.opencontainers.image.licenses="AGPL-3.0" \
       org.opencontainers.image.base.name="docker.io/library/eclipse-temurin:25-jre-alpine"
 
 ENV JAVA_TOOL_OPTIONS="-XX:+UseG1GC -XX:+UseCompactObjectHeaders -XX:+UseStringDeduplication -XX:MaxRAMPercentage=75.0 -XX:+ExitOnOutOfMemoryError"
@@ -71,7 +59,9 @@ RUN apk update && apk add --no-cache su-exec libstdc++ libgcc && \
 
 COPY packaging/docker/entrypoint.sh /usr/local/bin/entrypoint.sh
 RUN chmod +x /usr/local/bin/entrypoint.sh
-COPY --from=unrar-build /tmp/unrar/unrar /usr/local/bin/unrar
+
+COPY --from=unrar-layer /usr/bin/unrar-alpine /usr/local/bin/unrar
+
 COPY --from=backend-build /workspace/booklore-api/app.jar /app/app.jar
 
 ARG BOOKLORE_PORT=6060


Dockerfile
```diff
@@ -1,74 +1,79 @@
...
-      org.opencontainers.image.url="https://github.com/booklore-app/booklore" \
-      org.opencontainers.image.documentation="https://booklore.org/docs/getting-started" \
+LABEL org.opencontainers.image.title="Grimmory" \
+      org.opencontainers.image.description="Grimmory: a self-hosted, multi-user digital library with smart shelves, auto metadata, Kobo and KOReader sync, BookDrop imports, OPDS support, and a built-in reader for EPUB, PDF, and comics." \
+      org.opencontainers.image.source="https://github.com/grimmory-tools/grimmory" \
+      org.opencontainers.image.url="https://github.com/grimmory-tools/grimmory" \
+      org.opencontainers.image.documentation="https://grimmory.org/docs/getting-started" \
       org.opencontainers.image.version=$APP_VERSION \
       org.opencontainers.image.revision=$APP_REVISION \
       org.opencontainers.image.licenses="GPL-3.0" \
```
imnotjames: Is this correct?

Dockerfile
```diff
@@ -1,74 +1,79 @@
...
+
+ARG UNRAR_VERSION=7.2.4
+
+RUN apk add --no-cache build-base curl tar
+
+WORKDIR /tmp
+
+RUN curl -fsSL "https://www.rarlab.com/rar/unrarsrc-${UNRAR_VERSION}.tar.gz" -o unrarsrc.tar.gz && \
+    tar -xzf unrarsrc.tar.gz && \
+    make -C unrar
```
imnotjames: It might be beneficial to split this into different steps for clarity and caching, since it won't be in the final image.

You can also use `ADD` I think here and skip curl + get checksums?

```suggestion
ARG UNRAR_VERSION=7.2.4
ARG UNRAR_CHECKSUM=sha256:b02e571a33af7711cd803080500370dc1d28eea82b2032480819d27462ad8b31

RUN apk add --no-cache build-base tar

WORKDIR /tmp

ADD \
    --checksum="${UNRAR_CHECKSUM}" \
    "https://www.rarlab.com/rar/unrarsrc-${UNRAR_VERSION}.tar.gz"

RUN tar -xzf unrarsrc.tar.gz && \
    make -C unrar
```

Dockerfile
```diff
@@ -1,74 +1,79 @@
...
-COPY --from=angular-build /angular-app/dist/booklore/browser /springboot-app/src/main/resources/static
+COPY booklore-api/ ./
+COPY --from=frontend-build /workspace/booklore-ui/dist/booklore/browser /tmp/frontend-dist
 
-# Inject version into application.yaml using yq
-ARG APP_VERSION
-RUN apk add --no-cache yq && \
-    yq eval '.app.version = strenv(APP_VERSION)' -i /springboot-app/src/main/resources/application.yaml
+RUN mkdir -p build/resources/main/static && \
+    cp -r /tmp/frontend-dist/. build/resources/main/static/
```
imnotjames, line 27: I don't 100% understand - why copy to `/tmp/frontend-dist` and then run `mkdir` + `cp`?

Wouldn't `COPY --from=frontend-build /workspace/booklore-ui/dist/booklore/browser build/resources/main/static/` do the same thing?

Dockerfile
imajes: For now. i know it doesn't exist but we can handle it for now.

Dockerfile
imajes: typically we should avoid multiple run steps consecutively. but yes, we should add the checksum here, so that we know it's the right file. the downside is that this whole thing is an added dependency to deal with -- so i have a note to make a better decision on this tomorrow. thanks for flagging it!

Dockerfile
imajes: Yeah, that's a defensive move to ensure that the folder tree exists first. I have to do a fuller review of it all again now it's all changed up. alot of things are changing all at the same time, so.. 😓

README.md
```diff
@@ -86,17 +86,31 @@ MYSQL_DATABASE=grimmory
...
-    # Alternative: ghcr.io/grimmory-tools/grimmory:latest
+    image: grimmory/grimmory:v0.38.2
+    # Convenience tag:
+    # image: grimmory/grimmory:latest
+    # Alternative: ghcr.io/grimmory-tools/grimmory:v0.38.2
     # To build from source instead: comment out 'image' and uncomment below
     # build: .
     container_name: grimmory
     environment:
+      - APP_VERSION=v0.38.2
```
imnotjames: Why would you need to specify the `APP_VERSION` if you're pulling the image that's versioned?  What would this accomplish?

deploy/compose/docker-compose.yml
```diff
@@ -0,0 +1,60 @@
...
+  grimmory:
+    # Stable release example:
+    image: grimmory/grimmory:v0.38.2
+    # Convenience tag:
+    # image: grimmory/grimmory:latest
+    # Or the GHCR image:
+    # image: ghcr.io/grimmory-tools/grimmory:v0.38.2
+    container_name: grimmory
+    environment:
+      - APP_VERSION=v0.38.2
```
imnotjames: Why would we need to specify `APP_VERSION`?  This is injected into the docker image on build, isn't it?

Dockerfile
```diff
@@ -1,74 +1,79 @@
...
+
+FROM alpine:3.22 AS unrar-build
+
+ARG UNRAR_VERSION=7.2.4
+
+RUN apk add --no-cache build-base curl tar
+
+WORKDIR /tmp
+
+RUN curl -fsSL "https://www.rarlab.com/rar/unrarsrc-${UNRAR_VERSION}.tar.gz" -o unrarsrc.tar.gz && \
```
NamLe029: we can grab unrar artifacts from [linuxserver/unrar](https://hub.docker.com/r/linuxserver/unrar), there is no need to manually build them

.github/workflows/publish-release.yml
```diff
@@ -328,8 +269,8 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ github.token }}
 
-      - name: Publish GitHub Draft Release
+      - name: Publish GitHub Release
         env:
           GITHUB_TOKEN: ${{ github.token }}
         run: |
-          gh release edit ${{ env.new_tag }} --draft=true
+          gh release edit "${{ env.new_tag }}" --draft=false
```
imnotjames: Instead of a second step, perhaps the `publish` option for `release-drafter` could help?

```suggestion
      - name: Update GitHub Release Draft
        uses: release-drafter/release-drafter@6db134d15f3909ccc9eefd369f02bd1e9cffdf97 # v6
        with:
          tag: ${{ env.new_tag }}
          name: "Release ${{ env.new_tag }}"
          publish: true
        env:
          GITHUB_TOKEN: ${{ github.token }}
```

.github/workflows/publish-release.yml
```diff
@@ -160,165 +182,84 @@ jobs:
...
+            type=registry,ref=ghcr.io/grimmory-tools/grimmory:buildcache
           cache-to: |
             type=gha,mode=max
-            type=registry,ref=ghcr.io/booklore-app/booklore:buildcache,mode=max
+            type=registry,ref=ghcr.io/grimmory-tools/grimmory:buildcache,mode=max
+
+      - name: Create Release Tag
+        run: |
+          git tag -a "${{ env.new_tag }}" -m "Release ${{ env.new_tag }}"
+          git push origin "${{ env.new_tag }}"
```
imnotjames: Why a git tag here instead of letting the `release-drafter` create it?

Dockerfile
imajes: that's an idea. i'll consider that.

.github/workflows/publish-release.yml
imajes: we're expressing the git tag ourselves (i.e. choosing the tag we want) vs allowing release drafter to figure out the next tag.

imnotjames: Specifying `tag: ${{env.new_tag}}` is supposed to specify exactly what tag we want, I thought - that's not working?

Per [the release-drafter docs on `tag`](https://github.com/release-drafter/release-drafter?tab=readme-ov-file#action-inputs): 

> The tag name to be associated with the GitHub release that's created or updated. This will override any tag-template specified in your release-drafter.yml if defined.

Dockerfile
imnotjames: Updated to combine tar + make.

README.md
```diff
@@ -2,7 +2,7 @@
 > This project is in early development. Some links (website, documentation, demo, repositories) are not yet active and will come online as infrastructure is set up.
 
 > [!NOTE]
-> Grimmory is the successor of booklore.
+> Grimmory is the successor of the original project.
```
balazs-szucs: I personally can't really make sense why this is needed?

README.md
```diff
@@ -86,13 +86,27 @@ MYSQL_DATABASE=grimmory
...
-Create a `docker-compose.yml`:
+Stable images are published as `vX.Y.Z` plus `latest` from `main`. Nightly images are published as `nightly` from `develop`.
+
+> [!NOTE]
+> Upgrading from an existing Booklore container? Keep your current service name, `container_name`, database name/user, ports, and mounted volumes exactly as they are and replace only the `image:` line with `grimmory/grimmory:<tag>` or `ghcr.io/grimmory-tools/grimmory:<tag>`.
+
+```yaml
+services:
+  booklore:
+    image: grimmory/grimmory:v0.38.2
```
balazs-szucs: v0.38.2 seems to be totally random, I can see for this specifically is an example, but for can put the first "release" instead of random version? I've seen at quite few other places, may not the biggest deal, but annoys that part of my brain.

README.md
```diff
@@ -86,13 +86,27 @@ MYSQL_DATABASE=grimmory
...
 ```yaml
 services:
   grimmory:
-    image: grimmory/grimmory:latest
-    # Alternative: ghcr.io/grimmory-tools/grimmory:latest
+    image: grimmory/grimmory:v0.38.2
+    # Convenience tag:
+    # image: grimmory/grimmory:latest
+    # Alternative: ghcr.io/grimmory-tools/grimmory:v0.38.2
+    # The image already embeds its own release version metadata.
```
balazs-szucs: There is no reason for the README to have this detail. I don't think any ordinary user care about release version metadata

docs/forward-auth-with-proxy.md
```diff
@@ -36,8 +36,8 @@ REMOTE_AUTH_GROUPS_DELIMITER=\\s+          # Regex pattern for splitting groups.
 
 ```yaml
 services:
-  booklore:
-    image: ghcr.io/adityachandelgit/booklore-app:latest
+  grimmory:
+    image: ghcr.io/grimmory-tools/grimmory:v0.38.2
```
imnotjames, line 40: I'm on board with a specific versioned release, but this will lead to a number of folks showing up for support saying "it doesn't work" because the version doesn't exist / using an old version.

At the very least this should be 2.3.0?  If not `latest` still.

README.md
imajes: agreed. the readme needs a full overhaul but for now just trying to get it functional.

README.md
imajes: agreed, no idea why that showed up.

README.md
imajes: Yeah that was an errant booklore rename. fixed.

.github/workflows/publish-release.yml
imajes: Yeah- i looked at all of this and came to the conclusion the way I had structured it, plus using release-drafter was not exactly what i want. Instead i've moved to semantic-release.

Dockerfile
imajes: Thanks-- I went with this.

Dockerfile
imajes: I switched this to consume linuxserver/unrar as per another reviewer's suggestion.

PR #82, opened 2026-03-20
refactor(file): improve path handling and validation

balazs-szucs:
## 📝 Description

This pull request focuses on improving file path validation and security throughout the codebase, especially to prevent directory traversal attacks and ensure safe file operations. It also includes some bug fixes and minor improvements to the OPDS feed and related services.

These changes collectively harden the application against path traversal vulnerabilities.

**Linked Issue:** Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * OPDS feeds now return UTF‑8 encoded payloads and have improved XML escaping.
  * Migration ignores suspicious resource names to avoid unsafe writes.
  * Path validation blocks directory traversal and malformed paths.

* **Improvements**
  * Upload/download flows enforce files remain within expected filesystem roots.
  * Pagination URLs rebuilt from a whitelist and safely URL‑encoded.
  * File/folder hashing and path utilities now validate and normalize paths.


booklore-api/src/main/java/org/booklore/service/book/BookDownloadService.java
zachyale: we made changed to [`downloadBook()`](https://github.com/grimmory-tools/grimmory/pull/82/changes#diff-c442d87c5cfc4ab9f81942e178449a1572798cad67a397815ee2ba279e469ca7L62) but didn't make the same hardening apply for `downloadBookFile()`. 

Should we not add similar hardening, i.e.:


```suggestion
            Path libraryRoot = Path.of(bookFileEntity.getBook().getLibraryPath().getPath());
            Path file = FileUtils.requirePathWithinBase(bookFileEntity.getFullFilePath(), libraryRoot);
```

booklore-api/src/main/java/org/booklore/service/upload/FileUploadService.java
```diff
@@ -302,35 +313,80 @@ private void validateAlternativeFormatDuplicate(String fileHash) {
...
+        }
+    }
+
+    private Path toSafeRelativePath(String relativePath) {
+        if (!StringUtils.hasText(relativePath)) {
+            throw ApiError.GENERIC_BAD_REQUEST.createException("Invalid upload target path");
+        }
+
+        try {
+            Path parsed = Path.of(relativePath).normalize();
```
zachyale: Similar to the other issue, we've got an ordering issue where it's normalized first and then validated after normalization, no?

If I'm understanding this correctly, should we not be validating traversal on the raw parsed path before normalizing? Otherwise, traversal indicators wouldn't be included in the check

booklore-api/src/main/java/org/booklore/service/upload/FileUploadService.java
```diff
@@ -77,10 +78,10 @@ public void uploadFile(MultipartFile file, long libraryId, long pathId) {
             final String uploadPattern = fileMovingHelper.getFileNamingPattern(libraryEntity);
 
             final String relativePath = PathPatternResolver.resolvePattern(metadata, uploadPattern, originalFileName);
-            final Path finalPath = Paths.get(libraryPathEntity.getPath(), relativePath);
+            final Path finalPath = resolvePathWithinRoot(libraryRoot, relativePath);
```
zachyale, line 83: The main thing I'm noticing is the order of validation vs traversal- we traverse after normalizing, so relative `..` segments would be removed before validation, no? Should we not be validating the raw path first and then normalizing?

booklore-api/src/main/java/org/booklore/service/book/BookDownloadService.java
zachyale: (I don't often write suggested changes inline like this so forgive me if it's not formatted properly 🫶 )

booklore-api/src/main/java/org/booklore/service/upload/FileUploadService.java
zachyale, line 186: For consistency around handling relative paths, should we also add normalization to the library paths in this flow as well?

zachyale (commented):
Sorry for the mess in for not batching these comments into one review, but I think there may be something to some of the coderabbit comments. Let me know what you think about these specific cases, would like to hear your perspective on these surface level observations as you're closer to the metal on this

booklore-api/src/main/java/org/booklore/service/upload/FileUploadService.java
balazs-szucs: done

booklore-api/src/main/java/org/booklore/service/upload/FileUploadService.java
balazs-szucs: i think thats a good idea :+1:

booklore-api/src/main/java/org/booklore/service/upload/FileUploadService.java
balazs-szucs: done

booklore-api/src/main/java/org/booklore/service/book/BookDownloadService.java
balazs-szucs: looks great thank you for the suggestion :'D

zachyale (approved):

PR #128, opened 2026-03-22
fix(remote): fix lowercase headers

lukas-hetzenecker:
## Description

Load balancers like traefik add headers in the form of e.g. "Remote-User".

When configuring these headers in booklore/grimmory, they are evaluated as the lower-case variant "remote-user". These are not found in the header mapping. 

To fix this, forcing the headers to be lower-case has been removed. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed remote authentication header handling to read configured header names as provided (removed case-normalization), improving login reliability with custom headers and in environments where header key casing varies.
  * Preserved existing fallback behavior for email-based user lookup to maintain compatibility with prior login flows.


balazs-szucs (changes requested):
Use case insensitive map. 

This was changed I believe because HTTP/2 mandates lowercase. So this essentially, fixes HTTP/1 but breaks HTTP/2. 

I think I'm quite certain about 99% that there some Spring method/Spring way to this without doing whacky custom stuff. The solution for this is in Spring docs... somewhere.

balazs-szucs:
headers.get is already case-insensitive... apparently. There is more to this than what you are submitting.

PR #250, opened 2026-03-28
fix(backend): Fix metadata fields being stripped from API response, causing issues with filtering.

JVT038:
## Description

For a while, the backend would automatically strip all the metadata fields from the books API endpoint.
This caused issues with filtering, because metadata information was removed from the API response, the frontend couldn't properly filter the correct books, especially with magic shelves.

**Linked Issue:** Fixes an issue from the Booklore repo, but I can't be bothered to recreate the issue, sorry.

## Changes

I separated the `withDescription` boolean from the `stripListForView` boolean. Previously, the `stripListForView` was dependent on the value of `withDescription`, and because `withDescription` was always false, `stripListForView` would be true (thus stripping the metadata fields). 

Now there's a new (optional) parameter for the books API endpoint, namely `stripListForView` which by default is true. If it's true, then the metadata fields will be stripped from the API response (thus saving response size with a couple of bytes) and if it's false, then the metadata fields will be left intact (for frontend filtering).

## Release Notes

* **New Features**
  * Introduced `stripForListView` query parameter for the book list endpoint, enabled by default. This new parameter controls which metadata fields are included in API responses, allowing the server to automatically exclude unnecessary fields and optimize list view performance. Clients can disable it when complete book details are required.


balazs-szucs (approved):
Thank you, looks good.

PR #264, opened 2026-03-29
fix(bookdrop): resolve single-file processing hang and harden bookdrop delete safety

kemko:
## Description

Fix bookdrop processing bug where only one file is added to the library per application restart. When multiple books are placed in the bookdrop folder, grimmory discovers all of them but processes only one before the worker thread hangs indefinitely.

**Linked Issue:** Fixes #

### Root Cause
The `RestClient` in `HardcoverBookSearchService` and `HardcoverSyncService` was built **without any HTTP timeouts**. Spring's default `SimpleClientHttpRequestFactory` uses infinite timeouts, so when the Hardcover API fails to respond (rate limiting, server issues), the HTTP call blocks forever.

Since bookdrop file processing uses a **single worker thread** (`BookdropFileProcessor`), a hung Hardcover request blocks all subsequent file processing. On each restart, one more file gets saved to the DB before the thread hangs again on the next Hardcover API call.

### Fix 1: Add HTTP timeouts to Hardcover RestClient (primary fix)
- **`HardcoverBookSearchService`**: Add 10s connect / 15s read timeout via `SimpleClientHttpRequestFactory`
- **`HardcoverSyncService`**: Same timeout configuration

Timeout values are consistent with other services in the codebase (`OidcDiagnosticService`: 10s/10s, `DoubanBookParser`: 15s, `LubimyCzytacParser`: 10s).

### Fix 2: Harden worker thread exception handling (defense-in-depth)
- **`BookdropEventHandlerService.processQueue()`**: Wrap `processFile()` in `catch(Exception)` so any uncaught error logs and continues to the next file instead of killing the worker thread
- **`BookdropEventHandlerService.processFile()`**: Add `try-catch` to the `ENTRY_DELETE` branch (previously had no error handling, unlike the CREATE/MODIFY branch)
- Add `return` after `InterruptedException` for clean shutdown

## Summary by CodeRabbit

* **Bug Fixes**
  * Worker loop and event processing are more resilient—individual event failures are logged and processing continues; shutdown interrupts are handled cleanly. File-deletion now reports exact and child counts and logs failures without stopping processing.

* **New Features**
  * Centralized, configurable Hardcover API client with connection/read timeouts (10s/15s) for more reliable external calls; services now use the shared client.

* **Tests**
  * Unit tests updated to construct services with the injected Hardcover client.

### DIFF
--- a/booklore-api/src/main/java/org/booklore/repository/BookdropFileRepository.java
+++ b/booklore-api/src/main/java/org/booklore/repository/BookdropFileRepository.java
@@ -30,7 +30,7 @@ public interface BookdropFileRepository extends JpaRepository<BookdropFileEntity
 
     @Transactional
     @Modifying
-    @Query("DELETE FROM BookdropFileEntity f WHERE f.filePath LIKE CONCAT(:prefix, '%')")
+    @Query("DELETE FROM BookdropFileEntity f WHERE f.filePath LIKE CONCAT(:prefix, '%') ESCAPE '\\'")
     int deleteAllByFilePathStartingWith(@Param("prefix") String prefix);
 
     @Query("SELECT f.id FROM BookdropFileEntity f WHERE f.id NOT IN :excludedIds")
--- a/booklore-api/src/main/java/org/booklore/service/bookdrop/BookdropEventHandlerService.java
+++ b/booklore-api/src/main/java/org/booklore/service/bookdrop/BookdropEventHandlerService.java
@@ -171,7 +171,8 @@ public class BookdropEventHandlerService {
                 log.info("Detected deletion event: {}", deletedPath);
 
                 int exactCount = bookdropFileRepository.deleteByFilePath(deletedPath);
-                int childCount = bookdropFileRepository.deleteAllByFilePathStartingWith(deletedPath + File.separator);
+                String escapedPrefix = escapeLikeWildcards(deletedPath + File.separator);
+                int childCount = bookdropFileRepository.deleteAllByFilePathStartingWith(escapedPrefix);
                 int deletedCount = exactCount + childCount;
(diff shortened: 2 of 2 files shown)


kemko:
Should I implement AI suggested fixes? I didn't change this lines, I just adjust the indentation before it.

booklore-api/src/main/java/org/booklore/config/HardcoverClientConfig.java
```diff
@@ -0,0 +1,34 @@
...
+@Configuration
+public class HardcoverClientConfig {
+
+    static final String HARDCOVER_API_URL = "https://api.hardcover.app/v1/graphql";
+    static final Duration CONNECT_TIMEOUT = Duration.ofSeconds(10);
+    static final Duration READ_TIMEOUT = Duration.ofSeconds(15);
+
+    @Bean
+    public RestClient hardcoverRestClient() {
+        SimpleClientHttpRequestFactory factory = new SimpleClientHttpRequestFactory();
```
balazs-szucs, line 25: Seems rather unnecesary what preventing us from just setting some global variable for this e.g.,

```yaml
spring:
  http:
    client:
      connect-timeout: 30s
      read-timeout: 15s
```

or similar?

There are some much better solutions to this here: https://www.baeldung.com/spring-rest-timeout

But HardcoverClientConfig.java smells.

balazs-szucs: You can also just use a `@Bean` using the Spring API instead of SimpleClientHttpRequestFactory

You can use some hybrid setup of Resilliance4j and Spring

But this solution you have right now I'm really not passionate about. I'd forget this exists in 1 day, and it wouldn't fun to DEBUG afterwards.

PR #329, opened 2026-04-01
refactor(library): implement asynchronous book addition and cover generation with event handling

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes
This pull request introduces improvements to the book processing workflow, focusing on event-driven architecture, asynchronous processing, and enhanced cover image handling. The main changes include refactoring how book addition side effects are managed, introducing new services for cover generation and book group processing, and updating the asynchronous execution configuration.

**Event-driven book addition and asynchronous processing:**

* Replaced direct calls to `KoboAutoShelfService` and notification broadcasting in `BookDropService` with the publishing of a new `BookAddedEvent`. This decouples side effects from the main transaction and enables more modular handling of post-book-creation actions
* Added the `BookAddedEvent` record and the `BookAddedEventListener` component, which asynchronously handles book addition events after transaction commit. The listener broadcasts notifications and updates Kobo shelves, ensuring these actions are performed outside the main transaction for better reliability and scalability. 

**Book file and cover management enhancements:**

* Introduced the `BookCoverGenerator` service, which generates and attaches cover images to books from additional files, supporting both audiobooks and ebooks. This service ensures that cover images are extracted and saved appropriately when new formats are added to a book.
* Added the `BookGroupProcessor` service, which processes groups of book files in their own transactions. It selects the primary file, processes additional formats, attaches them to the book, and triggers the `BookAddedEvent`. This improves robustness and maintainability of the book import workflow.

**Asynchronous execution configuration updates:**

* Refactored async configuration by moving `@EnableAsync` from the main application class to a dedicated configuration class. The task executor is now a `DelegatingSecurityContextExecutor` wrapping a virtual thread executor, ensuring proper security context propagation for async tasks. 

## Summary by CodeRabbit

* **New Features**
  * Event-driven, post-commit async handlers for book-added and library-scan workflows
  * Library scan request event and listener with in-memory deduplication to avoid overlapping scans
  * Automated cover-generation component and a group-based file attachment processor

* **Improvements**
  * Background execution moved to virtual-thread–backed async executor with propagated security context
  * Watcher now skips processing while a library scan is active

* **Refactor**
  * Orchestrator simplified into a focused group processor/delegator

* **Tests**
  * Unit tests updated for event-driven flows, cover generation, and delegation changes


booklore-api/src/main/java/org/booklore/service/library/BookCoverGenerator.java
```diff
@@ -0,0 +1,159 @@
...
+                    log.warn("Failed to decode cover image for audiobook '{}'", audioFile.getFileName());
+                    return;
+                }
+                boolean saved = fileService.saveAudiobookCoverImages(originalImage, bookEntity.getId());
+                originalImage.flush();
+
+                if (saved) {
+                    bookEntity.getMetadata().setAudiobookCoverUpdatedOn(Instant.now());
+                    bookEntity.setAudiobookCoverHash(BookCoverUtils.generateCoverHash());
+                    bookRepository.save(bookEntity);
```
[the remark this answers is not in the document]
balazs-szucs: :+1:

zachyale (commented):
On a surface level nothing crazy, but let's hold off until we get the 3rd party contributor stuff in so we don't mess with their PRs

balazs-szucs:
> On a surface level nothing crazy, but let's hold off until we get the 3rd party contributor stuff in so we don't mess with their PRs

Right now, scanning in a library or any event that requires these book events to work well... Simply dont. They are blocking each other, and you have hard refresh for anything... What 3rd contributor stuff is related to this?

booklore-api/src/main/java/org/booklore/service/bookdrop/BookDropService.java
```diff
@@ -493,7 +494,7 @@ private BookdropFileResult processMovedFile(BookdropFileEntity bookdropFile,
                     .build();
 
             metadataRefreshService.updateBookMetadata(context);
-            koboAutoShelfService.autoAddBookToKoboShelves(bookEntity.getId());
+        eventPublisher.publishEvent(new BookAddedEvent(fileProcessResult.getBook()));
```
zachyale, line 497: We still send `notificationService.sendMessage(Topic.BOOK_ADD, fileProcessResult.getBook());` within `BookDropService`, while also adding the new event- is it our intent here to consolidate these behaviours or are we making note of this for the future (to refactor to only use one event)

booklore-api/src/main/java/org/booklore/service/watcher/LibraryFileEventProcessor.java
```diff
@@ -70,6 +72,11 @@ public void init() {
     }
 
     public void processEvent(WatchEvent.Kind<?> eventKind, long libraryId, Path fullPath, boolean isDirectory) {
+        if (libraryScanListener.isScanning(libraryId)) {
```
zachyale, line 109: We early return if a scan is pending- doesn't that mean if a library scan is pending (for a very large library, 10k-100k+, which could take a minute or so), and you performed a file action, like deleting a file entry or re-organizing the book into a new entry, that it would early exit and that event would be ignored?

So if your library is mid scan on a huge library that takes considerable processing time, you're effectively in a state where any underlying file changes don't take place but you don't know that, so it appears broken?

balazs-szucs: i do not personally see any problems with this approach. Programming a scanning that would be robust against operation that take place against files midflight, would be _much_ larger undertaking. I think the point you raise, is partially valid, but i would rather make logic to _inform_ users about such architecture decision, rather then actually forcing code through that'd allow this. E.g., "You have pending file deletion" or something along the lines.

booklore-api/src/main/java/org/booklore/service/bookdrop/BookDropService.java
balazs-szucs: this is something i sorta inherited. This was old behavior, that got translated into a new architecture. If you scan in your library you can kinda see this happening real-time in the logs, that there 2 separate logs for each book.

i do not personally mind this, also.

zachyale (approved):
There may be value in making follow up issues re: the two points but they aren't blockers

PR #369, opened 2026-04-04
fix(api): use ASIN when specified for retrieving Amazon / Audible metadata

### LINKED ISSUE #366: Amazon always searches even when ASIN is present
### What Went Wrong?

The ASIN being present in the database does not prevent an Amazon search, so the ASIN can be replaced.

1. Set Amazon as primary / only metadata
2. Set ASIN field
3. Fetch metadata
4. See ASIN changes

ASIN should be used for metadata

_No response_

Goodreads prevents a search if the ID is set.

### Your Setup

Grimmory 2.3.0

### Before Submitting

imnotjames:
## Description

Updates the `AmazonBookParser` and `AudibleParser` to use existing ASIN when fetching top metadata - to match behavior of `GoodReadsParser`

**Linked Issue:** Fixes #366 

## Changes

* updates `AmazonBookParser` to use `List` for return type of `getAmazonBookIds`
* uses the ASIN instead of searching by ISBN / title when available for `AmazonBookParser`
* uses the ASIN instead of searching by ISBN / title when available for `AusibleParser`
* adds minimal test for changes to `AmazonBookParser`
* adds minimal test for changes to `AudibleParser`

## Summary by CodeRabbit

* **Refactor**
  * External ID lookups now always return empty lists instead of null and pick a single top ID (preferring a validated stored ID) for metadata fetches.

* **Bug Fixes**
  * Improved ID extraction and validation to skip malformed, duplicate, or empty IDs, normalize inputs, and treat empty-result paths consistently to reduce failed lookups.

* **Tests**
  * Added tests verifying selection, sanitization, domain-aware lookups, and fallback search behavior for external-ID resolution.

### DIFF
--- a/booklore-api/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
+++ b/booklore-api/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
@@ -33,7 +33,7 @@ import java.util.stream.Collectors;
 @Service
 @AllArgsConstructor
 public class AmazonBookParser implements BookParser, DetailedMetadataProvider {
-
+    private static final Pattern ASIN_PATTERN = Pattern.compile("([A-Z0-9]{10})");
     private static final Pattern TRAILING_BR_TAGS_PATTERN = Pattern.compile("(\\s*<br\\s*/?>\\s*)+$");
     private static final Pattern LEADING_BR_TAGS_PATTERN = Pattern.compile("^(\\s*<br\\s*/?>\\s*)+");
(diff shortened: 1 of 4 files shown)


booklore-api/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
imnotjames: Ensure we never get null in our book IDs.

```suggestion
                    String bookId = extractAmazonBookId(item);
                    if (bookId == null) {
                        continue;
                    }
                    bookIds.add(bookId);
```

imnotjames:
Marked as draft for a bit.  Going to add tests and handle edge cases.

PR #376, opened 2026-04-04
refactor(frontend): custom font files caching

NamLe029:
## Description

Utilize persistent CacheStorage to cache font files on client-side, which may slightly improve load time.

## Changes

- Use CacheStorage to cache font files.
- Include `Last-Modified` header in server response.
- Pass authentication token by `Authorizatrion` header (used to be passed as a URL parameter).

#### Notes

- Must be merged after #361, as `Authorizatrion` header authentication for custom font is only supported until then

## Summary by CodeRabbit

* **New Features**
  * Optional custom-font caching: when enabled, fonts load from local cache for faster startup; when disabled, fonts are fetched on demand.

* **Improvements**
  * Font downloads now set clearer HTTP cache and content headers (including Last-Modified) to improve client handling.

* **Frontend**
  * Font loading uses blob-derived object URLs for consistent rendering across cache/no-cache modes.

* **Tests**
  * Updated tests to cover the new cache-dependent font-loading behavior.


zachyale (changes requested):
@NamLe029 The functional changes in the PR look good, but specs are failing and need to be updated for closer compliance with your new implementation of `EpubCustomFontService` and the `HttpClient` + header/cache approach.

Let me know when specs have been addressed and we can get this in 👍

balazs-szucs:
Hi,

this has been stale for while, please re-raise later if still make sense.

PR #377, opened 2026-04-04
feat: add reading progress endpoints to mobile api

### LINKED ISSUE #364: Add reading progress endpoints to the Mobile App API
### What's Your Idea?

Add dedicated reading progress endpoints to the Mobile App API (/api/v1/app/):

  - GET /api/v1/app/books/{bookId}/progress — retrieve current reading progress for the authenticated user
  - PUT /api/v1/app/books/{bookId}/progress — update reading progress

  This would give mobile clients a clean, JWT-authenticated way to both push and pull reading progress without relying on kosync or fetching full book details.

### Why Would This Be Helpful?

The app API currently has no way to directly read or write reading progress. Mobile apps that poteintailly integrate with Grimmory via JWT auth are stuck with workarounds:

  - Fetching the entire book detail (GET /api/v1/app/books/{bookId}) just to check if progress changed — wasteful when the client only needs to sync progress
  - Using kosync — requires separate KoReader credentials, document hashing, and is a different integration path entirely
  - The main REST API has POST /api/v1/books/progress for pushing but no GET equivalent for pulling

  This gap makes it difficult to build efficient progress sync for potential third-party mobile readers. The app API already has per-book endpoints for status (PUT .../status) and rating (PUT  .../rating) — progress is the missing piece that would complete the pattern.

### Anything Else? (Optional)

The response for the GET endpoint could reuse the existing progress DTOs already returned in AppBookDetail:

  {
    "readProgress": 0.42,
    "epubProgress": { "cfi": "epubcfi(...)", "href": "chapter1.xhtml", "percentage": 0.42 },
    "pdfProgress": { "page": 50, "percentage": 0.17 },
    "cbxProgress": null,
    "audiobookProgress": null,
    "koreaderProgress": { "percentage": 0.42, "document": "hash", "timestamp": 1711440600 }
  }

  The PUT endpoint could accept the same ReadProgressRequest shape already used by POST /api/v1/books/progress, keeping the implementation consistent.

### Want to Help Out?

### Have You Considered Any Alternatives? (Optional)

The current workarounds are:
  1. Parsing progress out of the full book detail response — functional but inefficient for sync-only use cases
  2. Using kosync — works but requires separate credential management and is designed for KOReader compatibility, not general-purpose mobile API usage

I currently use the work around for a mobile app I made for my personal use and it works but it feels janky. 

### Before Submitting

- [x] I've searched existing issues and confirmed this feature hasn't been requested yet

afairgiant:
## Description

This pull request adds app API support for retrieving and updating a user's reading progress for a specific book in Grimmory. It introduces new endpoints, a response DTO, and supporting service and mapping logic to handle book progress data in a structured way.

**Linked Issue:** Fixes #364

## Changes

**API Enhancements:**

* Added two new endpoints to `AppBookController`: one for fetching (`GET /{bookId}/progress`) and one for updating (`PUT /{bookId}/progress`) the reading progress of a book, using the new `AppBookProgressResponse` DTO and `ReadProgressRequest` respectively.

**Data Transfer Object (DTO) Additions:**

* Introduced `AppBookProgressResponse` to encapsulate reading progress details, including progress metrics for various book formats and last read time.

**Service Layer Updates:**

* Added `getBookProgress` and `updateBookProgress` methods to `AppBookService` to handle fetching and updating user reading progress, with appropriate access validation and delegation to `BookService` for updates.

**Mapping Improvements:**

* Implemented a new `toProgressResponse` method in `AppBookMapper` to map progress entities to the new response DTO.

**Dependency Injection and Testing:**

* Registered `BookService` as a dependency in `AppBookService` and updated related tests to mock and inject this dependency. [[1]](diffhunk://#diff-c0813625cfeee4e1a6ce316cb4c55c42231baf5ad04dca346fc54866e370a325R55) [[2]](diffhunk://#diff-e8827c38b58873f6690eba0a9f38a945c8df633fa719421778d987fb1811d704R51) [[3]](diffhunk://#diff-e8827c38b58873f6690eba0a9f38a945c8df633fa719421778d987fb1811d704L62-R64)

## Summary by CodeRabbit

* **New Features**
  * View detailed reading progress and status for any book, including per-format progress (epub, pdf, cbx, audiobook, koreader) and last-read time via a new progress endpoint.
  * Update reading progress via a single progress endpoint (accepts format-specific progress or a completion timestamp).

* **Tests**
  * Added coverage for progress read/update flows, access control, and validation.


booklore-api/src/main/java/org/booklore/app/controller/AppBookController.java
```diff
@@ -46,6 +47,22 @@ public ResponseEntity<AppBookDetail> getBookDetail(
...
     }
 
+    @GetMapping("/{bookId}/progress")
+    public ResponseEntity<AppBookProgressResponse> getBookProgress(
+            @PathVariable Long bookId) {
+
+        return ResponseEntity.ok(mobileBookService.getBookProgress(bookId));
+    }
+
+    @PutMapping("/{bookId}/progress")
```
zachyale, line 53: This enpoint is path scoped but still accepts `ReadProgressRequest` (which requires `bookId` in the body). That gives the app API two sources of truth for the same identifier- we should probably just switch this to an app specific request DTO w/o `bookId`, then map the path ID into `ReadProgressRequest`. This would better align with the existing patterns for `{bookId}/status` and `{bookId}/rating`

booklore-api/src/main/java/org/booklore/app/service/AppBookService.java
```diff
@@ -109,6 +113,45 @@ public AppBookDetail getBookDetail(Long bookId) {
...
+    public void updateBookProgress(Long bookId, ReadProgressRequest request) {
+        BookLoreUser user = authenticationService.getAuthenticatedUser();
+        Set<Long> accessibleLibraryIds = getAccessibleLibraryIds(user);
+
+        BookEntity book = bookRepository.findByIdWithBookFiles(bookId)
+                .orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(bookId));
+
+        validateLibraryAccess(accessibleLibraryIds, book.getLibrary().getId());
+
+        request.setBookId(bookId);
```
zachyale: By overriding the `request.BookId`, it overrules client mistakes and makes the request contract a bit ambiguous- if we decide to instead keep reusing `ReadProgressRequest`, we should at least reject body/path mismatches w/ a 400 instead of overwriting them. Regardless, still preferable to just remove `bookId` from the request payload entirely

booklore-api/src/main/java/org/booklore/app/service/AppBookService.java
```diff
@@ -109,6 +113,45 @@ public AppBookDetail getBookDetail(Long bookId) {
...
+
+        UserBookFileProgressEntity fileProgress = userBookFileProgressRepository
+                .findMostRecentAudiobookProgressByUserIdAndBookId(userId, bookId)
+                .orElse(null);
+
+        return mobileBookMapper.toProgressResponse(progress, fileProgress);
+    }
+
+    @Transactional
+    public void updateBookProgress(Long bookId, ReadProgressRequest request) {
```
zachyale: We should add focused test coverage for this new flow- at minimum:
- a test for the request contract here to ensure `bookId` is the authority, and mismatches are rejected or impossible
- a test to ensure forbidden book/library access still fails properly here

zachyale (changes requested):
Sorry for the delay in a review here- we've been working around the clock to get some major refactors in- luckily no conflicts on this area 😄 

Some minor comments w/ some suggested changes- open to your input here, but at minimum if we're going to introduce this support we should ensure we're using one source of truth for `bookId`

zachyale (approved):

zachyale:
@afairgiant  Thanks for the contribution- sorry for the delay on re-review 🚀

PR #428, opened 2026-04-08
chore(api): drop unused book download proxy

### LINKED ISSUE #349: Kobo Download URL forward to store does not operate as expected
### What Went Wrong?

The book download URL path proxies to kobo store but doesn't do anything.

Given that the URLs in the sync won't be for this path anyway, we can drop the `proxyCurrentRequest()` call entirely.

1. Set up the kobo sync token
2. Make a request to `http://localhost:6060/api/kobo/{token}/books/foo/download`
3. See nothing for a response

404 because this URL is only for downloading grimmory books

_No response_

_No response_

### Your Setup

Grimmory 2.3.0

### Before Submitting

imnotjames:
## Description

the book download proxy code path should never be called.  this is because the book download URL received as part of the kobo proxy points at the Rakuten Kobo CDN rather than at grimmory.  further, even if this code could download the file from the CDN (it doesn't), it would be pointless as we don't return it.  instead of fixing the return this drops the proxy from that code path

**Linked Issue:** Fixes #349

## Changes

* drops unused book download kobo store proxy call

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling for invalid book IDs in the download functionality, now properly returning "Not Found" responses.

### DIFF
--- a/booklore-api/src/main/java/org/booklore/controller/KoboController.java
+++ b/booklore-api/src/main/java/org/booklore/controller/KoboController.java
@@ -176,11 +176,9 @@ public class KoboController {
     public void downloadBook(@Parameter(description = "Book ID") @PathVariable String bookId, HttpServletResponse response) {
         if (StringUtils.isNumeric(bookId)) {
             bookDownloadService.downloadKoboBook(Long.parseLong(bookId), response);
-        } else if (isForwardingToKoboStore()) {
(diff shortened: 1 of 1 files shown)


balazs-szucs (approved):

balazs-szucs:
Thank you :)

PR #435, opened 2026-04-08
fix(reader): implement signal-based state management and cleanup in EbookReaderComponent

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

This pull request refactors the ebook reader component and its event service to improve state management and resource cleanup. The main changes include migrating UI state variables to Angular signals for better reactivity, improving cleanup logic for event listeners (especially for iframes), and updating the template and component code to use the new reactive patterns.

## Summary by CodeRabbit

* **Performance Improvements**
  * Optimized change detection strategy for faster rendering and reduced resource consumption.
  * Improved event handling efficiency with better listener management.

* **Bug Fixes**
  * Enhanced event cleanup mechanisms to prevent memory leaks and ensure stable operation during extended reading sessions.


zachyale (commented):
I think this PR still has a real memory retention issue, despite the refactor.

The current approach seems to store per-iframe cleanup in `iframeCleanupFns`, and those closures capture the iframe `Document`/`EventTarget`- foliate then swaps iframe documents as each section changes, but those cleanup fn's are only killed in `destroy()`, so old secn. docs stay reachable for the entire reader session instead of becoming collectible when foliate removes them.

Is this a blocker? not really, but it doesn't fully address the issues in this space. The reader still appears to work, but the detached iframes can accumulate across long sessions. Thoughts?

zachyale (changes requested):
as per my comment, i can be convinced to approve this as is but want to hear your response

balazs-szucs:
for me the memory graphs look significantly better. I was not aiming for anything large. What you are saying probably makes sense, but as far as quick wins go this is the best i could do.



Screenshot 2026-04-10 at 11 23 00
Screenshot 2026-04-10 at 11 23 45

zachyale (approved):

zachyale:
@balazs-szucs as per our convo let's make a follow up issue to revisit this for further modernization

PR #473, opened 2026-04-11
fix(kobo): handle missing book when pulling kobo metadata

### LINKED ISSUE #274: KoboEntitlementService.getMetadataForBook()` crashes with 500 when book does not exist
### What Went Wrong?

[This bug report was written with the assistance of an LLM, as I'm not tech savvy enough to understand]

[This may be related to https://github.com/grimmory-tools/grimmory/issues/93 but I wasn't sure.]

When a Kobo device requests metadata for a book that either does not exist in Grimmory or exists without an associated file record, `KoboEntitlementService.getMetadataForBook()` throws an unhandled `java.util.NoSuchElementException` instead of returning a graceful 404. This causes the Kobo to receive a 500 and retry the sync in an infinite loop until the condition is resolved.
**Root cause**

`getMetadataForBook()` calls `List.getFirst()` on the result of a database query for the book's file records. When the list is empty — because the book was deleted, or the book record exists but has no associated file — `List.getFirst()` throws `NoSuchElementException`. This is not caught, and `GlobalExceptionHandler` maps it to a 500.

The crashing line:

```
at org.booklore.service.kobo.KoboEntitlementService.getMetadataForBook(KoboEntitlementService.java:280)
```

[Unsure. May work if you add a book, sync, delete it from the server then try to download if from the client without syncing first.]

Grimmory should return HTTP 404 when metadata is requested for a nonexistent book ID, allowing the Kobo to handle the missing book gracefully and move on.

Grimmory returns HTTP 500 on every metadata request for the missing book. The Kobo interprets this as a transient error and retries continuously, producing the following log loop approximately every 13 seconds:

```
INFO  o.b.service.kobo.KoboLibrarySyncService  : Synced 11 reading states to Kobo
INFO  o.b.service.kobo.KoboEntitlementService  : Synced 1 tags to Kobo
ERROR o.b.exception.GlobalExceptionHandler     : Unhandled exception: null

java.util.NoSuchElementException
    at java.base/java.util.List.getFirst(Unknown Source)
    at org.booklore.service.kobo.KoboEntitlementService.getMetadataForBook(KoboEntitlementService.java:280)
    at org.booklore.controller.KoboController.getBookMetadata(KoboController.java:138)
```

Caddy access log confirming the 500:

```
"method":"GET"
"uri":"/api/kobo/<token>/v1/library/24/metadata"
"status":500
```

Adding a new book to BookLore that receives the missing book ID satisfies the Kobo's metadata request and breaks the loop. This is not a reliable or scalable workaround.

### Your Setup

Grimmory 2.3.0
Docker on Debian
Kobo Clara BW

### Before Submitting

imnotjames:
## Description

instead of an uncaught error, missing books should be treated as a 404 not-found in the kobo metadata endpoint

**Linked Issue:** Fixes #274

## Changes

- return null instead of cause a `NullPointerException` when kobo book metadata won't exist
- return 404 when kobo book metadata is null

## Release Notes

* **Bug Fixes**
  * Improved error handling for the library metadata endpoint when requested books are not found, ensuring consistent HTTP 404 responses.
  * Enhanced robustness of metadata retrieval service to properly handle cases where book data is unavailable, preventing unexpected exceptions.
  * Added test coverage for null metadata scenarios.

### DIFF
--- a/booklore-api/src/main/java/org/booklore/controller/KoboController.java
+++ b/booklore-api/src/main/java/org/booklore/controller/KoboController.java
@@ -124,12 +124,16 @@ public class KoboController {
     @GetMapping("/v1/library/{bookId}/metadata")
     public ResponseEntity<?> getBookMetadata(@Parameter(description = "Book ID") @PathVariable String bookId) {
         if (StringUtils.isNumeric(bookId)) {
-            return ResponseEntity.ok(List.of(koboEntitlementService.getMetadataForBook(Long.parseLong(bookId), token)));
+            KoboBookMetadata metadata = koboEntitlementService.getMetadataForBook(Long.parseLong(bookId), token);
+
+            if (metadata != null) {
+                return ResponseEntity.ok(List.of(metadata));
+            }
         } else if(isForwardingToKoboStore()) {
             return koboServerProxy.proxyCurrentRequest(null, false);
-        } else {
-            throw ApiError.GENERIC_NOT_FOUND.createException("Not Found");
         }
+
+        throw ApiError.GENERIC_NOT_FOUND.createException("Not Found");
     }
 
     @Operation(summary = "Get reading state", description = "Retrieve the reading state for a book.")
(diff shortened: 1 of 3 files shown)


zachyale (approved):

balazs-szucs (approved):
Many thanks for Kobo PRs. It has been a pain point for many. I'm hoping, we can get into a very good shape in the next few releases (esp. in regards to stability). Generally, these a go a _very_ long way. Even though i cannot test these right now, but I think Zach did have a go at it.

PR #538, opened 2026-04-16
fix(auth): fix authentication request handling to exclude specific endpoints from authorization checks leading to permission denied errors

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

Resolve unwanted permission denied errors in for instance the User creator UI.

## Summary by CodeRabbit

* **Bug Fixes**
  * Tightened which authentication endpoints are treated as public: only login, token refresh, remote auth, and OIDC flows are excluded from automatic token injection and 401-refresh handling; other auth-related paths now require credentials and will trigger normal auth behavior.
* **Tests**
  * Added tests to verify precise exclusion of the listed auth endpoints and to reject similar-but-different paths (exact-match and query-string handling).


booklore-api/src/main/java/org/booklore/config/security/SecurityConfig.java
```diff
@@ -57,7 +57,9 @@ public class SecurityConfig {
             "/kobo/**",                // Kobo API requests (auth handled in KoboAuthFilter)
             "/api/docs",               // API documentation UI
             "/api/openapi.json",       // OpenAPI spec (used by API documentation UI)
-            "/api/v1/auth/**",         // Login and token refresh endpoints (must remain public)
+            "/api/v1/auth/login",      // Login endpoint (must remain public)
+            "/api/v1/auth/refresh",    // Token refresh endpoint (must remain public)
+            "/api/v1/auth/remote",     // Remote auth endpoint (must remain public)
```
utafrali, line 62: Replacing `/api/v1/auth/**` with three explicit entries silently removes all OIDC endpoints (`/api/v1/auth/oidc/**`) from the public whitelist. `OidcAuthController` is mapped to `/api/v1/auth/oidc` and exposes unauthenticated flows like `GET /api/v1/auth/oidc/state`, `POST /api/v1/auth/oidc/callback`, `GET /api/v1/auth/oidc/redirect`, `POST /api/v1/auth/oidc/mobile/callback`, and `POST /api/v1/auth/oidc/backchannel-logout`. None of these carry a JWT at call time. Without a whitelist entry they fall into `jwtApiSecurityChain` (Order 9), which enforces `.anyRequest().authenticated()`, producing a 401 for every OIDC login attempt. Add `/api/v1/auth/oidc/**` to `COMMON_PUBLIC_ENDPOINTS` alongside the new explicit entries.

frontend/src/app/core/security/auth-interceptor.service.ts
```diff
@@ -13,14 +13,16 @@ export const AuthInterceptorService: HttpInterceptorFn = (req, next: HttpHandler
 
   const token = authService.getInternalAccessToken();
   const isApiRequest = req.url.startsWith(`${API_CONFIG.BASE_URL}/api/`);
-  const isAuthRequest = req.url.startsWith(`${API_CONFIG.BASE_URL}/api/v1/auth/`);
+  const isExcludedAuthRequest = req.url.includes('/api/v1/auth/login') ||
```
utafrali: Using `.includes('/api/v1/auth/login')` performs a substring search on the full URL string, not a path-segment comparison. A future endpoint like `/api/v1/auth/login-history` or a URL whose query string happens to contain the fragment would also match, preventing the bearer token from being attached. Switch to a path-exact test, for example:
```typescript
const path = new URL(req.url, 'http://x').pathname;
const isExcludedAuthRequest =
  path === '/api/v1/auth/login' ||
  path === '/api/v1/auth/refresh' ||
  path === '/api/v1/auth/remote';
```
This also needs a corresponding entry for the OIDC endpoints (see the backend comment).

utafrali (commented):
The fix correctly narrows the `/api/v1/auth/**` wildcard so that `/api/v1/auth/register` is no longer excluded from JWT authentication, which resolves the permission-denied issue for the user creation UI. However, the narrowing also drops all OIDC endpoints (`/api/v1/auth/oidc/**`) from the public whitelist, which is an unintended regression that will break OIDC login flows entirely.

booklore-api/src/main/java/org/booklore/config/security/SecurityConfig.java
balazs-szucs: hi,

thanks, it was... a rather hotfix kinda of deal. Funnily, the rabbit thing also flagged it: https://github.com/grimmory-tools/grimmory/pull/538#discussion_r3093048360. Should be easy enough to resolve. I believe, Zach is working on a bigger rework of this, or at least i vaguely remember talking about, so this situation most likely is not permanent. Anyways, as mentioned thanks for flagging nonetheless.

frontend/src/app/core/security/auth-interceptor.service.ts
balazs-szucs: Looks good, i will take a look. Thanks.

booklore-api/src/main/java/org/booklore/config/security/SecurityConfig.java
balazs-szucs: damn i got bamboozled by this clawdbot at least 2 already. I should be really less gullible. Anyways i will take look regardless...

balazs-szucs:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=grimmory-tools_grimmory&pullRequest=538) **Quality Gate failed**
> 
> Failed conditions  [1 Security Hotspot](https://sonarcloud.io/project/security_hotspots?id=grimmory-tools_grimmory&pullRequest=538&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=grimmory-tools_grimmory&pullRequest=538)

its a false flag.

frontend/src/app/core/security/auth-interceptor.service.spec.ts
```diff
@@ -176,4 +176,50 @@ describe('AuthInterceptorService', () => {
...
+      authService.getInternalAccessToken.mockReturnValue('token-123');
+      const response = await firstValueFrom(interceptor(new HttpRequest('GET', `${API_CONFIG.BASE_URL}/api/v1/auth/login-history`), next));
+      expect((response as HttpResponse<boolean>).body).toBe(true);
+    });
+
+    it('does NOT exclude /api/v1/auth/login?query=1 (exact match test)', async () => {
+      authService.getInternalAccessToken.mockReturnValue('token-123');
+      const response = await firstValueFrom(interceptor(new HttpRequest('GET', `${API_CONFIG.BASE_URL}/api/v1/auth/login?query=1`), next));
+      expect((response as HttpResponse<boolean>).body).toBe(false); // login with query is still login
+    });
```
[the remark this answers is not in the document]
balazs-szucs: not important

booklore-api/src/main/java/org/booklore/config/security/SecurityConfig.java
```diff
@@ -57,7 +57,10 @@ public class SecurityConfig {
             "/kobo/**",                // Kobo API requests (auth handled in KoboAuthFilter)
             "/api/docs",               // API documentation UI
             "/api/openapi.json",       // OpenAPI spec (used by API documentation UI)
-            "/api/v1/auth/**",         // Login and token refresh endpoints (must remain public)
+            "/api/v1/auth/login",      // Login endpoint (must remain public)
```
imnotjames, line 60: Would this break open user registration?  Is that even a thing we support?  `/v1/auth/register` is the endpoint I'm thinking of.

imnotjames: Duh.  There's no open registration.  We could move the register API endpoint out of the `/api/v1/auth` and instead put it somewhere else.. 🤔 I'm sure this is also fine though.

imnotjames (approved):
This technically seems to work and doesn't open up any extra endpoints we don't expect.

booklore-api/src/main/java/org/booklore/config/security/SecurityConfig.java
balazs-szucs: > Duh. There's no open registration. We could move the register API endpoint out of the `/api/v1/auth` and instead put it somewhere else.. 🤔 I'm sure this is also fine though.

i think so too, but for now this works 🤷‍♂️ 

I believe, Zach volunteered himself to take a look at the security filters later down the line, but that was not yesterday, so not sure if he still up for the challenge 😆

zachyale:
reimplemented in https://github.com/grimmory-tools/grimmory/pull/566 to avoid conflicts and get it in

PR #540, opened 2026-04-16
fix: properly handle single-file audiobooks

### LINKED ISSUE #446: Uploaded audiobooks disappearing from app
### What Went Wrong?

I noticed recently that a bunch of audiobooks I uploaded in the past (prior to moving over to the Grimmory image from Booklore) are no longer present in app.  I'm not sure if they had been present immediately after change the images.  When I try to re-upload them, it says the files already exist.  I did some digging into the database and can not find database entries associated with these missing audiobook files.  However when I checked the volumes specified in my docker file, the one here `./apps/booklore/book/:/book` , showed the missing audiobook files.

I tried deleting these files and then re-uploading them through Bookdrop which did work.  I could see them in the app.  But then I checked back the next day and they are gone again from the app.  I no longer see a database record but still see the files in the book folder.

Not sure why this isn't happening to all of the audiobooks I've uploaded.  In this case, I'm testing with the Dungeon Crawler Carl audiobook series.  For some reason book 2 is saved and in the database but I noticed that the book file name got stored as the author's name instead of the file's actual name.

Image

I'm not sure if that's the reason behind the issue.  I know previously I had also played around with trying to group the books together because I also own the physical books, but that isn't possible.  I doubt that caused the problem.

All the database entries I can find for the series (also the only entries I can find in the app):

Image

Missing audiobook files in the `/mnt/lan-access/apps/booklore/book/Matt Dinniman` location :

Image

I'm not sure exactly how to reproduce as I don't know what caused it in the first place, but to recreate this specific situation:

1. Look for an audiobook that you have previously uploaded but is missing from the app.
2. Try to upload the file again --> Grimmory throws an error saying file already exists.
3. Check the specific `./apps/booklore/book/:/book` location for the book and see if the file exists.  If it does, then delete it.
4. Try reuploading again which should work.  Book should be viewable from the app.
5. Wait for a system sync and see if the book disappears again.  I'm not sure if the system sync is what caused it to disappear again after I re-uploaded, but when I checked audit logs that's all that happened since then.

I have not noticed this with epubs that I've uploaded.  All audiobooks that I'm having issues with were uploaded through Bookdrop.

I would have expected the first time I uploaded the audiobooks for them to get properly organized within the app's file system.  I also expected that when I re-uploaded them again, for them to not disappear again.

_No response_

I suspect that something happened when I originally uploaded these books.  I remember being able to see all of them when I first uploaded which is why I didn't suspect anything.  It's possible when moving over to the Grimmory image that something messed it up.

It is also possible that the system sync is tangentially related because something is resetting the files after I re-upload them.

### Your Setup

- Grimmory version 2.3.0
- Running on a self hosted NAS
- Docker

### Before Submitting

imnotjames:
## Description

in some cases, single file audio books are being removed when the library scan is being done.  this is because of ambiguity in our audiobook detection.  often, multiple single file audiobooks gets detected as a folder-based audiobook which leads to us not finding those single files

this change updates the algorithm for detecting files so we check for both folder & file based audiobooks

**Linked Issue:** Fixes #446

## Changes

* switches away from the "key" system to paths
* ensures audiobooks get both folder & file based paths when scanning the library for new / deleted files
* handles both of these cases during a scan to allow for improperly detected books to be found

## Summary by CodeRabbit

* **Refactor**
  * Improved internal file path comparison logic for more reliable library file detection.

* **Tests**
  * Expanded test coverage for library file management, including folder-based audiobook handling and file detection scenarios.

### DIFF
--- a/backend/src/main/java/org/booklore/service/library/LibraryFileHelper.java
+++ b/backend/src/main/java/org/booklore/service/library/LibraryFileHelper.java
@@ -23,9 +23,9 @@ import java.nio.file.attribute.BasicFileAttributes;
 import java.util.ArrayList;
 import java.util.EnumSet;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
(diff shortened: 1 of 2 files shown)


imnotjames:
Clarifying the problem here as a commit for future review:

For example, say I have Author "Foo" who made the audiobook "Bar" and "Baz"

| Author | Book | File |
| --- | --- | --- |
| Foo | Bar | `/books/foo/bar.mp3` |
| Foo | Baz | `/books/foo/baz.mp3` |

The problem is that when checking for book files during a scan, Grimmory erroneously detects `/books/foo` as a folder-based audiobook and thus only sees `/books/foo`.  When it checks to see if `/books/foo/bar.mp3` exists, it can't find those files (As it only reads the folder!) & decides the book has been removed from the library - even though the file is there!

---

The issue why we can't just add all files is then the individual files in an actual folder-based audiobook will get detected as new books, so we have to handle that case.

---

I also think there are other edge cases that this entire system doesn't handle - it was never written to support audiobook folders in the first place, so cases like an audiobook folder that has a single cover jpeg may break things.  A tree or moving the folder detection logic up into the processor makes more sense - but would be a larger rewrite of the feature.

---

Actually, moving in the other direction could work, too?  The delete / add detection helpers moved into the library file helper & making them public would allow testing them more directly.

balazs-szucs:
> I also think there are other edge cases that this entire system doesn't handle - it was never written to support audiobook folders in the first place, so cases like an audiobook folder that has a single cover jpeg may break things. A tree or moving the folder detection logic up into the processor makes more sense - but would be a larger rewrite of the feature.

agreed on both counts. Some of the weirdness sadly, is rather obvious. For instance, when you select audiobooks in book-view, and generate custom cover nothing happens, you also cannot also regenerate from the file. Whether auto-metadata match works or not, i think so but 🤷‍♂️ 

Interestingly, when you do it in edit metadata UI... then it works?

Anyways, this is to say, PRs there are very much welcome. For me it has not been priority because it's moreso an "opt-in" features with lot of great alternatives. E.g., Audiobookshelf.

Thank you for your work :)

imnotjames:
This is a bit much of a PR to review, probably - It's a refactor and then 1631578f7c73408470f9c0ac3b902d5f1fe33b4b is the actual fix.  If need be, I can split this into two PRs -- one to refactor with no behavior changes, and one to fix.

EDIT: Per DM, splitting would help.

imnotjames:
@coderabbitai review

PR #761, opened 2026-04-22
ci: temp disable dependabot typescript updates

imnotjames:
## Description

Configure dependabot to ignore typescript because it's incompatible to upgrade with our angular version

## Summary by CodeRabbit

* **Chores**
  * Updated dependency management configuration.


balazs-szucs (approved):
Many thanks for the quick PR!

PR #791, opened 2026-04-23
fix(repositories): add automatic flushing to delete methods in various repositories

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

This pull request refactors a number of repository classes to improve the handling of modifying queries in Spring Data JPA. The main focus is on ensuring that changes to the database are immediately synchronized with the persistence context and database by enabling both `clearAutomatically` and `flushAutomatically` in the `@Modifying` annotation. Additionally, several repository methods that previously relied on Spring Data's derived query methods for deletes have been converted to explicit JPQL `@Query` methods, often with added transactional annotations and parameter bindings for clarity and consistency.

Repository method improvements:

* Added `clearAutomatically = true, flushAutomatically = true` to all `@Modifying` annotations on update and delete methods across repositories, ensuring the persistence context is kept in sync after modifying queries. 
* Converted several delete methods from derived query methods to explicit JPQL queries with `@Query`, `@Modifying`, and `@Transactional` for greater control and clarity. This includes methods such as `deleteByBookIdAndUserId`, `deleteByUser`, `deleteByRevokedTrueAndCreatedAtBefore`, and others. 
* Added missing imports for `@Modifying`, `@Transactional`, `@Query`, and `@Param` where needed to support the new method implementations. 

Consistency and code clarity:

* Ensured that all repositories use a consistent pattern for modifying and deleting queries, improving maintainability and reducing the risk of stale data in the persistence context. (all references above)
* Removed unused or redundant query methods, such as a duplicate count query in `MetadataFetchJobRepository`.

## Summary by CodeRabbit

* **Refactor**
  * Improved backend bulk delete/update behavior and transaction handling for more reliable and consistent data operations.
  * Tuned automated cleanup processes for better stability during maintenance tasks.

* **Bug Fixes**
  * Fixed a task’s reported pending cleanup count source for more accurate reporting.
  * Made user restriction updates more deterministic by adjusting the update ordering.


backend/src/main/java/org/booklore/repository/AnnotationRepository.java
```diff
@@ -30,7 +32,10 @@ boolean existsByCfiAndBookIdAndUserIdExcludeId(
 
     long countByBookIdAndUserId(Long bookId, Long userId);
 
-    void deleteByBookIdAndUserId(Long bookId, Long userId);
+    @Modifying(clearAutomatically = true, flushAutomatically = true)
+    @Transactional
+    @Query("DELETE FROM AnnotationEntity a WHERE a.bookId = :bookId AND a.userId = :userId")
```
imnotjames, line 37: Is the reason why we are explicitly adding the query so we can add the transactional annotation?

backend/src/main/java/org/booklore/task/tasks/TempFetchedMetadataCleanupTask.java
```diff
@@ -71,7 +71,7 @@ public TaskType getTaskType() {
 
     @Override
     public String getMetadata() {
-        long count = metadataFetchJobRepository.countAll();
+        long count = metadataFetchJobRepository.count();
```
imnotjames: Why this change in this PR?  This seems unrelated to a singular goal for the change.

imnotjames:
Is there a related issue?  I get we're fixing something but I don't know what we're fixing.

backend/src/main/java/org/booklore/repository/AnnotationRepository.java
balazs-szucs: it is unused so i am 99% sure it can removed. yes and also @Query issues a single SQL DELETE instead of SELECT N only to then call another query to DELETE on it.

backend/src/main/java/org/booklore/task/tasks/TempFetchedMetadataCleanupTask.java
balazs-szucs: count is inherited from the spring JPA, countAll is ACX's whatever the hell he was doing.

imnotjames: I understands that, but it seems like an unrelated refactor.  Shouldn't it be a separate change?

balazs-szucs:
> Is there a related issue? I get we're fixing something but I don't know what we're fixing.

there is discussion to this, but the fix itself is at this 1-1.5 months old, i was just too lazy to actually maintain it so i always just made another branch instead of keep rebasing the old. See: #717 #523 and there was one before those, but i cannot find that...

I believe it will fix this issues: https://discord.com/channels/1483208318720675860/1483220176500428911/1496190276652240917

however, i have not able to reproduce therefore i haven't been to confirm that.

balazs-szucs:
this is however idiomatic hibernate usage, it's a best practice in general... if you think about it makes lot of sense why these are needed. Even if i cannot point to 1 specific issue this is guaranteed to fix, i am rather certain this is an improvement. I will not baby the code forever though.

backend/src/main/java/org/booklore/task/tasks/TempFetchedMetadataCleanupTask.java
balazs-szucs: why?

balazs-szucs: they are doing the same thing

imnotjames: A good rule of thumb is to try to follow the single responsibility principal (the S in SOLID).  In this case, it would mean one PR makes one and only one logical change (business logic, configuration, technical change).

This helps code review, future debugging, and most importantly, communication.  "Why was this change made" can be made very clear - because it was needed for the singular task described.

That's not always possible, but in this case it is.

backend/src/main/java/org/booklore/repository/AnnotationRepository.java
imnotjames: I think that's because in some cases that's "better" because a `DELETE` will block vs the `SELECT` which won't.

You use the select with what _could_ be a slow query to get the list of primary key IDs & then do a FAST delete with the primary keys - rather than a blocking longer delete which does the query + delete.

In this case probably fine.

imnotjames (approved):

PR #811, opened 2026-04-23
refactor(caddy): integrate Caddy as a reverse proxy and add precompression for frontend assets

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

This pull request introduces a change to the deployment architecture by adding a Caddy-based reverse proxy in the Docker image. The frontend build is now precompressed with multiple algorithms for improved performance, and the backend no longer embeds the frontend by default. Entrypoint and build scripts have been updated to support these changes and to allow for more flexible deployment options.

**Deployment and Proxy Integration:**
* Added a `Caddyfile` and a new Caddy layer to the Docker build, configuring Caddy as a reverse proxy for the frontend and backend, with static asset caching and precompressed file serving. (`packaging/docker/Caddyfile`, `Dockerfile`, 
* Updated `entrypoint.sh` to optionally launch Caddy alongside the backend, with logic to avoid port collisions and handle process lifecycles robustly. 

**Frontend Build and Compression:**
* Enhanced the frontend production build to precompress assets using gzip, Brotli, and Zstandard (when available), via a new `precompress-dist.mjs` script.

**Backend Build and Embedding:**
* Changed backend build so that the frontend is not embedded by default (`embedFrontendDist=false`), with a new build property to control this behavior.

## Release Notes

* **New Features**
  * Added optional Caddy reverse proxy integration for improved request handling and response caching.
  * Implemented automatic frontend asset precompression (gzip, brotli, zstd) to reduce bandwidth and improve load times.
  * Enhanced Docker deployment with optimized caching strategies and static file serving.


packaging/docker/Caddyfile
```diff
@@ -0,0 +1,54 @@
...
+        encode zstd gzip
+
+        @backend path /api /api/* /ws /ws/* /komga /komga/*
+        reverse_proxy @backend 127.0.0.1:{$GRIMMORY_BACKEND_PORT}
+
+        @workerFiles path /ngsw-worker.js /ngsw.json /safety-worker.js /worker-basic.min.js
+        handle @workerFiles {
+            header Cache-Control "public, no-cache"
+            file_server {
+                precompressed br gzip
```
pyrox0: Per https://caddyserver.com/docs/caddyfile/directives/file_server#precompressed this is basically the default value so you can remove this from every file_server that sets this.

packaging/docker/Caddyfile
```diff
@@ -0,0 +1,54 @@
+{
+    auto_https off
+    admin off
+}
+
+:{$GRIMMORY_HTTP_PORT} {
+    route {
+        root * /srv/frontend
+        encode zstd gzip
```
pyrox0:
```suggestion
        encode
```
This is the default value

packaging/docker/Caddyfile
balazs-szucs: Thanks for the review, addressed in: [8e80ad7](https://github.com/grimmory-tools/grimmory/pull/811/commits/8e80ad74555e556c145886da6c8f7f2224c7d112)

imnotjames:
On the general direction, I am in agreement.  On the implementation, it makes sense but I have a lot of comments and questions in a bunch of different directions.

I see different things happening here:

* Supporting an API build without the frontend
* Running multiple services in a docker file reliably with a reverse proxy to host the frontend
* Cache control headers on frontend assets
* Pre-processing of frontend assets

I'll follow up later tonight focusing on these individually.

balazs-szucs:
> On the general direction, I am in agreement. On the implementation, it makes sense but I have a lot of comments and questions in a bunch of different directions.
> 
> I see different things happening here:
> 
>     * Supporting an API build without the frontend
> 
>     * Running multiple services in a docker file reliably with a reverse proxy to host the frontend
> 
>     * Cache control headers on frontend assets
> 
>     * Pre-processing of frontend assets
> 
> 
> I'll follow up later tonight focusing on these individually.

I am not sure i am following you here. I do not believe I see some of those thing happening here 🤔

oleastre:
First time commenting on this projet, but everyone needs to start somewhere :-) 

Having build a lot of docker image for various professional projects, most of the time we try to avoid having multiple services in the same docker image.

For applications with a static frontend, most of the time, I build a static folder in the app image, and the delegate serving those to another container. For that, I use either `volumes_from` with a readonly mount in the case of a docker-container setup, or an init container that copies content in case of a kubernetes setup.

The way I see it, is that this PR could be split in 2 different parts: 
- offer the frontend content in a fixed folder in the grimmory image
- document the recommended way of serving the files with a caddy setup in docker compose or using a reverse proxy.

balazs-szucs:
> First time commenting on this projet, but everyone needs to start somewhere :-)

Hi, welcome :)


> Having build a lot of docker image for various professional projects, most of the time we try to avoid having multiple services in the same docker image.

i agree with the premise, i am not sure how to satisfy the separation of concern AND not make even more complicated to deploy the software. There are people already having problem with the DB/FE + BE docker setup, i imagine adding this as another container would make that even worse. We are not serving enterprise people, moreso homer-labbers/hobbyist etc...

> For applications with a static frontend, most of the time, I build a static folder in the app image, and the delegate serving those to another container. For that, I use either volumes_from with a readonly mount in the case of a docker-container setup, or an init container that copies content in case of a kubernetes setup.

yeah, i think we would have lot of problems explaining to some of the user-base

> The way I see it, is that this PR could be split in 2 different parts:
> 
> - offer the frontend content in a fixed folder in the grimmory image
> - document the recommended way of serving the files with a caddy setup in docker compose or using a reverse proxy.

i am not insisting on this PR, if you want to help out, feel free to. Right i am inclined to keep it as is, though. I do not see that huge of disadvantages to already putting Caddy into to the mix at our level, and already serving compressed/well configured cached static resources, making it trivial for users to get benefits of this with little (from where i am standing) downsides.

NamLe029:
> i agree with the premise, i am not sure how to satisfy the separation of concern AND not make even more complicated to deploy the software. There are people already having problem with the DB/FE + BE docker setup, i imagine adding this as another container would make that even worse. We are not serving enterprise people, moreso homer-labbers/hobbyist etc...

Could you explain more about the increased complexity? I still don't understand how adding just one more service in the docker-compose file would complicate the deployment process.
Is the issue raised so significant that it overcomes the structural advantages brought by the split?

balazs-szucs:
> Could you explain more about the increased complexity? I still don't understand how adding just one more service in the docker-compose file would complicate the deployment process.
> Is the issue raised so significant that it overcomes the structural advantages brought by the split?



We are mostly talking about services where each container has to be spin up on it's own. This includes:
- (Some) compose UIs (Portainer standalone, Yacht, etc.) operating in per-container mode require a separate setup step per service.
- Unraid manages Docker via individual GUI templates with no native Compose support. Each additional container means another template to configure manually.
- Proxmox users typically run LXCs, where adding more containers increases per-service configuration overhead and those ppl are already mad at me

and these are just from my the top of my mind. In practice, i image other services that i am less knowledgeable about, may cause further headaches, e.g., saltbox

And even with plain docker-compose it's already noticeably more annoying, then just having 1 container, the two services need to be on the same network, DB credentials (in-case of the DB) have to be passed around via env vars, and you need to make sure the DB actually starts before the app does (depends_on + health checks). Nothing impossible, but each of these trips people up, i imagine similar problem would be even worse when we add 1 more to the mix. These are generally not problems for the tech-savvy people running the most basic setup, but you put avarage hobbyist + lets say unraid together, and 1-2 hours off my life, each time.

oleastre:
Maybe a good solution would be to have an "all in one" container that contains everything (db + frontend + backend) and a "basic" one for more advanced users.

balazs-szucs:
> Maybe a good solution would be to have an "all in one" container that contains everything (db + frontend + backend) and a "basic" one for more advanced users.

Now we are talking. I think this seems to be a good idea. Though it does require much more work than this but i like it. @imnotjames thoughts? (if you have any?)

imnotjames:
> > Maybe a good solution would be to have an "all in one" container that contains everything (db + frontend + backend) and a "basic" one for more advanced users.
> 
> Now we are talking. I think this seems to be a good idea. Though it does require much more work than this but i like it. @imnotjames thoughts? (if you have any?)

An omnibus image is usually considered the simpler image approach for stuff like this but it has a lot of drawbacks.

More easily resource constrained, harder to debug, generally more complex to build.

I've used s6-overlay for supporting multiple background services.  The biggest issue is getting all of the logs properly output to stdout/stderr.

balazs-szucs:
> An omnibus image is usually considered the simpler image approach for stuff like this but it has a lot of drawbacks.
> 
> More easily resource constrained, harder to debug, generally more complex to build.
> 
> I've used s6-overlay for supporting multiple background services. The biggest issue is getting all of the logs properly output to stdout/stderr.

So, what's your call here? Seems to me like a bit of a cryptic message.

balazs-szucs:
Due to lack of interest, closing.

Issue #815, opened 2026-04-23
Fix lubimyczytac + ranobedb + audible app book rating filters


zachyale:
## Summary
The book browser no longer uses the old local `sidebar-filter` utility for app book filtering. The mentioned 3 rating filters are also not passed to the B/E for filtering. Therefore, we should:

- remove the deprecated local sidebar filter helper and its test
- verify that any behaviour previously handled locally is now covered by the backend filter path
- close the remaining frontend-to-backend mapping gaps for sidebar filters

## Why
During some testing I found that the old local sidebar filter path was effectively dead code for the book browser, but if someone didn't test sufficiently/understand that, they might on first pass try to update that dead code (🙃)

## Scope
- remove the deprecated local sidebar filter utility and related dead tests
- verify that all visible app book sidebar filters are wired through `BookBrowserComponent` into `AppBookFilters`

PR #817, opened 2026-04-23
fix(filters): fix app file size filtering for books + make the f/e aware of the filter type

### LINKED ISSUE #808: App Book file size filter doesn't work + is only ever treated as an OR
### Discussed in https://github.com/orgs/grimmory-tools/discussions/807

<sup>Originally posted by **zachyale** April 23, 2026</sup>
### What went wrong?

3 issues: 
1. The filter 500s due to an unsupported SQL query generated by hibernate
2. When using the app book file size filter on the book browser, `AND` is never handled properly on the F/E
3. The filter only accounts for the file with the lowest file ID for a given book

### How can we reproduce it?

1. Open the book browser and the developer network monitor
2. Try to filter by file size
3. Note the 500, note the request is always as an `OR`

Stable

### Your setup

- Grimmory stable

_No response_

### Before submitting

zachyale:
## Description

Fixes app book file size filtering so it works on MariaDB and evaluates all book format files instead of just the file w/ the lowest file ID.

**Linked Issue:** Fixes https://github.com/grimmory-tools/grimmory/issues/808

## Changes

The previous file size filter was broken in 2 ways:
1. The backend query used a `MIN(id)` subquery comparison to identify a single file row, which was rendered into an SQL shape that Maria rejected, causing a 500 when filtering by file size.
2. The filter only considered one book file- this meant a book could be excluded even when one of its other valid formats matched the current size bucket. for ex., if a book had a 5 MB EPUB (primary, file ID 1) and a 500 KB PDF (file id 2), filtering for 0-1 MB should still include that book

The changes also add regression coverage in a new `AppBookSpecificationTest` for the following cases:
- matching against the first book-format file only
- `not` behavior
- multi-select `or`
- multi-select `and`

Notes:
- We should backfill this new test in a subsequent PR to guard against future regressions in other filters after this is merged
- The `sidebar-filter` is no longer used anywhere since the move to backend-based filtering. This work made it clear we should should kill it in a subsequent PR/do a scan to ensure all filters previously available on the F/E logic are fully available via the new approach. Follow up issue: https://github.com/grimmory-tools/grimmory/issues/815

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file size filtering logic to correctly handle multiple filter modes and ensure accurate book selection based on file size criteria.

* **Tests**
  * Added comprehensive integration tests for file size filtering functionality to validate correct behavior across various filtering scenarios.

### DIFF
--- a/backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
+++ b/backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
@@ -634,38 +634,53 @@ public class AppBookSpecification {
         return (root, query, cb) -> {
             List<Integer> ids = parseIntList(rangeIds, "fileSize");
             if (ids.isEmpty()) return cb.conjunction();
-            
-            Subquery<Long> sub = query.subquery(Long.class);
-            Root<BookEntity> subRoot = sub.correlate(root);
-            Join<BookEntity, BookFileEntity> bfJoin = subRoot.join("bookFiles", JoinType.INNER);
-            
-            Subquery<Long> minIdSub = query.subquery(Long.class);
-            Root<BookEntity> minIdRoot = minIdSub.correlate(root);
-            Join<BookEntity, BookFileEntity> minIdJoin = minIdRoot.join("bookFiles", JoinType.INNER);
-            minIdSub.select(cb.min(minIdJoin.get("id")))
-                    .where(cb.equal(minIdJoin.get("isBookFormat"), true));
 
-            Expression<Long> size = bfJoin.get("fileSizeKb");
+            if ("and".equals(mode)) {
+                List<Predicate> predicates = new ArrayList<>();
+                for (Integer id : ids) {
+                    Subquery<Long> sub = query.subquery(Long.class);
+                    Root<BookFileEntity> bfRoot = sub.from(BookFileEntity.class);
+
+                    sub.select(bfRoot.get("book").get("id"))
+                            .where(
+                                    cb.equal(bfRoot.get("isBookFormat"), true),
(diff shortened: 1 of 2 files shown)


backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
```diff
@@ -634,38 +634,53 @@ public static Specification<BookEntity> withFileSizes(List<String> rangeIds, Str
...
+            case 0 -> cb.and(cb.greaterThanOrEqualTo(size, 0L), cb.lessThan(size, 1024L));
+            case 1 -> cb.and(cb.greaterThanOrEqualTo(size, 1024L), cb.lessThan(size, 10240L));
+            case 2 -> cb.and(cb.greaterThanOrEqualTo(size, 10240L), cb.lessThan(size, 51200L));
+            case 3 -> cb.and(cb.greaterThanOrEqualTo(size, 51200L), cb.lessThan(size, 102400L));
+            case 4 -> cb.and(cb.greaterThanOrEqualTo(size, 102400L), cb.lessThan(size, 512000L));
+            case 5 -> cb.and(cb.greaterThanOrEqualTo(size, 512000L), cb.lessThan(size, 1048576L));
+            case 6 -> cb.and(cb.greaterThanOrEqualTo(size, 1048576L), cb.lessThan(size, 2097152L));
+            case 7 -> cb.greaterThanOrEqualTo(size, 2097152L);
+            default -> throw new APIException("Invalid fileSize bucket ID: " + id, HttpStatus.BAD_REQUEST);
         };
```
balazs-szucs, line 684: org.springframework.util.unit.DataSize

is what you are looking for here

PR #833, opened 2026-04-24
chore(deps):  update Spring Boot version to 4.0.6 and adjust testing behavior for new version

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Summary by CodeRabbit

* **Chores**
  * Upgraded Spring Boot Gradle plugin to version 4.0.6.

* **Tests**
  * Adjusted file download tests to accept a simplified UTF‑8 filename header format while retaining encoded filename support, improving robustness of download header validation.

### DIFF
--- a/backend/src/test/java/org/booklore/service/book/BookDownloadServiceTest.java
+++ b/backend/src/test/java/org/booklore/service/book/BookDownloadServiceTest.java
@@ -21,6 +21,8 @@ import java.util.List;
 import java.util.Optional;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.Mockito.mockStatic;
 import static org.mockito.Mockito.when;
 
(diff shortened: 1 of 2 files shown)


backend/src/test/java/org/booklore/service/book/BookDownloadServiceTest.java
```diff
@@ -71,8 +73,6 @@ public void downloadBook_includesContentDispositionAscii() {
 
     @Test
     public void downloadBook_includesContentDispositionUTF8() {
-        String expected = "attachment; filename=\"=?UTF-8?Q?=C9=87xample.epub?=\"; filename*=UTF-8''%C9%87xample.epub";
```
imnotjames, line 74: Looks like this was changed by https://github.com/spring-projects/spring-framework/pull/36328

Can we instead check this so we understand when something like this changes again?

```
        String expected = "attachment; filename=\"_xample.epub\"; filename*=UTF-8''%C9%87xample.epub";
```

The fact that the utf8 is `_xample.epub` in the fallback is important for us to track, right?

balazs-szucs: Sure thing, thanks for the review.

imnotjames (approved):
nit pick, the title as `update Srping Boot version to 4.0.6` on its own would be reasonable as it's updating the tests to match the new behavior introduced by 4.0.6.

When I originally read it I thought we were changing some validation logic in the app.

balazs-szucs:
> nit pick, the title as `update Srping Boot version to 4.0.6` on its own would be reasonable as it's updating the tests to match the new behavior introduced by 4.0.6.
> 
> When I originally read it I thought we were changing some validation logic in the app.

Fair! I adjusted the title :)

Good to keep release notes more informative via better PR titles.

PR #841, opened 2026-04-24
feat(pdf-reader): PDF reader for vertical/infinite scroll layouts and improved responsiveness

### LINKED ISSUE #838: Update to 3.0 is missing horizontall scroll in pdf view (viewer mode).
### Discussed in https://github.com/orgs/grimmory-tools/discussions/786

<sup>Originally posted by **saiwal** April 23, 2026</sup>

Earlier version had a convenient horizontal scroll when viewing a pdf file which is missing in 3.0. Changing to 'doc mode' does give the option to scroll horizontally but the editing mode interferes when reading document (links are by default not navigable, instead ask for options/edits).

Would request to add a scroll layout option in book view similar to doc mode.

makes viewing document on wide screens convenient.

### Anything else?

_No response_

### Want to help out?

Just sharing the idea for now

### Before submitting

balazs-szucs:
## Description

**Linked Issue:** Fixes #838

## Changes

This pull request introduces a new scroll layout (vertical/horizontal) option for the PDF reader, enhances accessibility by adding ARIA attributes to toolbar buttons, and expands the toolbar overflow menu with additional controls. There are also corresponding updates to the backend and frontend models to support the new scroll layout setting.

**PDF Reader Functionality and UI:**

* Added a scroll layout toggle button to the PDF reader toolbar, allowing users to switch between vertical and horizontal scrolling modes. The button updates its icon and label based on the current mode. The scroll layout state is also visually reflected in the viewer container.
* Expanded the toolbar overflow menu with new controls, including toggles for sidebar, annotation tools, scroll layout, bookmarks, spread mode, zoom presets, theme, fullscreen, and close actions, each with appropriate icons and labels. 

**Model Updates:**

* Updated the frontend `PdfViewerSetting` interface to include an optional `scrollLayout` property, supporting both 'vertical' and 'horizontal' values.
* Updated the backend `EbookReaderSetting.PdfReaderSetting` class to include a `scrollLayout` property, ensuring the user's preference is saved and loaded.

## Summary by CodeRabbit

* **New Features**
  * PDF scroll layout preference (vertical/horizontal) with toolbar toggle, settings UI, persisted viewer behavior, and embed/frame messaging.

* **Improvements**
  * Accessibility: added aria labels and pressed states for reader controls.
  * UI/UX: unified toolbar buttons, new reader icons, touch/tap styling, refined scroll/overflow behavior, DPR/tiling/rendering tweaks for small viewports.
  * Sync: more reliable initial page restore and pan/iframe messaging.

* **Bug Fixes**
  * Close-reader flow serialized and disables controls during close.

* **Tests**
  * Added unit tests validating annotation serialization/parsing and opacity handling.


frontend/src/assets/embedpdf-frame.html
```diff
@@ -164,23 +175,19 @@
...
-      [class*="border-t"][class*="bg-bg-surface"][class*="px-4"][class*="py-1"] {
+      [class*="border-t"][class*="bg-bg-surface"][class*="px-4"][class*="py-1"],
+      [class*="border-t"][class*="bg-bg-surface"][class*="px-4"][class*="py-2"],
+      [role="contentinfo"],
+      [data-epdf-i*="footer"],
+      [data-epdf-i*="status"],
+      [class*="fixed"][class*="bottom-0"][class*="inset-x-0"],
+      [class*="sticky"][class*="bottom-0"],
+      [class*="bottom-0"][class*="border-t"],
+      [class*="bottom-0"][class*="bg-bg-surface"] {
```
imnotjames: Why matching classes like this?

backend/src/main/java/org/booklore/model/dto/BookLoreUser.java
```diff
@@ -190,6 +190,7 @@ public static class EbookReaderSetting {
         public static class PdfReaderSetting {
             private String pageSpread;
             private String pageZoom;
+            private String scrollLayout;
```
imnotjames, line 193: Is this basically a no-op?  It accepts the field but we do nothing with it.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -222,6 +226,18 @@ export class EmbedPdfBookService {
     this.applyLocale(localeCode || 'en');
   }
 
+  setScrollLayout(layout: PdfScrollLayout): void {
+    const page = this.currentPage;
+    this.scrollLayout = layout;
+    this.applyScrollLayout(layout);
+    // Use a short delay to allow the layout engine to recalculate before restoring position
+    setTimeout(() => this.scrollToPage(page, 'instant'), 80);
```
imnotjames: There's no events or anything for this?

We have to hard code a wait?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -416,6 +432,49 @@ export class EmbedPdfBookService {
     this.i18n.setLocale('en');
   }
 
+  private applyScrollLayout(layout: PdfScrollLayout): void {
+    if (!this.scroll) return;
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const scrollLayoutCap = this.scroll as any;
```
imnotjames: Why do we need to cast as any here

imnotjames (commented):
This seems like a `feat` not a `refactor` - what do you think?

balazs-szucs:
> This seems like a feat not a refactor - what do you think?

i am not attached to either. I can change to whatever you'd like.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: because of the underlying library does not specificy it either.  I am assuming this is some problem due to WASM or something

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: hi, my bad. It was needed because some of the UI elements need some time after rendering to realise, i can make this "more elengant" i think. Apologies.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: fixed

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: fixed

frontend/src/assets/embedpdf-frame.html
balazs-szucs: fixed

backend/src/main/java/org/booklore/model/dto/BookLoreUser.java
balazs-szucs: we will need it, when we add the stuff to the settings menu

frontend/src/app/features/readers/pdf-reader/pdf-reader.component.ts
```diff
@@ -806,6 +859,17 @@ export class PdfReaderComponent implements OnInit, OnDestroy {
...
   }
 
+  cycleScrollLayout(): void {
+    const nextLayout: PdfScrollLayout = this.scrollLayout() === 'vertical' ? 'horizontal' : 'vertical';
+    this.scrollLayout.set(nextLayout);
+    if (this.viewerMode() === 'book') {
+      this.embedPdfBook.setScrollLayout(nextLayout);
+    } else if (this.embedPdfIframe?.contentWindow) {
+      this.embedPdfIframe.contentWindow.postMessage({ type: 'setScrollLayout', layout: nextLayout }, location.origin);
+    }
```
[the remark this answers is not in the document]
balazs-szucs: this may lead to other issues. We are assuming the two usage are so seperate that we do not wanna translate on to the other.

frontend/src/assets/embedpdf-frame.html
```diff
@@ -448,6 +468,46 @@
...
+    if (msg.type === 'setPanMode') {
+      const panPlugin = registry?.getPlugin('pan')?.provides();
+      if (panPlugin) {
+        if (msg.enable) {
+          panPlugin.enablePan();
+        } else {
+          panPlugin.disablePan();
+        }
+      }
+    }
```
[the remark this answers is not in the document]
balazs-szucs: done

backend/src/main/java/org/booklore/model/dto/BookLoreUser.java
imnotjames: Ah, it's stored as a JSON blob in the database so it's okay to add a field and do nothing with it -- since it will be stored in the database because it's on this model.  That makes more sense.

frontend/src/app/features/readers/ebook-reader/shared/icon.component.ts
```diff
@@ -203,6 +218,78 @@ const ICONS: Record<ReaderIconName, IconPath[]> = {
   'link': [
     {d: 'M10 13a5 5 0 0 0 7.54.54l3-3a5 5 0 0 0-7.07-7.07l-1.72 1.71'},
     {d: 'M14 11a5 5 0 0 0-7.54-.54l-3 3a5 5 0 0 0 7.07 7.07l1.71-1.71'}
+  ],
+  'sidebar': [
+    {d: 'M3 3h18v18H3z'},
+    {d: 'M9 3v18'}
+  ],
```
imnotjames, line 225: This is not saying we need to do anything to this PR.  This is a note for the future that would make a PR like this easier to review:

This would be a great refactor for its own unit of work.  It would reduce the size and complexity of this PR, and as far as I can tell is unrelated to the singular goal of this PR.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -416,6 +484,47 @@ export class EmbedPdfBookService {
     this.i18n.setLocale('en');
   }
 
+  private applyScrollLayout(layout: PdfScrollLayout): void {
+    if (!this.scroll) return;
+    const scrollLayoutCap = this.scroll as ScrollCapability & ScrollLayoutCapability;
```
imnotjames: Couldn't all this function checking be done via a type guard?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -608,23 +716,20 @@ export class EmbedPdfBookService {
...
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
-    const w = window as any;
+    const w = this.getMutableWindow();
     if (w.__grimmoryOrigDprDescriptor) {
       Object.defineProperty(window, 'devicePixelRatio', w.__grimmoryOrigDprDescriptor);
       delete w.__grimmoryOrigDprDescriptor;
     } else if (Object.getOwnPropertyDescriptor(window, 'devicePixelRatio')?.configurable) {
-      // eslint-disable-next-line @typescript-eslint/no-explicit-any
-      delete (window as any)['devicePixelRatio'];
+      Reflect.deleteProperty(window as Window & typeof globalThis, 'devicePixelRatio');
```
imnotjames: This needs the cast?  We use this in other places without the cast.

frontend/src/app/features/readers/pdf-reader/utils/annotation-converter.spec.ts
```diff
@@ -0,0 +1,81 @@
...
+  return {
+    annotation: {
+      type: 9,
+      pageIndex: 0,
+      id: 'annotation-1',
+      rect: {origin: {x: 1, y: 2}, size: {width: 10, height: 12}},
+      segmentRects: [{origin: {x: 1, y: 2}, size: {width: 10, height: 12}}],
+      color: '#FFFF00',
+      opacity,
+    } as never,
```
imnotjames, line 16: As never is a peculiar cast.  Is this trying to get around typing restrictions?  This is a test file, so it's fine either way, I'm just curious.

frontend/src/app/features/readers/pdf-reader/utils/annotation-converter.ts
```diff
@@ -23,7 +48,15 @@ export function parseStoredAnnotations(jsonString: string | null | undefined): A
 
     // New format: { format: 'embedpdf', version: 1, annotations: [...] }
     if (parsed && parsed.format === 'embedpdf' && Array.isArray(parsed.annotations)) {
-      return parsed.annotations.map(restoreArrayBuffers);
+      const storedItems = parsed.annotations as SerializableAnnotationTransferItem[];
```
imnotjames, line 51: Why do we need to update the annotation converter to implement the vertical/infinite scroll feature?

frontend/src/assets/embedpdf-frame.html
```diff
@@ -448,6 +471,46 @@
...
+        let handled = false;
+        const restorePage = () => {
+          if (handled) return;
+          handled = true;
+          scrollPlugin.scrollToPage({ pageNumber: page, behavior: 'instant' });
+        };
+
+        if (typeof scrollPlugin.onLayoutReady === 'function') {
+          const unsub = scrollPlugin.onLayoutReady(() => {
+            if (typeof unsub === 'function') unsub();
```
imnotjames: Is this duplicative of above?

frontend/src/app/features/settings/reader-preferences/pdf-reader-preferences/pdf-reader-preferences-component.ts
```diff
@@ -50,4 +55,13 @@ export class PdfReaderPreferencesComponent {
     this.userSettings.pdfReaderSetting.pageZoom = value;
     this.readerPreferencesService.updatePreference(['pdfReaderSetting', 'pageZoom'], value);
   }
+
+  get selectedScrollLayout(): 'vertical' | 'horizontal' {
+    return this.userSettings.pdfReaderSetting.scrollLayout || 'vertical';
```
imnotjames: Should this use `??`

```suggestion
    return this.userSettings.pdfReaderSetting.scrollLayout ?? 'vertical';
```

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -19,12 +19,45 @@ import type {
...
 }
 
+export type PdfScrollLayout = 'vertical' | 'horizontal';
+
+type ScrollLayoutMethod = (layout: PdfScrollLayout) => void;
+type ReadScrollLayoutMethod = () => string;
+
+interface ScrollLayoutCapability {
+  onLayoutReady?: (cb: () => void) => () => void;
+  setScrollStrategy?: ScrollLayoutMethod;
```
imnotjames: These first two are duplicative and you can omit them with...


```suggestion
interface ScrollLayoutCapability extends ScrollCapability{
```

right?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -19,12 +19,45 @@ import type {
...
+  setScrollMode?: ScrollLayoutMethod;
+  setLayoutMode?: ScrollLayoutMethod;
+  getScrollStrategy?: ReadScrollLayoutMethod;
+  getScrollMode?: ReadScrollLayoutMethod;
+  getLayoutMode?: ReadScrollLayoutMethod;
```
imnotjames, line 41: I don't see these functions ever defined on the scroll capabilities object that is returned.  Neither in browser, nor over in the [embedpdf scroll plugin code](https://github.com/embedpdf/embed-pdf-viewer/tree/main/packages/plugin-scroll/src).

Where are these being defined?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -222,6 +260,35 @@ export class EmbedPdfBookService {
...
+    // We skip the current replayed value and wait for the next emission.
+    this.layoutReady$.pipe(skip(1), take(1)).subscribe(() => {
+      clearTimeout(timeoutId);
+      restorePage();
+    });
+  }
+
+  getScrollLayout(): PdfScrollLayout {
+    return this.readScrollLayoutFromPlugin() ?? this.scrollLayout;
+  }
```
imnotjames: As far as I can tell, this function is never used.  Can it be removed?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -416,6 +484,47 @@ export class EmbedPdfBookService {
...
+      const mode = scrollLayoutCap.getScrollMode();
+      return mode === 'horizontal' ? 'horizontal' : 'vertical';
+    }
+
+    if (typeof scrollLayoutCap.getLayoutMode === 'function') {
+      const mode = scrollLayoutCap.getLayoutMode();
+      return mode === 'horizontal' ? 'horizontal' : 'vertical';
+    }
+
+    return null;
```
imnotjames: It looks like this method is never called.  Can it be removed in that case?

frontend/src/app/features/readers/pdf-reader/utils/annotation-converter.ts
balazs-szucs: Because of "any" usage that you flagged earlier.

frontend/src/app/features/readers/pdf-reader/utils/annotation-converter.spec.ts
balazs-szucs: yes.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: `ScrollCapability` (imported from `@embedpdf/snippet`) does not officially expose a method for setting scroll direction (vertical/horizontal) at runtime on the `ScrollCapability` type. Hence, the custom definitions

frontend/src/app/features/readers/ebook-reader/shared/icon.component.ts
balazs-szucs: but then sonarcloud complains. I also did not do it for fun. I am also not well versed in sonarcloud enough to make the failing CI go away when complains about code duplication

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: yes

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: works yes

frontend/src/app/features/settings/reader-preferences/pdf-reader-preferences/pdf-reader-preferences-component.ts
balazs-szucs: yeah then i guess will handle undefined etc... more explicitly.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
imnotjames: Right - but even during runtime I never see these functions.  As far as I can tell they never exist and are only defined in this interface.  Where are they coming from?  Are they being defined somewhere in this PR?

balazs-szucs: they are for the vertical vs horizontal scroll switch if i remember correctly.

balazs-szucs: ... is this a blocker?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -19,12 +19,45 @@ import type {
...
+
+interface DocumentOpenedEvent {
+  id: string;
+  pageCount?: number;
+}
+
+interface DocumentManagerCapability {
+  onDocumentOpened?: (cb: (ev: DocumentOpenedEvent) => void) => () => void;
+}
+
```
imnotjames, line 55: This seems to be a type provided by `@embedpdf/snippet` - are we able to use that one?

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
```diff
@@ -19,12 +19,45 @@ import type {
...
+  setLayoutMode?: ScrollLayoutMethod;
+  getScrollStrategy?: ReadScrollLayoutMethod;
+  getScrollMode?: ReadScrollLayoutMethod;
+  getLayoutMode?: ReadScrollLayoutMethod;
+}
+
+interface DocumentOpenedEvent {
+  id: string;
+  pageCount?: number;
+}
```
imnotjames, line 50: This type doesn't seem to match with the object during runtime.  `DocumentState` is the one defined by embedpdf which is closer to..

```typescript
interface DocumentOpenedEvent {
  id: string;
  document: {
    pageCount: number;
  } | null
}
```

If this is part of trying to avoid `any` maybe not worth the effort to try to fix it in this PR, and I was mistaken in my comment for the `any` removal.

You're - of course - always welcome to push back.  Especially in something like this, "This matches the previous code's behavior; maybe we can open an issue to fix it later" is always a possible option.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
imnotjames: Ah, wait - this was part of trying to avoid the `any`, right?

If that's leading to a lot of refactors because of previous issues in the code I think we can omit that.

While it's best practice to avoid `any`, it's also best practice to reduce code churn if we can. :shrug:

Your call either way to go with it.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
imnotjames: If this was done because of my previous comment with regards to the `any` type, we can probably omit the changes from `any` type and come back to try to correct the incorrect types in the future?

Something feels off but maybe it's a thing to fix another day.

imnotjames: > ... is this a blocker?

It's not a blocker, I am just trying to better understand what the intent of the code is.

If we have a bunch of functions we call that don't do anything because they're never defined it's an area I'd expect anyone reviewing to question.

imnotjames (approved):

balazs-szucs:
> If this was done because of my previous comment with regards to the any type, we can probably omit the changes from any type and come back to try to correct the incorrect types in the future?
> 
> Something feels off but maybe it's a thing to fix another day.
>      
> 
> It's not a blocker, I am just trying to better understand what the intent of the code is.
> 
> If we have a bunch of functions we call that don't do anything because they're never defined it's an area I'd expect anyone reviewing to question.

we can omit probably I think one of them, but plugins are loaded at runtime with no gurantees. Hence why `any` was used. Everything is dynamically loaded when it comes embedpdf plugins.

frontend/src/app/features/readers/pdf-reader/services/embedpdf-book.service.ts
balazs-szucs: i can do this yes.

PR #846, opened 2026-04-24
refactor(api): Add caching for static assets and API endpoints

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

This pull request introduces HTTP cache control headers for both static frontend resources and REST API responses, improving frontend performance and ensuring clients always receive up-to-date data where needed. The changes primarily enhance cache management for Angular build assets and API endpoints.

**Frontend static resource caching:**

* Added explicit `Cache-Control` policies for different types of static resources in `WebMvcConfig`:
  - Fingerprinted Angular bundles and media assets are now cached for one year and marked immutable for optimal performance.
  - Service worker and `index.html` files are set to require revalidation to ensure users receive the latest app version after deployment
  - Third-party assets like icons and libraries receive a one-day public cache for a balance between freshness and performance.

**API response caching:**

* The `/books` endpoint now returns a `Cache-Control: private, max-age=60` header, allowing clients to cache book lists for up to 60 seconds.
* The `/users/me` endpoint now returns a `Cache-Control: no-cache, no-store, must-revalidate` header, ensuring user profile data is never cached and always up-to-date.

## Summary by CodeRabbit

* **Performance**
  * Added comprehensive HTTP caching for static assets, icons, fonts, SPA fallback and media with tailored durations; many endpoints now support ETag-based conditional GETs (304) and explicit Cache-Control headers.
  * Server-side caching expanded to more data regions (larger capacity) with targeted eviction on writes; profile endpoints now explicitly prevent caching.

* **New Features**
  * Sidecar endpoint now returns structured sidecar metadata with caching semantics; reader/file endpoints expose last-modified checks.

* **Bug Fixes / Security**
  * Icon storage now validates paths to prevent traversal.

* **Tests**
  * Updated tests to account for caching/self-injection behavior.


backend/src/main/java/org/booklore/service/IconService.java
```diff
@@ -159,6 +159,17 @@ public void deleteSvgIcon(String svgName) {
         }
     }
 
+    private Path validateAndResolvePath(String filename) {
```
imnotjames, line 162: This seems like an unrelated addition to the change set - was this accidentally included?  It wasn't described.

imnotjames (commented):
This seems to be adding both HTTP caching and application level caching.  Is it possible to split those two, as you noted this is urgent?

backend/src/main/java/org/booklore/service/IconService.java
balazs-szucs: yeah i can delete.

balazs-szucs:
> This seems to be adding both HTTP caching and application level caching. Is it possible to split those two, as you noted this is urgent?

... if that finally get this moving

PR #253, opened 2026-03-28
fix(kobo-sync): use KoboSpan for progress sync

alexhb1:
## Description

This PR migrates to the use of KoboSpan data for progress tracking between Grimmory and Kobo. The existing implementation didn't work for two-way progression sync and failed to do anything other than show a dummy percentage figure in the Nickel menu. This new approach maps existing EPUB percentage + CFI data to a KoboSpan position within the sent KEPUB file, as well as the reverse to update Grimmory with accurate positioning.

Approx +1195 lines is production code (Most of this is heavily inspired by Komga for a tried and tested approach), another +1035 lines of added tests. 

**Linked Issue:** Fixes #36

## Changes

- New services for managing this KoboSpan based tracking: `KoboSpanMapExtractionService`, `KoboSpanMapService`, `KoboBookmarkLocationResolver`, `KoboReadingStateBuilder`, `KoboReadingStateService`, `CfiConvertor`. Exact behavior is largely similar to other projects which implement robust Kobo Sync support. 
- New `kobo_span_map` table that caches the extracted span map per book file, invalidated automatically via file hash when a book is re-imported
- New content_source_progress_percent column on `user_book_file_progress` to store progress from external sources 
- Two-way sync: web reader progress (CFI + percentage) is translated to a KoboSpan bookmark for the device, and Kobo progress is translated back to a percentage+CFI usable by Grimmory. Freshest data is used for the latest position, Kobo timestamp is determined from the moment the book is closed, not from the moment it is synced which could be some time later. 
- Batch path in `KoboEntitlementService` pre loads the span maps
- Fallback chain for position resolution. Tries CFI first, then chapter href, then global progress percentage
- Bumps the Kobo sync batch size for performance (5 -> 100)

I'm reasonably happy with the actual guts of the code, nothing here is reinventing what's already done elsewhere, but @balazs-szucs would appreciate a look at the database/migration pieces and whether that's up to scratch. It works fine but not sure if there's a better way to do this in-line with what you're working on. 

## Verification
Tested live with a Kobo Libra Colour to confirm full two-way position syncing is working, freshest data is used, and to the best of our ability the exact book position is translated between percentage+CFI and KoboSpan forms. Updates in both directions are also instant. 

Full backend + frontend tests ran and passed, including new tests. 

## Summary by CodeRabbit

* **New Features**
  * Span-map extraction, persistence and precise Kobo span-based bookmark resolution for more accurate bookmarks.

* **Improvements**
[...]

PR #859, opened 2026-04-25
fix(reader): fix kobo progress race + apply in-chapter progress

alexhb1:
## Description

Fixes a web reader race regression with Kobo sync (mentioned in #253), where Kobo-sourced in-chapter progress was not restored in the EPUB reader. This helps deal with Foliate's internal eagerness to emit `relocate` before all data is ready. 

## Changes

- Restores saved EPUB position through a dedicated initial restore path
- Navigates to the saved chapter href, then applies the in-chapter fraction from `contentSourceProgressPercent`
- Retries briefly while Foliate section data is still settling, then falls back to normal progress handling

## Release Notes

* **Bug Fixes**
  * Improved reliability when restoring your last reading position in ebooks with automatic retry capability for cases where the saved position cannot be immediately recovered
  * Enhanced progress tracking during navigation within ebook chapters
  * Better state management and memory optimization during reader sessions and shutdown


balazs-szucs (approved):
Works well, also

Thank you!

PR #864, opened 2026-04-25
fix(comicvine): update Comicvine ID handling for issues and volumes in metadata extraction

### LINKED ISSUE #863: Comicvine Metadata Search Alignment
### Discussed in https://github.com/orgs/grimmory-tools/discussions/852

<sup>Originally posted by **Noggog** April 25, 2026</sup>
### What went wrong?

When I search for items via comicvine, the meta suggestion swaps to an unrelated one when opening details.   Movie here for example:

https://github.com/user-attachments/assets/192402fe-2a3a-4183-9786-f0cda124ec9e

Feels like it only affects Comicvine metadata.  And also, it's some odd situation where the initial data is correct, but upon loading the detail view it suddenly swaps out to another metadata set (you can see it on that last section of the video)

### How can we reproduce it?

Wire up a comicvine API key, search for `Leviathan` like in my video, click into one of the comicvine entries.

Stable

### Your setup

- 3.0.0
- Podman
- Chrome

_No response_

### Before submitting

balazs-szucs:
## Description

**Linked Issue:** Fixes #863

## Changes

This pull request updates how Comicvine IDs are handled throughout the metadata extraction and writing flow, adding support for both issue and volume IDs with explicit prefixes. The changes ensure that Comicvine URLs and IDs are consistently parsed, stored, and generated, improving metadata accuracy and compatibility.

**Comicvine ID and URL handling improvements:**

* Updated the Comicvine URL pattern in `CbxMetadataExtractor` to support both `issue` and `volume` URLs, enabling extraction of IDs for both types.
* Modified `ComicvineBookParser` to:
  - Parse Comicvine IDs with prefixes (e.g., `4000-` for issues, `4050-` for volumes) and fetch details accordingly.
  - Store Comicvine IDs with explicit prefixes when building metadata for issues (`4000-<id>`) and volumes (`4050-<id>`). 

**URL generation logic:**

* Updated `CbxMetadataWriter` to generate the correct Comicvine URL based on the Comicvine ID prefix, ensuring links point to the appropriate issue or volume page.

## Summary by CodeRabbit

* **New Features**
  * Added broader ComicVine support for both issues and volumes, including prefixed IDs and correct issue/volume link generation.
  * Improved detailed metadata retrieval for volume and issue records and preserved original search ordering.

* **Bug Fixes**
  * Broader URL recognition and safer ID parsing to avoid mis-parsing and improve metadata accuracy and reliability.


backend/src/main/java/org/booklore/service/metadata/parser/ComicvineBookParser.java
```diff
@@ -597,7 +653,7 @@ private BookMetadata convertToBookMetadata(Comic comic, Comic volumeContext) {
 
         BookMetadata metadata = BookMetadata.builder()
                 .provider(MetadataProvider.Comicvine)
-                .comicvineId(String.valueOf(comic.getId()))
+                .comicvineId("4000-" + comic.getId())
```
imnotjames: Can this be made a constant?

backend/src/main/java/org/booklore/service/metadata/parser/ComicvineBookParser.java
```diff
@@ -626,7 +682,7 @@ private BookMetadata buildVolumeMetadata(Comic volume) {
         
         return BookMetadata.builder()
                 .provider(MetadataProvider.Comicvine)
-                .comicvineId(String.valueOf(volume.getId()))
+                .comicvineId("4050-" + volume.getId())
```
imnotjames: Same here, can this be a named constant?

backend/src/main/java/org/booklore/service/metadata/parser/ComicvineBookParser.java
```diff
@@ -384,7 +395,52 @@ private List<BookMetadata> searchIssuesInVolume(Comic volume, String issueNumber
...
+    private BookMetadata fetchVolumeDetails(int volumeId) {
+        String apiToken = getApiToken();
+        if (apiToken == null) return null;
+
+        URI uri = UriComponentsBuilder.fromUriString(COMICVINE_URL)
+                .path("/volume/4050-" + volumeId + "/")
+                .queryParam("api_key", apiToken)
+                .queryParam("format", "json")
+                .queryParam("field_list", VOLUME_FIELDS)
+                .build()
```
imnotjames: Does using the vars so it properly escapes make sense here?  Could be excessive, what do you think?

```suggestion
                .path("/volume/{prefix}-{volumeId}/")
                .queryParam("api_key", apiToken)
                .queryParam("format", "json")
                .queryParam("field_list", VOLUME_FIELDS)
                .buildAndExapand("4050", volumeId)
```

backend/src/main/java/org/booklore/service/metadata/parser/ComicvineBookParser.java
```diff
@@ -384,7 +395,52 @@ private List<BookMetadata> searchIssuesInVolume(Comic volume, String issueNumber
...
+        
+        int id;
+        try {
+            id = Integer.parseInt(idStr);
+        } catch (NumberFormatException e) {
+            log.warn("Invalid comicvineId format: {}", comicvineId);
+            return null;
+        }
+
+        if ("4050".equals(prefix)) {
```
imnotjames: Using a constant here would make the code a little clearer

backend/src/main/java/org/booklore/service/metadata/parser/ComicvineBookParser.java
```diff
@@ -201,15 +204,20 @@ private List<BookMetadata> searchVolumesAndIssues(String seriesName, String issu
         final String finalSeriesName = seriesName;
         log.debug("searchVolumesAndIssues: seriesName='{}', issueNumber='{}', year='{}'", finalSeriesName, issueNumber, extractedYear);
         
-        List<Comic> volumes = searchVolumes(finalSeriesName);
+        List<Comic> volumes = new ArrayList<>(searchVolumes(finalSeriesName));
```
imnotjames, line 221: Why does this need to be wrapped in an array list?

imnotjames (commented):
This all looks pretty good to me.  If we could add tests this would be fantastic.

backend/src/main/java/org/booklore/service/metadata/parser/ComicvineBookParser.java
balazs-szucs: To temporary save it, and re-order it based perceived relevance. If we do not wrap it, then we are left "accepting" the first hit. And since CV does not seem to function like a search engine (e.g., the order in which volumes are returned do not follow any particular pattern) unlike Goodreads/Amazon etc... it degrades the quality when we straight up accept the first thing.

imnotjames: Sorry, I don't understand - the code to get different volumes other than the first does not work without wrapping this in an array list?

`searchVolumes()` returns a list already.  How do the underlying concrete implementation change the behavior?

balazs-szucs: it does work. It does not allow to easily store it in memory, and then reorder.

imnotjames: The sort below fails? 

Edit: if so, can a test be written to validate this behavior works as expected?  And a comment added to ensure it's clear why we are wrapping this

balazs-szucs: The ArrayList creates a mutable copy of the results which makes .sort() later possible

balazs-szucs: List<Comic> searchVolumes(String seriesName) is not modifiable because it's used as a cache

imnotjames: > List searchVolumes(String seriesName) is not modifiable because it's used as a cache

Okay.

I was just confused because I would have expected to see an unsupported operation exception.

A comment here would make sense, or we could move the list copy operation into the `searchVolumes` method with the comment.

If that's the case, maybe we wouldn't need a test since we'd be validating the array list functionality at that point rather than how we're writing our code? 🤔  Not sure, that's getting a bit too philosophical for me to say one way or another..

balazs-szucs: 👍

imnotjames (approved):

PR #867, opened 2026-04-25
feat(metadata): import adjacent OPF metadata during scan

0xstillb:
﻿## Summary
- add support for importing adjacent external OPF metadata during library scans
- apply OPF metadata across supported file processors after embedded metadata extraction
- add focused tests for OPF parsing, file matching, and processor integration

## What this supports
This PR adds support for Calibre-style OPF files that live next to a book file, for example:

- `Book.pdf` + `metadata.opf`
- `Book.pdf` + `Book.opf`
- `Book.epub` + `metadata.opf`

The importer prefers a same-stem `.opf` file first, then falls back to `metadata.opf`.

## Metadata mapped from OPF
- title
- authors
- publisher
- published date
- description
- language
- categories
- isbn10 / isbn13
- calibre series name / series index

## Notes
- OPF parse failures are logged and do not fail the full scan
- this branch intentionally excludes fork-specific container publishing workflow changes

## Testing
- added `OpfMetadataExtractorTest`
- added `AdjacentOpfMetadataApplierTest`
- updated `AbstractFileProcessorTest`

Local targeted test command used:

```bash
./gradlew test --tests "org.booklore.service.metadata.extractor.OpfMetadataExtractorTest" --tests "org.booklore.service.metadata.AdjacentOpfMetadataApplierTest" --tests "org.booklore.service.fileprocessor.AbstractFileProcessorTest"
```

## Summary by CodeRabbit

* **New Features**
  * Books now automatically extract metadata from adjacent OPF files across all supported formats, including title, authors, categories, ISBN, series information, and publication dates. Processing continues seamlessly if extraction encounters errors.


balazs-szucs:
Okay, this needs a bit more discussion than what was initiated. Purely looking at the description, I already have some concerns. For example, what about custom metadata fields? How are fields that differ between Grimmory and Calibre connected to each other?

I also have some code quality concerns, mainly, there are a lot of weird patterns that I could already spot from doing a high-level parse.

Also, i am not quite convinced we should supporting this when, we already support at leaast 1 method from importing from Calibre that is, if you polish it/write Calibre stuff to the files.

0xstillb:
> Okay, this needs a bit more discussion than what was initiated. Purely looking at the description, I already have some concerns. For example, what about custom metadata fields? How are fields that differ between Grimmory and Calibre connected to each other?
> 
> I also have some code quality concerns, mainly, there are a lot of weird patterns that I could already spot from doing a high-level parse.
> 
> Also, i am not quite convinced we should supporting this when, we already support at leaast 1 method from importing from Calibre that is, if you polish it/write Calibre stuff to the files.

Thanks, this is very fair feedback.

To clarify the intent, I meant this PR to be much narrower than “Calibre support” in general. The goal here was only to read adjacent OPF files as a source for the small set of metadata fields that already map cleanly onto Grimmory’s existing model.

Right now that is limited to:
- title
- authors
- publisher
- published date
- description
- language
- categories
- isbn10 / isbn13
- calibre series / series_index

It does not attempt to import custom Calibre columns, and it does not try to define new mappings where there isn’t already an obvious Grimmory equivalent.

I also agree with you that embedded metadata is the preferred path when it exists. The use case I was trying to address is specifically exported libraries where the Calibre metadata is only present in adjacent OPF files, particularly for formats where writing metadata back into the file is incomplete, inconsistent, or not always practical.

balazs-szucs:
> It does not attempt to import custom Calibre columns, and it does not try to define new mappings where there isn’t already an obvious Grimmory equivalent.

but we do map those when it comes to embedded metadata. Whether that mapping is the best or not, i believe is highly subjective, but this is purely from my side is worrying, that you have not taken a look at how embedded metadata fields are handled, just pumped this out.

> To clarify the intent, I meant this PR to be much narrower than “Calibre support” in general. The goal here was only to read adjacent OPF files as a source for the small set of metadata fields that already map cleanly onto Grimmory’s existing model.

i know from experience, this never ends well. It is essentially signing up for never ending user tickets saying:

"hey, why doesn't this parse in X field or Y field" or "Buggy Calibre opf parser"

Doing stuff "half-way" is not a design choice that easy to explain end-users.

Previously we had the same exact problem with the embedded stuff. And we have had in the old upstream repo, too many tickets, too many conversations, too many time wasted because of this.


> The use case I was trying to address is specifically exported libraries where the Calibre metadata is only present in adjacent OPF files, particularly for formats where writing metadata back into the file is incomplete, inconsistent, or not always practical.

From my side, it is better to tell people no support for X path and use Y instead, rather than having well supported and maintained path, and a half baked one. Embedded metadata, as far as I know works well. It is also code I signed off, so i have brief understanding how it works.

balazs-szucs:
From what I understand from the Calibre ecosystem embedding metadata shouldn't be that big of a hardship towards end-users to justify this.

0xstillb:
> From what I understand from the Calibre ecosystem embedding metadata shouldn't be that big of a hardship towards end-users to justify this.

Thanks, that makes sense.

I understand the concern better now. My intention was to solve a narrow import case, but I agree that adding a second metadata path that does not behave the same way as the existing embedded metadata path can easily create confusing expectations for users.

I also agree that if Grimmory already has a supported Calibre-oriented path via embedded metadata, then adding adjacent OPF support “partially” may not be a good upstream design unless it is treated as a fully designed metadata source with the same mapping rules, precedence rules, tests, and documentation as the existing embedded metadata handling.

Given that, I am happy to not push this PR.

PR #882, opened 2026-04-25
refactor(jwt): remove Jackson 2 compatibility note and add unit tests for JWT utilities 

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

This pull request introduces a new unit test class for the JWT utility logic and makes a minor dependency ordering adjustment in the build configuration. The main focus is on improving test coverage for JWT-related functionality.

**Testing improvements:**

* Added a new test class `JwtUtilsTest` to verify access and refresh token generation, validation, and claims extraction for the `JwtUtils` class. This includes tests for extracting the username, user ID, and custom claims from tokens.

**Build configuration:**

* Reordered the `jackson-module-blackbird` dependency in `backend/build.gradle.kts` to be consistent with other Jackson dependencies.

## Summary by CodeRabbit

* **Tests**
  * Added authentication token generation and validation tests.

* **Chores**
  * Updated dependency configuration for compatibility.

### DIFF
--- a/backend/build.gradle.kts
+++ b/backend/build.gradle.kts
@@ -171,11 +171,8 @@ dependencies {
     implementation(platform("tools.jackson:jackson-bom:3.1.2"))
     implementation("tools.jackson.core:jackson-core")
     implementation("tools.jackson.core:jackson-databind")
-    implementation("tools.jackson.module:jackson-module-blackbird")
-
-    // --- Jackson 2 (Compatibility) ---
-    // jackson-annotations version is managed by Jackson 3 BOM (requires 2.20+)
     implementation("com.fasterxml.jackson.core:jackson-annotations")
+    implementation("tools.jackson.module:jackson-module-blackbird")
 
     // --- Caching ---
     implementation("org.springframework.boot:spring-boot-starter-cache")
--- a/backend/src/test/java/org/booklore/config/security/JwtUtilsTest.java
+++ b/backend/src/test/java/org/booklore/config/security/JwtUtilsTest.java
@@ -1,0 +1,68 @@
+package org.booklore.config.security;
+
+import io.jsonwebtoken.Claims;
+import org.booklore.model.entity.BookLoreUserEntity;
+import org.booklore.service.security.JwtSecretService;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
+import org.mockito.InjectMocks;
+import org.mockito.Mock;
+import org.mockito.junit.jupiter.MockitoExtension;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.when;
+
+@ExtendWith(MockitoExtension.class)
+class JwtUtilsTest {
+
+    @Mock
+    private JwtSecretService jwtSecretService;
+
+    @InjectMocks
+    private JwtUtils jwtUtils;
+
+    private BookLoreUserEntity user;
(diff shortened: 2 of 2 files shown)


backend/build.gradle.kts
```diff
@@ -171,11 +171,8 @@ dependencies {
     implementation(platform("tools.jackson:jackson-bom:3.1.2"))
     implementation("tools.jackson.core:jackson-core")
     implementation("tools.jackson.core:jackson-databind")
-    implementation("tools.jackson.module:jackson-module-blackbird")
-
-    // --- Jackson 2 (Compatibility) ---
-    // jackson-annotations version is managed by Jackson 3 BOM (requires 2.20+)
     implementation("com.fasterxml.jackson.core:jackson-annotations")
+    implementation("tools.jackson.module:jackson-module-blackbird")
```
imnotjames, line 175: This seems to be just moving the implementation definition down a few lines and removing a comment.  Is this actually changing the version installed?  I tried before and after and didn't see a change

backend/src/test/java/org/booklore/config/security/JwtUtilsTest.java
```diff
@@ -0,0 +1,68 @@
+package org.booklore.config.security;
```
imnotjames, line 1: Can this test change be its own PR?  I don't see how it's related to the Jackson change.  I tried it without the gradle change and it still passes

backend/build.gradle.kts
balazs-szucs: Because the comment makes no sense. It does not pull Jackson 2 AFAICT.

backend/src/test/java/org/booklore/config/security/JwtUtilsTest.java
balazs-szucs: According to the official docs this is "officially" Jackson 2 compatible only.

It is somewhat chaotic PR, yes.

balazs-szucs: By this i mean JJWT

backend/build.gradle.kts
balazs-szucs: Admittedly, may not have my best PR ever made. I guess, it is somewhat redundant :(

imnotjames: We don't pull all jackson 2 but `com.fasterxml.jackson.core` is Jackson 2 annotations, and `com.fasterxml.jackson.com:jackson-annotations` is in place because it supports the jackson 2 compat, right?  

```
$ ./gradlew dependencies | grep jackson-annotations
+--- com.fasterxml.jackson.core:jackson-annotations -> 2.21
```

Hence the comment above the implementation, because the Jackson 3 BOM enforces that this is a compatible known version for supporting compat for old versions.

I think this comment is still relevant?

More information:

[JSTEP-1](https://github.com/FasterXML/jackson-future-ideas/wiki/JSTEP-1#handling-of-jackson-annotations)

> Annotation evolution is slower, and so far they have been strictly backwards compatibility across both major versions

["Fix upcoming confusing versioning of jackson-annotations"](https://github.com/FasterXML/jackson-future-ideas/discussions/90)

> My concern is about the choice that has been made to have distinct Jackson 2 and Jackson 3 jackson-annotations which are almost the same, but not exactly, and to publish different versions for these 2 variants. I think that's creates uncessary confusion when there is a need to mix Jackson 2 and Jackson 3 "engines" in Frameworks (Spring Boot for example), in the community ecosystem, in end-users projects or at the scale of organizations that are using jackson-annotations to annotate the domain model and where the various microservices or applications will not migrate to Jackson 3 at the same time.

balazs-szucs: ah, you are right, thanks for the research!

Issue #853, opened 2026-04-25
Use the `org.grimmory` package name


imnotjames:
It would be helpful to match the java package name with our domain, `org.grimmory`.  This helps make it clear that this is a different application from others, and is another step in the right direction.

PR #899, opened 2026-04-26
refactor(api): use package imports in more situations

imnotjames:
## Description

For some reason, a number of references to packages are not using imports.  This makes it more difficult to do a clean rename to use `org.grimmory`, so fixing these imports is helpful.  this does not fix all cases of this - it seems some files do this more -- but it at least should get all cases of `org.booklore`

**Linked Issue:** Related to #853 

## Changes

* use `import` where possible for referencing packages

## Summary by CodeRabbit

* **Refactor**
  * Internal code cleanup: simplified import statements and replaced fully-qualified type references with imported types across backend modules for improved code readability.

* **Tests**
  * Updated test imports and type references to maintain consistency with codebase changes.

### DIFF
--- a/backend/src/main/java/org/booklore/service/progress/ReadingProgressService.java
+++ b/backend/src/main/java/org/booklore/service/progress/ReadingProgressService.java
@@ -1,7 +1,6 @@
 package org.booklore.service.progress;
 
-import
-        org.booklore.config.security.service.AuthenticationService;
+import org.booklore.config.security.service.AuthenticationService;
 import org.booklore.exception.ApiError;
 import org.booklore.model.dto.*;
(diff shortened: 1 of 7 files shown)


backend/src/main/java/org/booklore/service/progress/ReadingProgressService.java
```diff
@@ -1,7 +1,6 @@
 package org.booklore.service.progress;
 
-import
-        org.booklore.config.security.service.AuthenticationService;
+import org.booklore.config.security.service.AuthenticationService;
```
imnotjames, line 3: This one was weird and I fixed it here to simplify string replace

balazs-szucs (approved):

PR #911, opened 2026-04-26
refactor(security): replace JWT library with Nimbus JOSE and improve token validation

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

This pull request migrates JWT handling in the backend from the `jjwt` library to the `nimbus-jose-jwt` library. The migration includes updating the build dependencies, refactoring the `JwtUtils` class to use Nimbus APIs, and improving security and validation around JWT secret handling and claim extraction.

**Dependency changes:**

* Replaced the `jjwt` dependencies with `com.nimbusds:nimbus-jose-jwt:10.9` in `backend/build.gradle.kts` and removed the OAuth2 client starter, as it's no longer required for JWT handling.

**JWT utility refactor:**

* Refactored `JwtUtils.java` to use Nimbus JOSE JWT APIs for token generation, validation, and claim extraction, replacing all usages of the previous `io.jsonwebtoken` (`jjwt`) library. 

**Security improvements:**

* Added strict validation of the JWT secret length (minimum 32 bytes for HS256) at startup and before signing/verifying tokens, with appropriate logging and error handling.
* Improved error handling and logging throughout token generation, validation, and claim extraction to provide clearer diagnostics and fail safely. 

## Summary by CodeRabbit

* **Chores**
  * Switched JWT/OIDC implementation and updated auth handling.
* **New**
  * Added a dedicated invalid-JWT error mapped to 401 Unauthorized.
* **Bug Fixes**
  * Enforced stronger secret length and added startup validation to catch misconfiguration.
  * Tightened token parsing, signature and claims verification; tokens with malformed/expired/invalid claims now yield clear 401 responses.
* **Behavior**
  * Enforced stricter issuer validation for incoming tokens.


backend/src/main/java/org/booklore/config/security/JwtUtils.java
```diff
@@ -57,45 +93,93 @@ public String generateRefreshToken(BookLoreUserEntity user) {
...
+        }
 
-        // Graceful issuer enforcement: reject tokens with a wrong issuer,
-        // but allow legacy tokens that have no issuer claim (issued before this check).
-        // After a full token rotation cycle (~30 days), switch to hard .requireIssuer().
+        // Check issuer
         String issuer = claims.getIssuer();
         if (issuer != null && !"booklore".equals(issuer)) {
-            throw new JwtException("Invalid issuer: " + issuer);
+            throw new RuntimeException("Invalid issuer: " + issuer);
```
imnotjames: Does this lead to a 5xx level error?  Should we make it a 4xx level?

backend/src/main/java/org/booklore/config/security/JwtUtils.java
```diff
@@ -57,45 +93,93 @@ public String generateRefreshToken(BookLoreUserEntity user) {
...
-                .verifyWith(getSigningKey())
-                .build()
-                .parseSignedClaims(token)
-                .getPayload();
+    /**
+     * Checks if claims are valid (not expired and correct issuer).
+     * @throws RuntimeException if claims are invalid.
+     */
+    private void validateClaims(JWTClaimsSet claims) {
+        // Check expiration
```
imnotjames: Is there a helper in jose to do this?

Usually there is because there's more to check than just `exp` - `nbf` is another.  The validator usually also handles clock skew.

Perhaps `JwtTimestampValidator` or maybe nimbus has something else built in?

backend/src/main/java/org/booklore/config/security/JwtUtils.java
```diff
@@ -57,45 +93,93 @@ public String generateRefreshToken(BookLoreUserEntity user) {
...
+        if (!signedJWT.verify(verifier)) {
+            throw new RuntimeException("Invalid token signature");
+        }
+        return signedJWT;
+    }
+
+    /**
+     * Validates the token's signature, expiration, and issuer.
+     */
     public boolean validateToken(String token) {
```
imnotjames, line 134: If we are changing up libraries, should we be using something built in to spring?  Are there spring built in oauth2 / jwt helpers that already exist?

balazs-szucs: JOSE is what spring on default. Hence why we changed to that.

backend/src/main/java/org/booklore/config/security/JwtUtils.java
balazs-szucs: thanks for the review. Addressed: [37b048a](https://github.com/grimmory-tools/grimmory/pull/911/commits/37b048a00d8f51e4bb4fdf9f4f62eb0150a9f545)

backend/src/main/java/org/booklore/config/security/JwtUtils.java
balazs-szucs: thanks for the review. Addressed: [37b048a](https://github.com/grimmory-tools/grimmory/pull/911/commits/37b048a00d8f51e4bb4fdf9f4f62eb0150a9f545)

backend/src/main/java/org/booklore/config/security/JwtUtils.java
```diff
@@ -57,45 +108,92 @@ public String generateRefreshToken(BookLoreUserEntity user) {
...
+    /**
+     * Parses and verifies the JWT token signature.
+     * Does NOT check for expiration or other claims.
+     */
+    private SignedJWT parseAndVerify(String token) throws Exception {
+        SignedJWT signedJWT;
+        try {
+            signedJWT = SignedJWT.parse(token);
+        } catch (Exception e) {
+            throw ApiError.JWT_INVALID.createException("Malformed token");
```
imnotjames, line 121: This should probably log an error, too, right?


```suggestion
            log.error("Malformed token", e);
            throw ApiError.JWT_INVALID.createException("Malformed token");
```

imnotjames (approved):
I have tested this thoroughly and see no regressions.

I also confirmed that it's not possible to use the `Algorithm.NONE` algorithm in this case, so that's good.  It fails at multiple different levels, so no leaks there.

PR #916, opened 2026-04-26
fix(regenerate-covers): wrap book query stream in transaction to prevent lazy init exception

### LINKED ISSUE #1079: Cover regeneration task fails with Hibernate lazy initialization error
### Discussed in https://github.com/orgs/grimmory-tools/discussions/909

<sup>Originally posted by **NominalHero0** April 26, 2026</sup>
### What went wrong?

Running the bulk cover regeneration task from the admin panel consistently fails with the following error:

`ERROR o.b.service.metadata.BookCoverService : Error during cover regeneration: Cannot lazily initialize collection of role 'org.booklore.model.entity.BookEntity.bookFiles' with key '129' (no session)`

The task fails every time it runs and appears to stop on this book rather than skipping it and continuing with the rest of the library.

Additional Details: 
`id: 138, book_id: 129, file_name: The Favorites A Novel - Layne Fargo (2025).pdf, book_type: PDF`

Relevant Logs:
`ERROR o.b.service.metadata.BookCoverService : Error during cover regeneration: Cannot lazily initialize collection of role 'org.booklore.model.entity.BookEntity.bookFiles' with key '129' (no session)`

### How can we reproduce it?

1. Go to Application tab in settings
2. Run the cover regeneration task
3. Error occurs consistently

Stable

### Your setup

Grimmory Version: latest (3.0.2) (as of April 26, 2026)
Installation Method: Docker via Docker Compose on Ubuntu 24.04 LTS

_No response_

### Before submitting

balazs-szucs:
## Linked Issue

Fixes #1079 

## Changes

This pull request refactors the `regenerateCovers` method in `BookCoverService` to ensure that the retrieval and filtering of book entities is executed within a transaction. This change improves data consistency and reliability during the cover regeneration process.

**Transaction management improvements:**

* The call to `bookQueryService.getAllFullBookEntities()` and the associated stream operations are now wrapped in a `transactionTemplate.execute(...)` block to ensure they are performed within a transactional context.

## Manual Testing Steps

1. Checked out code from branch
2. Opened "Application Settings" tab
3. Click on regenerate tasks
4. Observe no uncaught exceptions and regenerate covers completes

## Checklist

- [ ] I have disclosed any AI usage as per the organization AI Policy above.
- [ ] I understand all of my submitted changes.

## Summary by CodeRabbit

* **New Features**
  * Added a query and service method that return full book records with associated files.
* **Bug Fixes**
  * Safer cover regeneration: skips books without metadata or primary files and correctly respects cover-lock checks.
* **Refactor**
  * Book-loading updated to eagerly include file collections for fuller runtime data.
* **Tests**
  * Expanded regeneration tests to cover null metadata, missing primary files, and lock-related edge cases.


imnotjames (commented):
Looks good, but two small questions:

Is there an issue for this one?
How do we exhibit this / how do we test that it corrects the issue?

imnotjames:
> Looks good, but two small questions:
> 
> Is there an issue for this one? How do we exhibit this / how do we test that it corrects the issue?

Okay, answering my own question after spending time going through discord logs:

This seems to be related to https://github.com/orgs/grimmory-tools/discussions/909 - however, I cannot replicate it so I can't really validate that this corrects anything.

You validated it and confirmed this corrects the issue?

balazs-szucs:
> > Looks good, but two small questions:
> > Is there an issue for this one? How do we exhibit this / how do we test that it corrects the issue?
> 
> Okay, answering my own question after spending time going through discord logs:
> 
> This seems to be related to https://github.com/orgs/grimmory-tools/discussions/909 - however, I cannot replicate it so I can't really validate that this corrects anything.
> 
> You validated it and confirmed this corrects the issue?

Try that, before and after.

Screenshot_20260503_094405_Firefox

backend/src/main/java/org/booklore/service/book/BookQueryService.java
```diff
@@ -77,10 +77,31 @@ public List<Book> mapEntitiesToDto(List<BookEntity> entities, boolean includeDes
...
+                .filter(book -> !missingOnly || book.getBookCoverHash() == null)
+                .map(book -> new BookRegenerationInfo(book.getId(), book.getMetadata().getTitle()))
+                .toList();
+    }
+
+    private boolean isCoverLocked(BookEntity book) {
+        return book.getMetadata() != null && Boolean.TRUE.equals(book.getMetadata().getCoverLocked());
+    }
+
+    public record BookRegenerationInfo(Long id, String title) {}
```
imnotjames: When writing an API - in this case, a class that will be called by other classes - it's often best practice to reduce surface area as much as possible.  Best practices aren't rules, though.

We could probably keep BookRegenerationInfo in the BookCoverService and mark a getter in there as @Transactional(readOnly=true)?  I think we'd need to mark it as protected but that's better than public.  But it could reduce the size of the change set & cut down on the increased surface area.

What do you think?

imnotjames (commented):
With that I was able to replicate - I was going to the "tasks" section and clicking the wrong button.

I can also reproduce that this change fixes the issue, awesome.

backend/src/main/java/org/booklore/service/book/BookQueryService.java
balazs-szucs: yes, i know.

I also think that either way, it is out-of-scope for this PR. I am just glad we finally resolved this. I would mostly argue, this has (probably) the kind of general problem of "having X thing implement Y times" rather then having " X implemented once, and reused Y times"

imnotjames: Sure, I get that.

But the solution was `getAllFullBookEntitiesWithFiles`, not wrapping this in a transaction and copying the logic into `BookQueryService`, right?

That means the minimal fix could be ~8 lines (+ tests).  Less is more, but I know that's not always the most straightforward.

imnotjames (approved):
This has a few unrelated changes but generally it does fix the regenerate covers bug.

backend/src/main/java/org/booklore/service/book/BookQueryService.java
zachyale: I see james has approved it, so I won't put a blocking review on this- but as per my message to you earlier, can we either follow up with a ticket to address this in the future, or address this before merge?

balazs-szucs: that will not be needed because refactored to use James' solution, hence it is using the existing infra

PR #921, opened 2026-04-27
fix(metadata): bypass service worker for cross-origin cover thumbnails

### LINKED ISSUE #919: Metadata search and author match cover thumbnails fail to render after v3.0.0 service worker introduction
## Summary

On the **Metadata Search** and **Author Match** screens, cross-origin cover thumbnails (Goodreads, Amazon, etc.) silently fail to render. The `` element shows the broken-image placeholder despite the JPEG bytes arriving with HTTP 200 in the Network tab. Pasting the same URL in a new browser tab loads the image fine.

## Affected versions

- v3.0.0 onward (verified on v3.0.2)
- Reproduced in stock Chrome and Brave (with Shields disabled and all extensions disabled)
- Safari unaffected
- v2.3.1 unaffected (no service worker)

## Reproduction

1. Quit and reopen the browser (clean SW state).
2. Open any book → **Edit Metadata** → **Search Metadata** tab.
3. Run a search using a provider that returns external thumbnail URLs (Goodreads or Amazon).
4. Result cards show broken-image icons; DevTools Network shows successful `200`s for the JPEG requests, served via `ngsw-worker.js`.
5. Right-click any failing image's URL → open in a new tab → it loads correctly.

## Diagnosis

v3.0.0 introduced the Angular service worker (`frontend/ngsw-config.json`, `provideServiceWorker(...)` in `frontend/src/main.ts`). The SW intercepts all in-scope `fetch` events including cross-origin `` requests for metadata thumbnails. On first fetch the SW returns a response that Chromium's image decoder rejects, even though the bytes are valid JPEG (verified via DevTools Response → Hex Viewer: `FF D8 FF E0 ... JFIF`).

Hard refresh (`Cmd+Shift+R`) appears to fix it because Chromium bypasses the SW for the navigation request, but a fresh browser session reproduces the broken state.

Verified empirically by building three custom v3.0.2 images:

| Variant | Result |
|---|---|
| Removed `loading="lazy"` from the affected `` tags | ❌ Did not fix |
| Disabled the service worker entirely (`enabled: false` in `main.ts`) | ✅ Fixed |
| Applied `?ngsw-bypass=true` to the cross-origin thumbnail URLs (SW still active) | ✅ Fixed |

## Root cause

Default Angular SW behavior intercepts all in-scope fetches. For cross-origin URLs not declared in any `assetGroup` or `dataGroup`, the SW does a passthrough fetch, but on first load this passthrough produces a response object that Chromium will not paint into an ``. Same-origin cover URLs (e.g., book/series browsers) are unaffected because the SW handles them correctly via the `assets` `assetGroup` in `ngsw-config.json`.

## Affected components

- `frontend/src/app/features/metadata/component/book-metadata-center/metadata-searcher/metadata-searcher.component.html` (results grid ``)
- `frontend/src/app/features/author-browser/components/author-match/author-match.component.html` (author photo ``)

The `cover-search` component uses PrimeNG's `<p-image>` (different rendering pipeline) and is not affected.

## Proposed Fix

Apply Angular's documented SW escape hatch by appending `?ngsw-bypass=true` to the cross-origin thumbnail URLs. This pattern is already used elsewhere in the codebase: `frontend/src/app/features/readers/audiobook-player/audiobook.service.ts:33,42,47`.

Implementation will be a small standalone `NgswBypassPipe` (pure pipe) at `frontend/src/app/shared/pipes/ngsw-bypass.pipe.ts`, applied to the two affected templates. `loading="lazy"` is preserved. Same-origin images are not changed.

PR will be linked here once approved.

## Local verification

Custom-built `v3.0.2-ngsw-bypass` image deployed to a personal instance: covers render correctly across multiple sessions, browser restarts, and providers. SW remains enabled so PWA features are unaffected.

liquidate:
## Summary

Fixes #919.

Cover thumbnails on the **Metadata Search** and **Author Match** screens fail to render in Chromium browsers (Chrome, Brave). JPEG bytes arrive with HTTP 200 in the Network tab but the `` element shows the broken-image placeholder; the same URL loads correctly in a standalone tab.

Root cause is that the Angular service worker added in v3.0.0 intercepts cross-origin image fetches and on first load returns a response that Chromium's image decoder rejects, even though the bytes are valid JPEG.

## Changes

* New `NgswBypassPipe` (standalone, pure) at `frontend/src/app/shared/pipes/ngsw-bypass.pipe.ts` plus Vitest spec.
* Applied the pipe to `metadata-searcher.component.html` and `author-match.component.html`. These are the two surfaces that load cross-origin metadata thumbnails dynamically via SSE/signal updates.
* Same-origin Grimmory cover URLs (book/series browsers, `cover-search`) are unchanged.
* `loading="lazy"` is preserved.

The `?ngsw-bypass=true` escape hatch is Angular's documented way to opt a request out of SW handling and is already used in this codebase by `audiobook.service.ts:33,42,47`.

## Verification

Empirically validated by building three custom `v3.0.2` images and testing in fresh browser sessions in Chrome and Brave:

| Variant | Covers render? |
|---|---|
| Removed `loading="lazy"` | No |
| Disabled the service worker entirely | Yes |
| Applied `?ngsw-bypass=true` (this PR) | Yes |

Local instance has been running the patched build for several hours across multiple sessions and provider searches with no regressions.

## Test plan

- [x] `corepack yarn typecheck` (clean)
- [x] `corepack yarn lint` (clean)
- [x] `corepack yarn test`: new pipe spec passes (5/5). 5 pre-existing failures in `magic-shelf-utils.spec.ts` (timezone date assertions) are unrelated to this change and reproduce on unmodified `develop`.
- [x] Manual: Brave fresh session, Search Metadata returns covers across multiple providers and books.
- [x] Manual: Chrome fresh session, same.
- [x] Manual: Author Match cover thumbnails render.

## Notes for reviewers

* The pipe uses simple string concatenation rather than the `URL` API, matching the existing `audiobook.service.ts` pattern. URL fragments would technically place the query in the wrong position but no metadata provider CDN returns URLs with fragments in practice.
* The pipe is not idempotent (calling it twice on the same URL produces `&ngsw-bypass=true&ngsw-bypass=true`). Harmless but worth noting.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Release Notes

* **Refactor**
  * Updated image loading mechanisms in author profiles and book metadata to improve how images are cached and displayed, ensuring more reliable and consistent presentation across the application.

* **Tests**
  * Implemented comprehensive test coverage for image caching functionality to verify proper behavior and reliability.

### DIFF
--- a/frontend/src/app/features/author-browser/components/author-match/author-match.component.html
+++ b/frontend/src/app/features/author-browser/components/author-match/author-match.component.html
@@ -70,7 +70,7 @@
           <div class="result-card">
             <div class="result-image">
               @if (result.imageUrl) {
-                <img [src]="result.imageUrl" [alt]="result.name" loading="lazy"/>
+                <img [src]="(result.imageUrl | ngswBypass)" [alt]="result.name" loading="lazy"/>
               } @else {
                 <div class="result-placeholder">
                   <i class="pi pi-user"></i>
--- a/frontend/src/app/features/author-browser/components/author-match/author-match.component.ts
+++ b/frontend/src/app/features/author-browser/components/author-match/author-match.component.ts
@@ -8,6 +8,7 @@ import {TranslocoDirective, TranslocoService} from '@jsverse/transloco';
 import {MessageService} from 'primeng/api';
 import {AuthorService} from '../../service/author.service';
 import {AuthorDetails, AuthorMatchRequest, AuthorSearchResult} from '../../model/author.model';
+import {NgswBypassPipe} from '../../../../shared/pipes/ngsw-bypass.pipe';
 
 interface RegionOption {
   label: string;
@@ -25,7 +26,8 @@ interface RegionOption {
     InputText,
     Select,
     ProgressSpinner,
-    TranslocoDirective
+    TranslocoDirective,
+    NgswBypassPipe
   ]
 })
 export class AuthorMatchComponent implements OnInit {
--- a/frontend/src/app/features/metadata/component/book-metadata-center/metadata-searcher/metadata-searcher.component.html
+++ b/frontend/src/app/features/metadata/component/book-metadata-center/metadata-searcher/metadata-searcher.component.html
@@ -161,7 +161,7 @@
               (keydown.enter)="onBookClick(metadata)">
               <div class="card-image">
                 <img
-                  [src]="metadata.thumbnailUrl || 'assets/images/missing-cover.jpg'"
+                  [src]="(metadata.thumbnailUrl | ngswBypass) || 'assets/images/missing-cover.jpg'"
                   [alt]="metadata.title"
                   loading="lazy"/>
                 <div class="card-overlay">
--- a/frontend/src/app/features/metadata/component/book-metadata-center/metadata-searcher/metadata-searcher.component.ts
+++ b/frontend/src/app/features/metadata/component/book-metadata-center/metadata-searcher/metadata-searcher.component.ts
@@ -14,6 +14,7 @@ import {MetadataPickerComponent} from '../metadata-picker/metadata-picker.compon
 import {BookMetadataService} from '../../../../book/service/book-metadata.service';
 import {Tooltip} from 'primeng/tooltip';
 import {TranslocoDirective} from '@jsverse/transloco';
+import {NgswBypassPipe} from '../../../../../shared/pipes/ngsw-bypass.pipe';
 
 @Component({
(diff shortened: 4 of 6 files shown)


imnotjames (changes requested):
Per the discussion https://github.com/orgs/grimmory-tools/discussions/920 I cannot reproduce the problem you're describing.

Until we can confirm this is an issue affecting v3.0.2 we're unable to accept a pull request for it.

PR #926, opened 2026-04-27
refactor(metadata): avoid regex for AmazonBookParser

### LINKED ISSUE #924: CodeQL Reports inefficient Regular Expression in AmazonBookParser
The `AmazonBookParser` has an inefficient regular expression.

> Some regular expressions take a long time to match certain input strings to the point where the time it takes to match a string of length n is proportional to nk or even 2n. Such regular expressions can negatively affect performance, or even allow a malicious user to perform a Denial of Service ("DoS") attack by crafting an expensive input string for the regular expression to match.

The normal "regular expressions" are by nature not very efficient commentary applies, but this one can be replaced with processing the data using jsoup instead of a regex ahead of time.  Replace all of the `BR` processing with interacting with the document and this will be solved.

imnotjames:
## Description

this resolves the CodeQL warning that an inefficient regular expression is in place for AmazonBookParser and also avoids trying to parse HTML with a regular expression.  This defines whitespace as only blank text nodes & `br` elements.  In the future we could include other values or improve these in other ways, but this is close enough to match the existing behavior.

**Linked Issue:** Fixes #924

## Changes

* use jsoup rather than regex for cleaning up line breaks
* add test to confirm behavior

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved book description formatting for titles imported from Amazon, normalizing line breaks and trimming extraneous whitespace so content blocks render with consistent spacing.
* **Tests**
  * Added a unit test to verify normalized description output after HTML cleaning to prevent regressions.

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
@@ -884,18 +884,12 @@ public class AmazonBookParser implements BookParser, DetailedMetadataProvider {
     }
 
     private boolean isWhitespaceNode(Node node) {
-        if (node instanceof TextNode) {
-            if (((TextNode) node).isBlank()) {
-                return true;
-            }
+        if (node instanceof TextNode textNode) {
+            return textNode.isBlank();
         }
 
-        if (node instanceof Element) {
-            Element element = (Element) node;
-
-            if ("br".equals(element.tagName())) {
-                return true;
-            }
+        if (node instanceof Element element) {
+            return "br".equals(element.tagName());
         }
 
         return false;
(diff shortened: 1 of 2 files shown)


imnotjames:
This is related to Code scanning alert 126 but is frankly the best practice when dealing with HTML.  HTML is not a regular language, so regular expressions for parsing it becomes difficult if not outright impossible.

[More information here on why you should avoid parsi̶n̷g̷ HTML with a̶ ̴r̸e̵g̵u̵l̴a̵r̶ ̶e̷x̵p̷r̴e̴š̵̜̟̳̼͉̆̆̍̎͝s̸̤̰͈̭̑ì̵̡̭̖̟͒̾͊ō̸̧̻̿͜n̶͖̏͋](https://stackoverflow.com/questions/1732348/regex-match-open-tags-except-xhtml-self-contained-tags)

balazs-szucs (approved):

PR #937, opened 2026-04-27
refactor(stats): push bulk queries to streams, deduplicate aggregation logic

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

Refactors reading/listening session stat queries and their service-layer consumers to improve memory efficiency and correctness, and adds a missing `GET /listening/favorite-days` endpoint to bring audiobook stats to parity with reading stats.

- Convert `findPageTurnerSessionsByUser`, `findCompletionRaceSessionsByUserAndYear`, `findSessionStartTimesByUserAndPeriod`, and `findAllSessionStartTimesByUser` from `List` to `Stream` to avoid full result-set materialization
- Add `findDistinctReadingDatesByUser` to offload date deduplication to the DB
- Add `findReadingSessionTimestampsByUser` and `findListeningSessionTimestampsByUser` as explicit, book-type-filtered replacements for the shared `findSessionTimestampsByUser`

### Service

- Wrap all `Stream`-returning repository calls in `try-with-resources` to ensure cursor/connection release
- Update `groupByLocalDate` to accept `Stream<Instant>` directly, removing an unnecessary intermediate `collect(toList())`
- Add `getListeningFavoriteDays` service method

## Summary by CodeRabbit

* **New Features**
  * Added a listening analytics endpoint showing users’ favorite listening days by day-of-week with optional year/month filters.

* **Refactor**
  * Analytics reworked to stream session data for lower memory use and faster aggregation.
  * Page-turner and completion-race computations optimized to avoid full in-memory grouping.
  * Listening scatter pagination capacity increased for smoother, more comprehensive results.


backend/src/main/java/org/booklore/controller/UserStatsController.java
```diff
@@ -296,6 +296,19 @@ public ResponseEntity<List<PeakHoursResponse>> getListeningPeakHours(
         return ResponseEntity.ok(readingSessionService.getListeningPeakHours(year, month));
     }
 
+    @Operation(
+            summary = "Get listening favorite days",
+            description = "Returns listening activity distribution by day of week. Can be filtered by year and/or month.",
+            operationId = "getListeningFavoriteDays"
+    )
+    @GetMapping("/listening/favorite-days")
```
imnotjames, line 304: Was this endpoint missing?  Why are we adding it in a refactor?

backend/src/main/java/org/booklore/repository/ReadingSessionRepository.java
```diff
@@ -166,17 +173,48 @@ AND year(coalesce(ubp.dateFinished, ubp.readStatusModifiedTime, ubp.lastReadTime
...
+
+    @QueryHints(@QueryHint(name = "org.hibernate.fetchSize", value = "200"))
     @Query("""
             SELECT rs.startTime
             FROM ReadingSessionEntity rs
             WHERE rs.user.id = :userId
             ORDER BY rs.startTime
             """)
-    List<Instant> findAllSessionStartTimesByUser(@Param("userId") Long userId);
+    Stream<Instant> findAllSessionStartTimesByUserStream(@Param("userId") Long userId);
```
imnotjames: Why is this method suffixed with stream vs the others aren't?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -86,6 +87,40 @@ private List<ReadingSessionHeatmapResponse> groupByLocalDate(List<Instant> start
                 .collect(Collectors.toList());
     }
 
+    private List<PeakHoursResponse> computePeakHours(List<org.booklore.model.dto.SessionTimestampDto> sessions, ZoneId zone) {
+        Map<Integer, long[]> hourlyStats = new TreeMap<>(); // hour -> [count, duration]
+        for (var s : sessions) {
+            int hour = s.getStartTime().atZone(zone).getHour();
```
imnotjames, line 90: If this is a fix for a bug, should it be a separate PR to split the bug fixes from general refactors?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -859,7 +866,7 @@ public List<SessionScatterResponse> getListeningSessionScatter() {
         Long userId = authenticationService.getAuthenticatedUser().getId();
         ZoneId zone = ZoneId.systemDefault();
 
-        return readingSessionRepository.findListeningTimestampsByUserPaged(userId, PageRequest.of(0, 500))
+        return readingSessionRepository.findListeningTimestampsByUserPaged(userId, PageRequest.of(0, 1000))
```
imnotjames: Why increase this size?

backend/src/main/java/org/booklore/repository/ReadingSessionRepository.java
```diff
@@ -166,17 +173,48 @@ AND year(coalesce(ubp.dateFinished, ubp.readStatusModifiedTime, ubp.lastReadTime
...
             ORDER BY b.id, rs.startTime ASC
             """)
-    List<CompletionRaceSessionDto> findCompletionRaceSessionsByUserAndYear(
+    Stream<CompletionRaceSessionDto> findCompletionRaceSessionsByUserAndYear(
             @Param("userId") Long userId,
             @Param("year") int year);
 
+    @Query("""
+            SELECT rs.startTime as startTime,
+                   coalesce(rs.durationSeconds, 0) as durationSeconds
```
imnotjames, line 197: Isn't `durationSeconds` non-null?  If so, is this still needed?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: originally i wanted to increase everything to 1000, since it was inconsistent if i remember correctly. But then i walked back, and forgot this.

backend/src/main/java/org/booklore/controller/UserStatsController.java
balazs-szucs: because right now stats are being fetched in large queries, and it is also being eagerly loaded for some reason, so it longer term goal to both
- lazy fetch
- stream these as much as possible
- add smaller chunks, so that more is handled by the service and persistence layer and less dump on the client-side

backend/src/main/java/org/booklore/repository/ReadingSessionRepository.java
balazs-szucs: because it went from loading everything to memory, to stream, but may not necessary i guess.

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: it was claculated on the client side, if i remember correctly. Also since similar helpers already existed i copy pasted one a rewrote.

imnotjames: Right - but this PR is a refactor - but also a bug fix?  It may be harder to identify as a bug fix in the future.

This is important both for the release notes, but also for cases where we are identifying & fixing regressions.

Say I do a git bisect to identify where a regression takes place.  If I see this change, I see it's a refactor, and it causes a regression we'd obviously revert it.  But in doing so we'd be inadvertently re-introducing a bug.

balazs-szucs: it is not a bug fix. it does 3 essentially:

- refactor raw string SQL into criteria though this is still needs work.
- moves from dump everything to a more focused approach
- moves from let's load everything to memory into streaming approach.

imnotjames: But in the title

> fix timezone handling

and in the description you say

> Fix timezone correctness: hour and day-of-week extraction now always goes through atZone(zone) in Java; previous approach called DB-side hour() / day_of_week() on raw UTC instants, producing wrong results for users outside UTC

That's not fixing a bug?

balazs-szucs: i can remove it. Thanks!

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -421,7 +421,10 @@ public List<CompletionRaceResponse> getCompletionRace(int year) {
         BookLoreUser authenticatedUser = authenticationService.getAuthenticatedUser();
         Long userId = authenticatedUser.getId();
 
-        var allSessions = readingSessionRepository.findCompletionRaceSessionsByUserAndYear(userId, year);
+        List<CompletionRaceSessionDto> allSessions;
+        try (var sessionStream = readingSessionRepository.findCompletionRaceSessionsByUserAndYear(userId, year)) {
+            allSessions = sessionStream.collect(Collectors.toList());
```
imnotjames: I think another PR had switched these to use `toList()` directly.  Should this?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -86,6 +84,40 @@ private List<ReadingSessionHeatmapResponse> groupByLocalDate(List<Instant> start
...
+            hourlyStats.merge(hour, new long[]{1, s.getDurationSeconds()},
+                    (a, b) -> new long[]{a[0] + b[0], a[1] + b[1]});
+        }
+        return hourlyStats.entrySet().stream()
+                .map(e -> PeakHoursResponse.builder()
+                        .hourOfDay(e.getKey())
+                        .sessionCount(e.getValue()[0])
+                        .totalDurationSeconds(e.getValue()[1])
+                        .build())
+                .collect(Collectors.toList());
```
imnotjames:
```suggestion
                .toList();
```

backend/src/main/java/org/booklore/repository/ReadingSessionRepository.java
imnotjames: Right - but the others are called `findThingA` and this one is `findThingBStream`.  Should this just be `findThingB` to be consistent with the others?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: i think that wouldn't work because that is modified somewhere.

balazs-szucs: well i should look into it.

balazs-szucs: fixed

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: fixed

backend/src/main/java/org/booklore/service/ReadingSessionService.java
imnotjames: Given you forgot this, should it be moved back to 500?

backend/src/main/java/org/booklore/controller/UserStatsController.java
imnotjames: Okay - in the future this could make sense as its own PR then - so the refactor internally is separate from the new feature.

imnotjames (approved):

balazs-szucs:
> Given you forgot this, should it be moved back to 500?

Done

PR #941, opened 2026-04-27
refactor(ui): unify author/series/book browsers using tanstack virtual

### LINKED ISSUE #902: Unify virtual scrolling and layout across book / author / series browsers
The three browser pages have ended up with very different virtual scrollers and grid layouts - causing inconsistency, instability in some cases, and an ongoing source of maintenance in three different places. 

Three separate implementations: 
- Book browser is a custom solution that paginates when scrolling, it's not actually virtual. It means as the page gets longer and you continue to scroll down the book content loaded continues to grow and increases the risk of performance issues on large libraries. It's also an ongoing maintenance point to manage all this manually, scroll height calculations etc. 
- Author browser uses CDK's experimental autosize function
- Series browser uses fixed CDK scrolling, the autosize was unstable and caused problems. See #873 

Solution: 
Pick a standard virtualisation approach and use for all three pages. I've been playing with [Tanstack Virtual](https://tanstack.com/virtual/latest) and it works really well, especially with the book browser's pagination natively. It's lightweight and headless so the frontend can still handle its own styling. 

The idea would be to consolidate the grid and virtual scroller for the three different browser implementations, especially the book browser and all it's manual work with height calculations, pagination, etc. I think this would be a pretty big reduction in code, using a mature/maintained solution in TS virtual, and reducing this entire surface as a maintenance point.

alexhb1:
## Description

**Linked Issue:** Fixes #902

## Changes
- Replaces three individual browser virtual scroller implementations with Tanstack Virtual. Used across book, series, author, and author detail grids. 
- Simplified book browser grid rendering by removing manual scroll, resize, deferred render state, and estimated height functions. 
- New `virtual-grid.util.ts` which handles shared grid / virtual scrolling mechanics for all three browsers
- Reworked the size controls with a + / - control instead of a slider. Cards now fill available space with equal gaps. The mobile size option (choose between 2/3/4 columns) also works properly now. 
- Removed now unused utility code for array chunking, viewport column calculation, book-browser scroll storage, and deferred rendering.
- Updated tests

## Summary by CodeRabbit

* **New Features**
  - Discrete desktop density controls ("smaller" / "larger") for grid views.
  - New shared virtual-grid powering book, author, and series lists.
  - Scroll-position tracking that restores list scroll when navigating back.

* **Performance & UX**
  - Smoother, faster virtualized scrolling and rendering.
  - Cover/scale preferences persist immediately, clamped to allowed min/max.
  - Adjusting grid density preserves scroll position for uninterrupted browsing.


frontend/src/app/shared/components/grid-density-buttons/grid-density-buttons.component.ts
```diff
@@ -0,0 +1,32 @@
...
+        [title]="'common.decreaseCardSize' | transloco"
+        (click)="densityChange.emit('smaller')">−</button>
+      <button
+        type="button"
+        class="browser-grid-option-btn"
+        [attr.aria-label]="'common.increaseCardSize' | transloco"
+        [title]="'common.increaseCardSize' | transloco"
+        (click)="densityChange.emit('larger')">+</button>
+    </div>
+  `,
```
imnotjames: Don't we normally do the template as a separate HTML file?  Why embed it here?

frontend/src/app/features/author-browser/components/author-detail/author-detail.component.spec.ts
```diff
@@ -49,6 +49,16 @@ describe('AuthorDetailComponent', () => {
...
 
+  const setDescriptionContentRef = (
+    component: AuthorDetailComponent,
+    element: Pick<HTMLElement, 'scrollHeight' | 'clientHeight'>
+  ) => {
+    Object.defineProperty(component, 'descriptionContentRef', {
+      configurable: true,
+      value: signal({nativeElement: element} as ElementRef<HTMLElement>),
+    });
+  };
```
imnotjames: Is this to bypass the `readonly`?

frontend/src/app/features/book/components/book-browser/book-browser.component.ts
```diff
@@ -276,51 +272,15 @@ export class BookBrowserComponent implements AfterViewInit {
...
-  private readonly booksContextKey = computed(() => {
-    const {entityId, entityType} = this.entityInfo();
-    return Number.isNaN(entityId) ? entityType : `${entityType}:${entityId}`;
-  });
-
-  private readonly pipelineInputs = computed(() => {
+  readonly books = computed(() => {
     const books = this.appBooksApi.books();
     return this.seriesCollapseFilter.collapseBooks(books, this.forceExpandSeries());
   });
```
imnotjames, line 278: These can't be private anymore, right?

imnotjames (commented):
Testing this, it works great for the grid view.

The table view - because it's using a real table, probably - struggles a lot.  It tries to re-flow the table and that + the virtual grid leaves a lot to be desired.  In particular, scrolling leads to a flash of an empty page, the header disappears, and a second or two later it repaints.

Is there a path forward on the table view for improving that?

imnotjames:
Actually, once I switch to the table view and try to switch back I get the error `RangeError: Invalid array length`

I switched to the grid view, scrolled up and down a few times, stopped around the middle, then tried to switch back.  I can replicate it pretty regularly doing that.

```
An unexpected error occurred: RangeError: invalid array length
    Angular 7
        getMeasurements
        memoizedFunction
        calculateRange
        memoizedFunction
        getVirtualIndexes
        memoizedFunction
        getVirtualItems
<anonymous code>:1:147461
    overrideMethod <anonymous code>:1
    handleError global-error-handler.ts:44
    Angular 19
    BookBrowserComponent_ng_container_0_Template book-browser.component.html:93
    Angular 7
    RxJS 16

```

frontend/src/app/features/book/components/book-browser/book-browser.component.ts
```diff
@@ -276,51 +272,15 @@ export class BookBrowserComponent implements AfterViewInit {
...
-    onCleanup(() => {
-      clearTimeout(timeout);
-      this.booksRenderState.cancel(requestId);
-    });
-  });
   readonly isBooksLoading = computed(() => this.appBooksApi.isLoading());
   readonly booksError = this.appBooksApi.error;
   readonly bookIndexById = computed(() => {
     const books = this.books();
-    if (!books) return new Map<number, number>();
```
imnotjames: I think you can drop `bookIndexById` and `compareBookItems` as they're no longer in use.

frontend/src/styles.scss
```diff
@@ -67,6 +67,60 @@ html {
...
+.virtual-cell > app-book-card,
+.virtual-cell > app-series-card {
+  display: block;
+  width: 100%;
+  height: 100%;
+}
+
+.virtual-scroller-item {
+  position: relative;
+}
```
imnotjames: Should these all live within `.virtual-scroller` so they're namespaced both CSS-wise and organizationally?



```suggestion
.virtual-scroller {
  .virtual-content {
    position: relative;
    width: 100%;
  }
  
  .virtual-cell {
    position: absolute;
    top: 0;
    left: 0;
    display: flex;
    justify-content: center;
    box-sizing: border-box;
  }
  
  .virtual-cell > app-author-card,
  .virtual-cell > app-book-card,
  .virtual-cell > app-series-card {
    display: block;
    width: 100%;
    height: 100%;
  }
  
  .virtual-scroller-item {
    position: relative;
  }
}
```

frontend/src/app/shared/components/grid-density-buttons/grid-density-buttons.component.ts
alexhb1: Sonarcube / cloud failed me earlier because it was code duplication across author / series. can probably put it back or in HTML

imnotjames: If you had the `grid-density-buttons.component.html` it would complain about code duplication?

frontend/src/app/features/book/components/book-browser/book-browser.component.ts
alexhb1: Yep the html references it now

frontend/src/app/features/book/components/book-browser/book-browser.component.ts
alexhb1: Yeah looks like it, cheers

frontend/src/app/shared/components/grid-density-buttons/grid-density-buttons.component.ts
alexhb1: That should be fine I guess yeah, i'll move it

frontend/src/styles.scss
alexhb1: Actually will just have to move author detail grid to this css too first, then yeah it can live as `.virtual-scroller`

imnotjames (approved):
I tested this and it works well.  The code itself is clean.  I no longer see the issue with the table I was seeing before.

PR #967, opened 2026-04-28
refactor(metadata): simplify category and tag handling in CbxMetadata Extractor

balazs-szucs:
## Description

**Linked Issue:** Fixes #

## Changes

Reports Collection.addAll() and Map.putAll() calls immediately after an instantiation of a collection using a no-arg constructor.

Such constructs can be replaced with a single call to a parametrized constructor, which simplifies the code. Also, for some collections the replacement might be more performant.

Example:
```java
Set<String> set = new HashSet<>();
set.addAll(Arrays.asList("alpha", "beta", "gamma"));
```
After the quick-fix is applied:

```java
Set<String> set = new HashSet<>(Arrays.asList("alpha", "beta", "gamma"));
```
 
The JDK collection classes are supported by default. Additionally, you can specify other classes using the Classes to check panel.

## Summary by CodeRabbit

* **Refactor**
  * Streamlined metadata extraction for comic book files by simplifying how category and tag information is processed, improving internal efficiency and maintainability. No changes to visible behavior; users should see the same metadata while benefiting from slightly faster and more reliable processing behind the scenes.

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/extractor/CbxMetadataExtractor.java
+++ b/backend/src/main/java/org/booklore/service/metadata/extractor/CbxMetadataExtractor.java
@@ -132,12 +132,12 @@ public class CbxMetadataExtractor implements FileMetadataExtractor {
             builder.authors(authors);
         }
 
-        Set<String> categories = new HashSet<>(splitValues(getTextContent(document, "Genre")));
+        Set<String> categories = splitValues(getTextContent(document, "Genre"));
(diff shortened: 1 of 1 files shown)


imnotjames (approved):

imnotjames:
> The PR description is incomplete and off-topic. It contains boilerplate about Collection.addAll() patterns instead of describing the actual changes made to CbxMetadataExtractor.

I agree with coderabbit in this case but won't hold up anything for it.

PR #975, opened 2026-04-28
build: use vite 7.3.2 to avoid CVE-2026-39364

imnotjames:
## Description

given angular restricts us to vite 7, ensures that vitest also uses vite 7.3.2 to avoid CVE-2026-39364

this is in lieu of dependabot handling this because dependabot doesn't know what to do to resolve the issue

> [!NOTE]
> CVE-2026-39364 was only affecting our development environment, not the final build.
>
> This does not affect Grimmory the product, just Grimmory's developers.

## Changes

* update yarn lock manually to change transitive vite version
* drops all vite 8 dependencies

### DIFF
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -3147,13 +3147,6 @@ __metadata:
   languageName: node
   linkType: hard
 
-"@oxc-project/types@npm:=0.122.0":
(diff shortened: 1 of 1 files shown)


balazs-szucs (approved):
thank you

PR #978, opened 2026-04-28
build: use picomatch 4.0.4 to avoid CVE-2026-33672

imnotjames:
## Description

simplify angular dependencies to avoid CVE-2026-33672

this is in lieu of dependabot handling this because dependabot doesn't know what to do to resolve the issue

> [!NOTE]
> CVE-2026-33672 was only affecting our development environment, not the final build.
>
> This does not affect Grimmory the product, just Grimmory's developers.

## Changes

* update yarn lock manually to change transitive picomatch version

### DIFF
--- a/frontend/yarn.lock
+++ b/frontend/yarn.lock
@@ -210,7 +210,7 @@ __metadata:
   languageName: node
   linkType: hard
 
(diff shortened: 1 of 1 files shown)


balazs-szucs (approved):
thank you

PR #993, opened 2026-04-29
ci(release): publish Helm chart as OCI artifact on release

thibaultamartin:
## Description

This PR automatically publishes an OCI Helm chart when a release is cut. I could test that the image and chart are built and published (see https://github.com/thibaultamartin?tab=packages&repo_name=grimmory).

I could do a `helm install temp oci://ghcr.io/thibaultamartin/charts/grimmory:3.0.3-test` on my cluster. The chart behaved well but failed to pull Grimmory's image because the upstream image doesn't have a `v3.0.3-test`.

Discussed in #868 and https://github.com/grimmory-tools/grimmory/pull/881#discussion_r3142417221

**Linked Issue:** Fixes #

## Changes

1. Not hardcode gimmory's name and repo in the CI, so it's easier to test publishing images and charts in forks
2. Publish OCI Helm chart when a release is cut

## Summary by CodeRabbit

* **Chores**
  * Helm charts are now published automatically as part of the release process, enabling installation via the Helm package manager.
  * Release notifications are delayed until chart publishing completes, ensuring notifications reflect successful chart publication.

### DIFF
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -172,6 +172,7 @@ jobs:
     needs:
       - prepare-release-notification
     if: ${{ needs.prepare-release-notification.outputs.available == 'true' }}
+      - publish-helm-chart
     permissions: {}
     uses: ./.github/workflows/notify-discord-release-notes.yml
     secrets:
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -110,7 +110,7 @@ jobs:
           helm package deploy/helm/grimmory \
             --version "${VERSION}" \
             --app-version "${RELEASE_TAG}"
-          helm push "grimmory-${VERSION}.tgz" oci://ghcr.io/grimmory-tools/charts
+          helm push "grimmory-${VERSION}.tgz" "oci://ghcr.io/${{ github.repository_owner }}/charts"
 
   prepare-release-notification:
     name: Prepare Stable Release Notification
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -110,7 +110,7 @@ jobs:
           helm package deploy/helm/grimmory \
             --version "${VERSION}" \
             --app-version "${RELEASE_TAG}"
-          helm push "grimmory-${VERSION}.tgz" "oci://ghcr.io/${{ github.repository_owner }}/charts"
+          helm push "grimmory-${VERSION}.tgz" "oci://ghcr.io/${{ github.repository_owner }}/helm-charts"
 
   prepare-release-notification:
     name: Prepare Stable Release Notification
@@ -171,8 +171,8 @@ jobs:
     name: Notify Discord Release
     needs:
       - prepare-release-notification
-    if: ${{ needs.prepare-release-notification.outputs.available == 'true' }}
       - publish-helm-chart
+    if: ${{ needs.prepare-release-notification.outputs.available == 'true' }}
     permissions: {}
     uses: ./.github/workflows/notify-discord-release-notes.yml
     secrets:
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-release.yml
```diff
@@ -73,6 +73,45 @@ jobs:
...
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          set -eu
+          VERSION="${RELEASE_TAG#v}"
+          helm dependency update deploy/helm/grimmory
+          helm package deploy/helm/grimmory \
+            --version "${VERSION}" \
+            --app-version "${RELEASE_TAG}"
+          helm push "grimmory-${VERSION}.tgz" "oci://ghcr.io/${{ github.repository_owner }}/charts"
```
imnotjames: Making sure I understand: This would publish charts at `ghcr.io/grimmory-tools/charts/grimmory`, right?

That's what I am seeing when I try to run it.

My only concern would be that it could be unclear what a "chart" is for folks not in-the-know with kubernetes and could lead to a little confusion.  Not a blocker, but could be made better, right?

```suggestion
          helm push "grimmory-${VERSION}.tgz" "oci://ghcr.io/${{ github.repository_owner }}/helm-charts"
```

Might make it more clear what exactly this is, what do you think?  I'm going off a few open source projects (nextcloud, valkey) but I'm open to other suggestions.

imnotjames: This is assuming the chart name has to be last, I mean.

It might be nicer to have `ghcr.io/grimmory-tools/grimmory/helm` but I think that's not possible without some weird configurations..

thibaultamartin: > Making sure I understand: This would publish charts at ghcr.io/grimmory-tools/charts/grimmory, right?

Yes indeed!

> It might be nicer to have ghcr.io/grimmory-tools/grimmory/helm but I think that's not possible without some weird configurations.

Some projects also use a distinct repository for helm charts, something like `grimmory-tools/grimmory-helm` that only contains the helm chart. It's possible to trigger the CI of the helm repo whenever a release is cut in `grimmory-tools/grimmory` with [`repository_dispatch`](https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#repository_dispatch).

It's much more involved but also more cleanly separated. If Grimmory starts publishing charts I think it's important to take a "definitive" approach (choosing between in-repo chart, or a dedicated repo for charts, and not wobbling between the two) so admins don't have to upgrade their helm repos after installing a first version.

What do you reckon is best?

imnotjames: Yeah - I wouldn't want to change up where it is.  It's one thing to let renovate handle updating the helm versions, it's another for it to break because the helm chart is just missing.

I just wanted to ensure we don't move this after we create it.

Can we change it to the `helm-charts` for now that I suggested?  It should have limited change, just would show up as `grimmory-tools/helm-charts/grimmory` rather than the current `grimmory-tools/charts/grimmory`

thibaultamartin: Makes sense to me, I’ve accepted the changes. I can rebase on `develop` probably tomorrow, unless you want to do it before

imnotjames (approved):

imnotjames:
@thibaultamartin looks like there's some merge conflicts because we bumped some CI action versions.  If you can fix those this will be merged in.  Cheers!

.github/workflows/publish-release.yml
```diff
@@ -73,6 +73,45 @@ jobs:
...
+          username: ${{ github.actor }}
+          password: ${{ github.token }}
+
+      - name: Package and Push Helm Chart
+        env:
+          RELEASE_TAG: ${{ github.event.release.tag_name }}
+        run: |
+          set -eu
+          VERSION="${RELEASE_TAG#v}"
+          helm dependency update deploy/helm/grimmory
```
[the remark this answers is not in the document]
imnotjames: 🙄 if we hit this we will handle it then

PR #1057, opened 2026-05-03
chore(docker): drop mariadb container_name property

### LINKED ISSUE #155: [Bug] MariaDB name is not unique to Grimmory and can conflict with other containers
### What Went Wrong?

When a use tries to deploy Grimmory and another container already deployed has used `mariadb` as the database name, docker throws this error:

> Error response from daemon: Conflict. The container name "/mariadb" is already in use by container "xyz". You have to remove (or rename) that container to be able to reuse that name.

For example, this happens if someone tries to deploy Grimmory while having Booklore installed, since Booklore used `mardiadb` for the database name. Other apps might use this as well. 

Best practice is to give it a unique name like `grimmory-mariadb` to avoid conflicts.

1. Try to deploy a fresh installation of Grimmory while having another app that uses 'mariadb' as their database name (e.g. Booklore)

Grimmory should be able to install without conflicts with other containers based on database name. It should be unique to Grimmory.

_No response_

I'd suggest changing these for new installations:

In compose.yml:
```
mariadb:
    image: lscr.io/linuxserver/mariadb:11.4.5
    container_name: grimmory-mariadb   # <-- was just mariadb
```

In .env:
```
DATABASE_URL=jdbc:mariadb://grimmory-mariadb:3306/grimmory   # <-- was ...//mariadb:3306...
```

### Your Setup

- Grimmory version 2.3.0
- OS: macOS 15.7.3
- Browser: Firefox & Safari
- Installation: Docker compose
- Storage type: DAS

### Before Submitting

TheBrandonMorris:
## Description

Renamed `mariadb` instances to `grimmory_mariadb`

**Linked Issue:** Fixes #155

## Changes
Updated docker-compose as suggested by @obviouslyallie also included an update to the Readme.md as well

## Summary by CodeRabbit

* **Chores**
  * Removed explicit Docker container names from Compose configs so services use default Compose naming. This improves portability and reduces naming conflicts across environments.

### DIFF
--- a/deploy/compose/docker-compose.yml
+++ b/deploy/compose/docker-compose.yml
@@ -43,7 +43,7 @@ services:
 
   mariadb:
     image: lscr.io/linuxserver/mariadb:11.4.8
-    container_name: mariadb
+    container_name: grimmory_mariadb
     environment:
       - PUID=1000
       - PGID=1000
--- a/deploy/compose/docker-compose.yml
+++ b/deploy/compose/docker-compose.yml
@@ -43,7 +43,6 @@ services:
 
   mariadb:
     image: lscr.io/linuxserver/mariadb:11.4.8
-    container_name: grimmory_mariadb
     environment:
       - PUID=1000
       - PGID=1000
--- a/README.md
+++ b/README.md
@@ -144,7 +144,7 @@ services:
 
   mariadb:
     image: lscr.io/linuxserver/mariadb:11.4.5
-    container_name: mariadb
+    container_name: grimmory_mariadb
     environment:
       - PUID=${DB_USER_ID}
       - PGID=${DB_GROUP_ID}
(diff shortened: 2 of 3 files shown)


deploy/compose/docker-compose.yml
```diff
@@ -43,7 +43,7 @@ services:
 
   mariadb:
     image: lscr.io/linuxserver/mariadb:11.4.8
-    container_name: mariadb
+    container_name: grimmory_mariadb
```
imnotjames: Does everything still work if you omit the container name entirely?  If we can do that then you'd be able to use the default compose name generation.

We don't really need any of the container names by default as far as I can tell.

TheBrandonMorris: I'm inclined to agree. I can make those changes and push when I get back!

imnotjames (approved):

PR #1064, opened 2026-05-03
fix(koreader): web reader to koreader progress sync

### LINKED ISSUE #414: Koreader sync on Kindle does not work
### What Went Wrong?

I am intending to use Grimmory as the sync server between Koreader on my phone and Koreader on my jailbroken Kindle. I am able to connect to the sync server from Koreader on my phone and it works great to sync progress with Grimmory, but when I attempt to log in to the sync server on my Kindle it fails with "unknown server error". It's possible I'm missing something really basic, but I cannot figure it out.

Attempt to log into server on Kindle.
Log in fails.

Koreader on Kindle should log in to sync server without issue the same way Koreader on my phone does.

_No response_

_No response_

### Your Setup

Grimmory Version - 2.3.0
Koreader Version - 2026.03
Phone - Samsung zFold6
Kindle - 10th Gen Oasis

### Before Submitting

wesamjabali:
Fixes #414 

### a description of why you made this change

Kosync is an existing feature that promises 2 way sync, but it currently only syncs from the koreader app to grimmory and not the other way around. There have been multiple people complaining of this, myself one of them, and this PR fixes that issue. 

### how it fixes the issue you'd linked
#414 
Created a new functoin called `syncKoreaderFromBookloreProgress` which is called when new progress is made, and the correct settings are enabled. It uses all existing functionality to achieve getting the right location and setting the sync.

### how you tested this

1. Put up dev environment
2. add a book
4. enable OPDS
5. enable kosync
6. download the book to my kindle w/ koreader installed so the hashes match
7. make progress in koreader, view the progress matched on grimmory
8. make progress in grimmory, view the progress on koreader

## Summary by CodeRabbit

* **New Features**
  * KOReader sync now triggers when read progress includes a percentage and saves converted EPUB position when applicable.
  * Stores per-user KOReader progress and updates last-sync timestamp for synced users.

* **Tests**
  * Added tests ensuring KOReader sync runs when a percentage is present (EPUB/PDF cases) and is skipped when percentage is absent.

### DIFF
--- a/backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
+++ b/backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
@@ -6,6 +6,7 @@ import org.booklore.config.security.userdetails.KoreaderUserDetails;
 import org.booklore.exception.ApiError;
 import org.booklore.model.dto.progress.KoreaderProgress;
 import org.booklore.model.entity.*;
+import org.booklore.model.enums.BookFileType;
 import org.booklore.model.enums.ReadStatus;
 import org.booklore.repository.*;
 import org.booklore.service.hardcover.HardcoverSyncService;
@@ -90,6 +91,40 @@ public class KoreaderService {
             Float progressPercent = normalizeProgressPercent(koProgress.getPercentage());
             hardcoverSyncService.syncProgressToHardcover(book.getId(), progressPercent, authDetails.getBookLoreUserId());
         }
+    }
+
+    @Transactional
+    public void syncProgressToKoreader(Long bookId, Float percentage, Long userId) {
+        if (percentage == null) return;
+
+        koreaderUserRepository.findByBookLoreUserId(userId)
+                .filter(KoreaderUserEntity::isSyncEnabled)
+                .filter(KoreaderUserEntity::isSyncWithBookloreReader)
+                .ifPresent(koreaderUser -> {
+                    BookEntity book = bookRepository.findById(bookId).orElse(null);
+                    BookLoreUserEntity user = userRepository.findById(userId).orElse(null);
+                    if (book == null || user == null) return;
+
+                    UserBookProgressEntity progress = getOrCreateUserProgress(user, book);
+
+                    float koreaderPercent = percentage / 100.0f;
+                    progress.setKoreaderProgressPercent(koreaderPercent);
+                    progress.setKoreaderLastSyncTime(Instant.now());
+
+                    BookFileEntity primaryFile = book.getPrimaryBookFile();
+                    if (primaryFile != null && primaryFile.getBookType() == BookFileType.EPUB
+                            && progress.getEpubProgress() != null) {
+                        try {
+                            String xpointer = epubCfiService.convertCfiToProgressXPointer(
+                                    book.getFullFilePath(), progress.getEpubProgress());
+                            progress.setKoreaderProgress(xpointer);
+                        } catch (Exception e) {
+                            log.warn("Failed to convert CFI to XPointer for KOReader sync", e);
+                        }
+                    }
+
+                    progressRepository.save(progress);
+                });
     }
 
     private void saveToFileProgress(BookLoreUserEntity user, BookEntity book, UserBookProgressEntity progress) {
--- a/backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
+++ b/backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
@@ -12,8 +12,10 @@ import org.booklore.repository.*;
 import org.booklore.service.hardcover.HardcoverSyncService;
 import org.booklore.util.koreader.EpubCfiService;
 import org.springframework.http.ResponseEntity;
+import org.springframework.scheduling.annotation.Async;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.stereotype.Service;
+import org.springframework.transaction.annotation.Propagation;
 import org.springframework.transaction.annotation.Transactional;
 
 import java.math.BigDecimal;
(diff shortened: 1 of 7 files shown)


imnotjames:
Please include:

* a description of why you made this change
* how it fixes the issue you'd linked
* how you tested this

  Thanks!

wesamjabali:
Hey there, thanks for responding to this and the other PR so quickly. CodeRabbit actually made a couple good action items I'm working out now.

> * a description of why you made this change

Kosync is an existing feature that promises 2 way sync, but it currently only syncs from the koreader app to grimmory and not the other way around. There have been multiple people complaining of this, myself one of them, and this PR fixes that issue. 

> * how it fixes the issue you'd linked

Created a new functoin called `syncKoreaderFromBookloreProgress` which is called when new progress is made, and the correct settings are enabled. It uses all existing functionality to achieve getting the right location and setting the sync.

> * how you tested this

1. Put up dev environment
2. add a book
4. enable OPDS
5. enable kosync
6. download the book to my kindle w/ koreader installed so the hashes match
7. make progress in koreader, view the progress matched on grimmory
8. make progress in grimmory, view the progress on koreader

backend/src/main/java/org/booklore/service/progress/ReadingProgressService.java
```diff
@@ -267,6 +270,8 @@ public void updateReadProgress(ReadProgressRequest request) {
             progress.setDateFinished(request.getDateFinished());
         }
 
+        syncKoreaderFromBookloreProgress(userEntity, book, progress, percentage);
```
imnotjames: Looks like if you moved this down to line 278 you could avoid using `Float` and could have java auto-unbox the `Float` to `float`.

This would remove an extra `if` branch in your `syncKoreaderFromBookloreProgress` helper

backend/src/main/java/org/booklore/service/progress/ReadingProgressService.java
```diff
@@ -502,6 +507,32 @@ private List<BookStatusUpdateResponse> buildResetResponses(List<Long> bookIds, S
                 .toList();
     }
 
+    private void syncKoreaderFromBookloreProgress(BookLoreUserEntity user, BookEntity book,
+                                                   UserBookProgressEntity progress, Float percentage) {
+        if (percentage == null) return;
```
imnotjames: Similar to the hardcover service, perhaps much of this logic should live over in the `KoreaderService`?

Something like..

```
            void syncProgressToKoreader(long bookId, float percentage, long userId);
```

Then it's also a little easier to test directly, as it'd make sense as public over on the `KoreaderService`.

backend/src/main/java/org/booklore/service/progress/ReadingProgressService.java
wesamjabali: Makes sense. Moved it.

backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
```diff
@@ -92,6 +93,40 @@ public void saveProgress(String bookHash, KoreaderProgress koProgress) {
         }
     }
 
+    @Transactional
+    public void syncProgressToKoreader(Long bookId, Float percentage, Long userId) {
```
imnotjames: I think we can unbox these types.

```suggestion
    public void syncProgressToKoreader(long bookId, float percentage, long userId) {
```

backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
```diff
@@ -92,6 +93,40 @@ public void saveProgress(String bookHash, KoreaderProgress koProgress) {
...
+                    progress.setKoreaderProgressPercent(koreaderPercent);
+                    progress.setKoreaderLastSyncTime(Instant.now());
+
+                    BookFileEntity primaryFile = book.getPrimaryBookFile();
+                    if (primaryFile != null && primaryFile.getBookType() == BookFileType.EPUB
+                            && progress.getEpubProgress() != null) {
+                        try {
+                            String xpointer = epubCfiService.convertCfiToProgressXPointer(
+                                    book.getFullFilePath(), progress.getEpubProgress());
+                            progress.setKoreaderProgress(xpointer);
```
imnotjames: What happens if this fails to convert and the koreader progress progress is mismatched with koreader progress percent?

If that's a mismatch causes failure elsewhere, should we try to do the conversion first and then bail if it doesn't work as expected?

backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
```diff
@@ -92,6 +93,40 @@ public void saveProgress(String bookHash, KoreaderProgress koProgress) {
...
+                    float koreaderPercent = percentage / 100.0f;
+                    progress.setKoreaderProgressPercent(koreaderPercent);
+                    progress.setKoreaderLastSyncTime(Instant.now());
+
+                    BookFileEntity primaryFile = book.getPrimaryBookFile();
+                    if (primaryFile != null && primaryFile.getBookType() == BookFileType.EPUB
+                            && progress.getEpubProgress() != null) {
+                        try {
+                            String xpointer = epubCfiService.convertCfiToProgressXPointer(
+                                    book.getFullFilePath(), progress.getEpubProgress());
```
imnotjames: Instead of using the book full file path we should be explicit and use the `primaryFile` full file path, right?  That way if we ever have to change the book file to be something else (eg, we know a different file is being used by koreader) it's less churn to update it.

backend/src/main/java/org/booklore/service/koreader/KoreaderService.java
```diff
@@ -92,6 +93,40 @@ public void saveProgress(String bookHash, KoreaderProgress koProgress) {
...
+    public void syncProgressToKoreader(Long bookId, Float percentage, Long userId) {
+        if (percentage == null) return;
+
+        koreaderUserRepository.findByBookLoreUserId(userId)
+                .filter(KoreaderUserEntity::isSyncEnabled)
+                .filter(KoreaderUserEntity::isSyncWithBookloreReader)
+                .ifPresent(koreaderUser -> {
+                    BookEntity book = bookRepository.findById(bookId).orElse(null);
+                    BookLoreUserEntity user = userRepository.findById(userId).orElse(null);
+                    if (book == null || user == null) return;
```
imnotjames: We should probably log here - maybe a warn?  This probably shouldn't happen.

imnotjames:
I've tested this with koreader and it seems to work pretty well.  It's not perfect in some cases but it's close enough.

wesamjabali:
> I've tested this with koreader and it seems to work pretty well. It's not perfect in some cases but it's close enough.

I’ve spent a lot of time trying to make it more accurate and I simply can’t get it much better with the current framework. It might take some deeper digging to get it 100%, but this is an improvement. 

I see the comments you made and will get to them when I have the opportunity.

imnotjames:
This seems unrelated to #414

fxousselin:
# PR #1064 - Missing Kobo → KOReader Sync Support

## Issue

This PR successfully implements **Grimmory Reader → KOReader** sync with automatic CFI to XPointer conversion. However, **Kobo → KOReader** sync is not triggered.

## Root Cause

The `syncProgressToKoreader()` function is only called in `ReadingProgressService.updateReadProgress()`, which is triggered when:

- ✅ Reading in Grimmory web reader
- ✅ Manual progress updates via API

But **NOT** when:

- ❌ Kobo device syncs reading progress (uses `KoboLibrarySyncService`)

## Observed Behavior

**Test scenario:**

1. Read on KOReader to 10% → Push sync ✅
2. Read on Kobo device to 14% → Sync to Grimmory ✅
3. Pull sync on KOReader → **Still at 10%** ❌ (should be 14%)

**However:**

1. Read on KOReader to 10% → Push sync ✅
2. Read in **Grimmory web reader** to 15% ✅
3. Pull sync on KOReader → **Updated to 15%** ✅ (PR #1064 works!)

## Proposed Fix

Add a call to `KoreaderService.syncProgressToKoreader()` in `KoboLibrarySyncService` after updating Kobo reading progress.

### Suggested Code Location

In `KoboLibrarySyncService.java`, after updating `kobo_progress_percent`:

```java
// After updating Kobo progress in the database
if (koboProgressPercent != null) {
    koreaderService.syncProgressToKoreader(
        book.getId(), 
        koboProgressPercent, 
        user.getId()
    );
}
```

This would ensure that:

1. Kobo syncs its progress to Grimmory ✅
2. Grimmory automatically converts CFI → XPointer ✅
3. KOReader receives the updated position on next sync ✅

## Workaround

Currently using an external cron script that updates `koreader_progress_percent` based on `kobo_progress_percent`, but this doesn't benefit from the CFI → XPointer conversion that PR #1064 provides.

## Impact

This enhancement would make the bidirectional sync **fully automatic** for users with both:

- Kobo devices (native reading)
- KOReader instances (on Kobo or other devices)

Without this, users must read in Grimmory web reader to trigger the sync, which defeats the purpose of device-to-device synchronization.

---

**Tested on:** Grimmory fork `wesamjabali/grimmory:fix/koreader-web-reader-sync` (commit d412cedd)
**Device setup:** Kobo Libra Colour + KOReader

zachyale:
Since it's been over a month since any updates on this, I'm closing this PR. I've reopened it at https://github.com/grimmory-tools/grimmory/pull/1687 and have tagged you @wesamjabali, and included your commits from this branch so as to ensure you have proper attribution for your contributions :)

PR #1068, opened 2026-05-03
feat(helm): support external MariaDB

### LINKED ISSUE #1059: Helm chart doesn't support using an existing database
### What went wrong?

When trying a new deployment via the helm chart, I wanted to make it use a MariaDB I had deployed via the MariaDB operator.

It is possible to disable the MariaDB included by specifying `mariadb: false` in the values, and the [`Deployment` template](https://github.com/grimmory-tools/grimmory/blob/develop/deploy/helm/grimmory/templates/deployment.yaml#L64-L77) shows that it's possible to configure to configure the the credentials and database via the `mariadb.auth` section.

But it's not possible to specify the database name [in the database string](https://github.com/grimmory-tools/grimmory/blob/develop/deploy/helm/grimmory/templates/deployment.yaml#L66).

There should be a `mariadb.host` in the values to point to an existing database's service. A potential gotcha is to also update [the initContainer that waits for the db](https://github.com/grimmory-tools/grimmory/blob/develop/deploy/helm/grimmory/templates/deployment.yaml#L39-L40) to be up before launching Grimmory.

### How can we reproduce it?

1. Browse the helm chart
2. Realize that it's not possible to point Grimmory to an existing db

Stable

### Your setup

- Grimmoy v3.0.2
- Deployed via helm

_No response_

### Before submitting

_Originally posted by @thibaultamartin in https://github.com/grimmory-tools/grimmory/discussions/1048_

thibaultamartin:
## Description

This PR adds support for an external MariaDB database.

**Linked Issue:** Fixes #1059

## Changes

- Add an `externalDatabse` block in the values file
- Check that the chart either uses the MariaDB subchart or the externalDatabase but not both
- Check that the chart uses at least one of the MariaDB subchart or the externalDatabase
- Update the deployment template to connect to the correct database

## Summary by CodeRabbit

* **New Features**
  * Option to use an external database instead of the bundled MariaDB.
  * Commented example configuration included to show external DB connection fields.
  * MariaDB can be conditionally enabled or disabled via deployment settings.

* **Bug Fixes / Validation**
  * Deployment-time validation added to prevent conflicting or missing database configurations.
  * Connection parameters now resolve from the selected database source and fail early if required values are absent.

### DIFF
--- a/deploy/helm/grimmory/templates/_database.tpl
+++ b/deploy/helm/grimmory/templates/_database.tpl
@@ -1,0 +1,119 @@
+{{/*
+Validation — database configuration preconditions.
+
+Ensures that the bundled MariaDB subchart and external database
+configuration are not both specified, and that at least one is set.
+*/}}
+{{- define "grimmory.validate-database" -}}
+	{{- if and .Values.mariadb.enabled .Values.externalDatabase.enabled }}
+		{{- fail "mariadb.enabled and externalDatabase.enabled are mutually exclusive: disable the bundled MariaDB subchart or remove the externalDatabase" }}
+	{{- end }}
+	{{- if and (not .Values.mariadb.enabled) (not .Values.externalDatabase.enabled) }}
+		{{- fail "At least one database must be set: enable mariadb or provide an externalDatabase" }}
+	{{- end }}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- if and (not .host) (or (not .existingSecret.enabled) (and .existingSecret.enabled (not .existingSecret.hostnameKey)))}}
+				{{- fail "No host is set for the externalDatabase." }}
+			{{- end }}
+			{{- if and (not .password) (or (not .existingSecret.enabled) (and .existingSecret.enabled (not .existingSecret.passwordKey)))}}
+				{{- fail "No password is set for the externalDatabase." }}
+			{{- end }}
+		{{- end }}
+	{{- end }}
+{{- end -}}
+
+{{/*
+Connection helpers — resolve a single database configuration value.
+
+Usage:  {{ include "grimmory.dbHostname" . }}
+*/}}
+{{- define "grimmory.dbHostname" -}}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- if and .existingSecret.enabled .existingSecret.hostnameKey }}
+				{{- fail "Chart bug, this helper should not be called when an existing secret is set" }}
+			{{- else }}
+				{{- .host }}
+			{{- end }}
+		{{- end }}
+	{{- else }}
+		{{- printf "%s-mariadb" .Release.Name }}
+	{{- end }}
+{{- end -}}
+
+{{- define "grimmory.dbPort" -}}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- if and .existingSecret.enabled .existingSecret.portKey }}
+				{{- fail "Chart bug, this helper should not be called when an existing secret is set" }}
+			{{- else }}
+				{{- .port }}
+			{{- end }}
+		{{- end }}
+	{{- else }}
+		{{- .Values.mariadb.service.port }}
+	{{- end }}
+{{- end -}}
+
+{{- define "grimmory.dbName" -}}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- if and .existingSecret.enabled .existingSecret.databaseKey }}
+				{{- fail "Chart bug, this helper should not be called when an existing secret is set" }}
+			{{- else }}
+				{{- .database }}
+			{{- end }}
+		{{- end }}
+	{{- else }}
+		{{- .Values.mariadb.auth.database }}
+	{{- end }}
+{{- end -}}
+
+{{- define "grimmory.dbUserName" -}}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- if and .existingSecret.enabled .existingSecret.usernameKey }}
+				{{- fail "Chart bug, this helper should not be called when an existing secret is set" }}
+			{{- else }}
+				{{- .username }}
+			{{- end }}
+		{{- end }}
+	{{- else }}
+		{{- .Values.mariadb.auth.username }}
+	{{- end }}
+{{- end -}}
+
+
+{{- define "grimmory.dbPasswordSecretName" -}}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- required "No secretName was set for the externalDatabase existingSecret." .existingSecret.secretName }}
+		{{- end }}
+	{{- else }}
+		{{- with .Values.mariadb.auth }}
+			{{- if .existingSecret }}
+				{{- .existingSecret }}
+			{{- else }}
+				{{- printf "%s-mariadb" $.Release.Name }}
+			{{- end }}
+		{{- end }}
+	{{- end }}
+{{- end -}}
+
+{{- define "grimmory.dbPasswordSecretKey" -}}
+	{{- if .Values.externalDatabase.enabled }}
+		{{- with .Values.externalDatabase }}
+			{{- required "No passwordKey was set for the externalDatabase existingSecret." .existingSecret.passwordKey }}
+		{{- end }}
+	{{- else }}
+		{{- with .Values.mariadb.auth }}
+			{{- if not .existingSecret }}
+				{{- "mariadb-password" }}
+			{{- else }}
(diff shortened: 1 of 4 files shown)


thibaultamartin:
This PR has been successfully tested against a cluster with a pre-existing MariaDB database deployed via the mariadb-operator.

deploy/helm/grimmory/templates/deployment.yaml
```diff
@@ -36,8 +70,8 @@ spec:
             - /bin/sh
             - -c
             - |
-              echo "Waiting for database on port {{ .Values.mariadb.service.port }}..."
-              until nc -z -v -w5 {{ .Release.Name }}-mariadb {{ .Values.mariadb.service.port }}; do
+              echo "Waiting for database at {{ $dbHost }} port {{ $dbPort }}..."
+              until nc -z -v -w5 {{ $dbHost }} {{ $dbPort }}; do
```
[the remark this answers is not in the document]
thibaultamartin: Should we should use `{{ $dbHost | quote }}` for consistency with the rest of the PR?

In this specific instance it's part of a shell incantation and not yaml, so no strong opinion on what makes it more legible

thibaultamartin:
The issue [raised by coderabbit](https://github.com/grimmory-tools/grimmory/pull/1068#pullrequestreview-4216771605) about the sketchy way to wait for the db to be up is relevant and that's something I can fix, but I think it's worth having a distinct PR for.

If someone wants to create an issue for it I can send a PR for it too.

deploy/helm/grimmory/templates/deployment.yaml
```diff
@@ -1,3 +1,9 @@
+{{- if and .Values.mariadb.enabled .Values.externalDatabase.host }}
+{{- fail "mariadb.enabled and externalDatabase are mutually exclusive: disable the bundled MariaDB subchart or remove the externalDatabase" }}
+{{- end }}
+{{- if and (not .Values.mariadb.enabled) (not .Values.externalDatabase.host) }}
+{{- fail "At least one database must be set" }}
```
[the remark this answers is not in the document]
thibaultamartin: This seems to be about the docker compose docs. It would make sense to eventually have a section about the helm chart, but this is beyond the scope of this PR.

thibaultamartin:
After using the workaround in #1187 to dump and import data, I could successfully use this chart to migrate from the built in database to an external database in the same cluster.

I noticed conflicts, I’ll rebase today.

imnotjames:
> After using the workaround in #1187 to dump and import data, I could successfully use this chart to migrate from the built in database to an external database in the same cluster.
> 
> I noticed conflicts, I’ll rebase today.

Yeah conflicts from your other PR.  Once it's rebased I'll review.

thibaultamartin:
~~Rebased. If needed I can also rebase to have a cleaner commit history, since I flip-flopped between using variables and using helpers.~~

I cleaned up the commit history to make it easier to review, and cleaner to merge.

deploy/helm/grimmory/values.yaml
```diff
@@ -15,6 +15,17 @@ mariadb:
     #   userPasswordKey: mariadb-password
   service:
     port: 3306
+
+# Mutually exclusive with the bundled mariadb subchart above.
+# Set mariadb.enabled to false and provide externalDatabase.host to use this.
+externalDatabase:
```
imnotjames, line 21: I suggest the following schema

```
externalDatabase:
  enabled: false

  ## Database host. Optionally include a colon delimited port like "example:3306"
  host: ""

  ## Database user
  user: grimmory

  ## Database password
  password: ""

  ## Database name
  database: grimmory

  ## Use a existing secret
  existingSecret:
    enabled: false
    # secretName: nameofsecret
    usernameKey: db-username
    passwordKey: db-password
    # hostKey: db-hostname-or-ip
    # databaseKey: db-name
```

1. Without `enabled` it can be hard to turn these things off reliably.  You can't as easily remove an entire tree because of how kube works.  I see host is being used as a surrogate but explicit is more reliable.
2. Make existing secret a map with enabled as well.  Ideally with specific keys to override the other fields.

thibaultamartin: To make sure I got it right, if the admin sets `.externalDatabase.existingSecret.enabled:true`, do you want to strictly require them to use the existing secret for all fields? Or only for the username and password?

If both `.externalDatabase.user` and `.externalDatabase.existingSecret.usernameKey` are both set (assuming `.externalDatabase.enabled: true`), should the chart fail or should one take precedence over the other?

imnotjames: Only for fields that have been specified as non-empty.

`enabled` is because of how kubernetes does patches.

```yaml
foo:
  bar: Hello
baz: world
```

If I apply a new values with kubectl apply:

```yaml
baz: mars
```

The `foo` key remains.

```yaml
foo:
  bar: Hello
baz: mars
```

Setting `enabled` to `false` is not impacted by this.

---

If both are set we should either fail or prefer the secret.

thibaultamartin: I’m tempted to make it fail instead of preferring the secret, to make sure admins have made the conscious choice to use the secret and are not misled by reading values. Does it make sense to you too?

imnotjames: Yes

thibaultamartin:
@imnotjames I had requested a review but I've stumbled upon a logic error. Please discard the review, I'll fix the PR tonight :bow:

deploy/helm/grimmory/values.yaml
thibaultamartin: How strongly do you feel about the `host` containing the port, instead of having the `hostname` and `port` as distinct values?

I'm asking because right now I have to do a `lookup` when the host is defined in a secret, so I can split the host into a hostname and a port. This makes the chart impossible to lint.

Ideally I'd pass everything contained in secrets to the deployment as `valuesFrom`. Looking at https://github.com/grimmory-tools/grimmory/blob/d10396233643e7828150cad129622ef00edc718a/backend/src/main/resources/application.yaml#L59, it seems possible to replace the `DATABASE_URL` with `DATABASE_HOST`, `DATABASE_PORT`, `DATABASE_NAME`, `DATABASE_PASSWORD` and let the backend build the connection string itself.

I could then use environment variables in the grimmory container itself, but also pass the `DATABASE_HOST` and `DATABASE_PORT` to the initContainer so it can use them in the `nc` command.

thibaultamartin: Separately, failing when there's both a username and database defined in the values and in the secret adds some convoluted logic because there are default values in the chart's `values.yaml`.

It'd be much simpler to use the values in the existingSecret if they exist, and fallback to the values if they don't.

deploy/helm/values.yaml
```diff
@@ -0,0 +1,20 @@
+# Default values for grimmory.
```
imnotjames: This file is not needed.

deploy/helm/grimmory/templates/_database.tpl
```diff
@@ -0,0 +1,130 @@
...
+			{{- if and .host .existingSecret.enabled .existingSecret.hostKey}}
+				{{- fail "The host must be defined either in the externalDatabase section or in an existingSecret but not both." }}
+			{{- end }}
+			{{- if .host }}
+				{{- .host }}
+			{{- else }}
+				{{- $secret := lookup "v1" "Secret" $.Release.Namespace .existingSecret.secretName }}
+				{{- if not $secret }}
+					{{- fail printf "Existing secret %s not found in namespace %s" .existingSecret.secretName $.Release.Namespace }}
+				{{- end }}
```
imnotjames: I think we can assume that if a secret is specified it will exist or is otherwise not the concern of the helm template.

This causes some issues when trying to use `helm template` when you don't have direct access to the k8s cluster for example.

imnotjames (commented):
Aside fro the extra values file & the live secret lookup we are good to go.

deploy/helm/grimmory/values.yaml
imnotjames: Sure, we can split the host and port.  This may even be a good thing as it's how the mariadb operator generates secrets.

> It'd be much simpler to use the values in the existingSecret if they exist, and fallback to the values if they don't.

Sure, that's fine.  Or you can remove those default values from the helm values file.

thibaultamartin:
Alright, I've updated the `values.yaml` to split the db host and port, and I've made the logic so that if an existing secret is defined, it always takes precedence over the default values.

I've tested it against my k3s cluster by both:

1. Making a fresh deployment of grimmory with the subchart database
2. Upgrading my existing grimmory deployment (that relies on an external database provided by the mariadb operator, provides the hostname, port, username as a value, and the password as an existing secret)

Some notable updates: I'm not building the connection string for the grimmory container, but I rely on dedicated environment variables:

- I don't use `DATABASE_URL`
- I populate `DATABASE_HOST`, `DATABASE_PORT`, `DATABASE_USERNAME`, `DATABASE_PASSWORD` and  `DATABASE_NAME` either from the values or the secrets

imnotjames (approved):
Seems to work well when I'm testing it.  Thanks for all the hard work!

imnotjames (approved):

PR #1089, opened 2026-05-04
refactor(jpa): apply Hibernate-safe equals and hashCode implementations

balazs-szucs:
## Description

Resolves the JPA Buddy warning: 

> Using @EqualsAndHashCode for JPA entities is not recommended. It can cause severe performance and memory consumption issues.

This change establishes a consistent, reliable, and Hibernate-safe equals() and hashCode() implementation across the codebase. By utilizing org.hibernate.Hibernate.getClass() and getter methods, it ensures entities are compared accurately, even when wrapped in lazy-loaded proxies.

## Changes

Entity equality and hash code standardization:

* Updated all entity classes (`AuthorEntity`, `BookEntity`, `BookLoreUserEntity`, `CategoryEntity`, `LibraryEntity`, `MoodEntity`, `ShelfEntity`, `TagEntity`) to use `Hibernate.getClass(this)` and `getId()` in their `equals` and `hashCode` implementations, making them final and proxy-safe.

## Manual Testing Steps

1. the app compiles
2.
3.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved consistency and reliability of entity identity and equality across the app, so comparisons for authors, books, users, categories, libraries, moods, shelves, and tags behave predictably.
  * Reduced risk of duplicate or inconsistent listings, selection errors, and related UI issues in features that rely on entity identity.


backend/src/main/java/org/booklore/model/entity/AuthorEntity.java
```diff
@@ -48,14 +49,15 @@ public class AuthorEntity {
...
+        if (o == null || Hibernate.getClass(this) != Hibernate.getClass(o)) return false;
+        AuthorEntity that = (AuthorEntity) o;
+        return getId() != null && Objects.equals(getId(), that.getId());
     }
 
     @Override
-    public int hashCode() {
-        return getClass().hashCode();
+    public final int hashCode() {
+        return Hibernate.getClass(this).hashCode();
```
imnotjames, line 61: Doesn't this make the hash code the same for all `AuthorEntity` instances?

backend/src/main/java/org/booklore/model/entity/BookEntity.java
```diff
@@ -131,4 +133,17 @@ public BookFileEntity getPrimaryBookFile() {
     public boolean hasFiles() {
         return bookFiles != null && !bookFiles.isEmpty();
     }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || Hibernate.getClass(this) != Hibernate.getClass(o)) return false;
```
imnotjames: In most cases it's fine to compare classes this way as long as the classloader isn't messed with.. but given some of the jank we have with spring & jackson maybe we should be careful?

Should this use the `.equals()`?

backend/src/main/java/org/booklore/model/entity/BookLoreUserEntity.java
```diff
@@ -65,11 +65,7 @@ public class BookLoreUserEntity {
 
     @BatchSize(size = 20)
     @ManyToMany(fetch = FetchType.LAZY)
-    @JoinTable(
-            name = "user_library_mapping",
-            joinColumns = @JoinColumn(name = "user_id"),
-            inverseJoinColumns = @JoinColumn(name = "library_id")
-    )
+    @JoinTable(name = "user_library_mapping", joinColumns = @JoinColumn(name = "user_id"), inverseJoinColumns = @JoinColumn(name = "library_id"))
```
imnotjames, line 68: Is there a difference to the column definition here?

balazs-szucs: No, probably IDE auto-format kicking in.

backend/src/main/java/org/booklore/model/entity/BookEntity.java
balazs-szucs: Well, yes and no. Normally, you never compare a Jackson-deserialized object against a Hibernate entity, simply due to fact that they live in different layers. In a codebase where separation of concern is enforced, this should be not an issue.

The implementation strongly follow the official Hibernate ones: https://docs.hibernate.org/orm/current/userguide/html_single/#mapping-model-pojo-equalshashcode

So in-case there problems, the problem is no callee, but the caller who wants to cross a bundary between service vs data layer that should not be crossed.

backend/src/main/java/org/booklore/model/entity/AuthorEntity.java
balazs-szucs: Yes, by design. 

In practice this is not ideal, but works.

1, New entity added -> hash = AuthorEntity.class.hashCode()

2, Entity saved -> id gets assigned

3, Hash is still AuthorEntity.class.hashCode() -> nothing changed 

And this is kinda hint at the problem that is: when the hashCode first called, there is _no gurantee_ that an id field exist, i believe. In order for that to happen, and id has to be generated on service layer before step 1, and passed to the data layer ALREADY with existing id, which is kinda big problem in practice.

The problem is also that, there is no "good" valid natural key (in my opinion) we could base it on. There is Hibernate-magic under-the-hood that makes this work but a "good" natural key would obviously the best such as ASIN or something. Also this is sole reason why Hibernate.getClass(this) exists btw, a text book example of what that method is solving. This is more or less quite standard way of dealing with sub-optimal entity that was inherited.

backend/src/main/java/org/booklore/model/entity/BookEntity.java
imnotjames: Okay - you don't like my example.  Got it.  It was an example to give an idea of where I was coming from.

Let's go with Hibernate as the example, then, since we cannot avoid hibernate here, right?

Hibernate can have multiple classloaders.  In theory the `Hibernate.getClass()` will return the origin class given a proxy `this` or `o`.  However, if two different classloaders are used we'd get two different classes even if they're the same underlying bytecode.

Regardless, after researching this it doesn't matter so much in this case.

While I'd still suggest `equals` for the sake of "comparing objects should be done with `equals`", given [`Class` is final](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Class.java#L221) & it uses the [`Object` equals](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Object.java#L171-L173) this is moot.

balazs-szucs: > While I'd still suggest equals for the sake of "comparing objects should be done with equals", given [Class is final](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Class.java#L221) & it uses the [Object equals](https://github.com/openjdk/jdk/blob/master/src/java.base/share/classes/java/lang/Object.java#L171-L173) this is moot.


I am not particularly attached to that, if you want i can do this.


> Okay - you don't like my example. Got it. It was an example to give an idea of where I was coming from.

apologies if it came of rude, yes i took your example at face value, didn't clock that you were hinting to the bigger idea

balazs-szucs: [6d67297](https://github.com/grimmory-tools/grimmory/pull/1089/commits/6d67297afec1689b63a628d428f07965725bd19c)

backend/src/main/java/org/booklore/model/entity/AuthorEntity.java
imnotjames: > Also this is sole reason why Hibernate.getClass(this) exists btw, a text book example of what that method is solving. This is more or less quite standard way of dealing with sub-optimal entity that was inherited.

I don't really understand - `getClass()` just returns the underlying class that a proxy object is referencing, so then the `hashCode` of that would remain the same for every object.

So this means that we will run into performance when using sets / hashmaps.  We move from being `o(1)` for lookups to being an `o(n)` lookup.  In the case of a `Set` we'd have to check every item in the set before inserting or removing.

```java
AuthorEntity authorA = authorRepository.findById(1L);
AuthorEntity authorB = authorRepository.findById(2L);

authorA.hashCode() // org.booklore.model.entity.AuthorEntity@193792e6
authorB.hashCode() // org.booklore.model.entity.AuthorEntity@193792e6
```

Does this match up with [the docs you referenced in another comment](https://docs.hibernate.org/orm/current/userguide/html_single/#mapping-model-pojo-equalshashcode)? 

In particular, I understood those docs to list out these options:

* Omission of a custom hashcode and rely on Hibernate providing a stable object within a single session
* > A common initial approach is to use the entity’s identifier attribute as the basis for equals/hashCode calculations
* > Another option is to force the identifier to be generated and set prior to adding to the Set
* > The final approach is to use a "better" equals/hashCode implementation, making use of a natural-id or business-key.

What is a bit confusing here is that we're using the `id` field in the `equals` so I'm not sure why we wouldn't use the ID field here, too?

```
        return getId() != null && Objects.equals(getId(), that.getId());
```

If the `id` is NULL and we still need a has code, I think then it would make sense to fall back to the object's hash code given the documentation's note:

> This is, in fact, exactly the case when dealing with data coming from a database. Every time we load a specific Person from the database we would naturally get a unique instance. Hibernate, however, works hard to make sure that does not happen within a given Session. In fact, Hibernate guarantees equivalence of persistent identity (database row) and Java identity inside a particular session scope. So if we ask a Hibernate Session to load that specific Person multiple times we will actually get back the same instance:

Given this, `if (getId() == null) { return super().hashCode(); }` would actually make sense in many cases, right?

---

Given this isn't technically changing the behavior - the hash code is going to be static before and after this change -  I am okay with us not changing it today, but it's something we should investigate in the future.

backend/src/main/java/org/booklore/model/entity/BookEntity.java
```diff
@@ -131,4 +133,17 @@ public BookFileEntity getPrimaryBookFile() {
...
         return bookFiles != null && !bookFiles.isEmpty();
     }
+
+    @Override
+    public final boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || Hibernate.getClass(this).equals(Hibernate.getClass(o))) return false;
+        BookEntity that = (BookEntity) o;
+        return getId() != null && Objects.equals(getId(), that.getId());
+    }
```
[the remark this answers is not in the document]
balazs-szucs: MB i may have forgotten that will fix

backend/src/main/java/org/booklore/model/entity/AuthorEntity.java
balazs-szucs: Looks like the rabbit also got same thing: https://github.com/grimmory-tools/grimmory/pull/1089#discussion_r3211301695

balazs-szucs: > Another option is to force the identifier to be generated and set prior to adding to the Set

is exactly what we did here.


> return getId() != null && Objects.equals(getId(), that.getId());

i do not believe that would work, i have never seen it anywhere. For specific, i have to look up some docs though

balazs-szucs: The performance concern is irrelevant, any optimization that may gained here, is nothing, compared to using pagination.

See also: https://vladmihalcea.com/how-to-implement-equals-and-hashcode-using-the-jpa-entity-identifier/

Where in the section of: Why using a constant hashCode is not a performance issue for JPA entities

is this exact thing is discussed.

backend/src/main/java/org/booklore/model/entity/BookEntity.java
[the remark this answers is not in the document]
balazs-szucs: done

backend/src/main/java/org/booklore/model/entity/AuthorEntity.java
imnotjames: > i do not believe that would work, i have never seen it anywhere. For specific, i have to look up some docs though

https://github.com/grimmory-tools/grimmory/pull/1089/changes#diff-cb294fcea80c345a74ecad136eded1748bb74e173106e132d7860a9ff6c37be9R56

I think I'm misunderstanding you because this is what you had done.

imnotjames: The article is very much "This is how it should always be" when it's "This is how it works for my use cases".

but sure - big picture, it's limited in usefulness of performance for our use cases, even if it's not in the Hibernate docs.

imnotjames (approved):

PR #1194, opened 2026-05-07
feat(api): infer reading session data points when not specified

### LINKED ISSUE #1190: Infer values in reading session
When values are omitted in the reading session we should infer them where possible.

* `bookType` - infer from the primary file of the book ID
* `durationSeconds` - infer from the difference between start & end time
* `durationFormatted` - infer from the `durationSeconds`
* `progressDelta` - infer from the difference between start and end progress

This would allow for a more consistent experience & support simpler workflows.  In particular, this would simplify what data needs to be calculated on the client side from third party apps.

imnotjames:
## Description

The reading session recording endpoint has a number of redundant fields that can be inferred.  This adds difficulty for integrations to use the endpoints & can lead to incorrect data being published.

Instead, allow API clients to omit these fields, but use the inferred values so we don't lose fidelity of data.

## Linked Issue

Fixes #1190

## Changes

* allow omission of `durationSeconds` in reading session recording request
* update `ReadingSessionService` to fill in progress delta, duration, duration seconds, and book type if not specified
* adds tests for changed behavior in `ReadingSessionService`

## Manual Testing Steps

1. Send request to API with progress delta, duration seconds, duration seconds format, and book type present
2. Check recorded session information has the specified data
3. Send request to API without data
4. Check recorded session information has inferred data

## Screenshots (Optional)

N/A

This came out of investigations into integrating third party reading tools with Grimmory & while it's not required does make the use of the API easier.

## AI Disclosure

None.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced that reading sessions must include a non-null end time.

* **New Features**
  * Automatically compute missing duration, human-readable formatted duration, and progress delta from start/end when not provided.
  * Infer book type from associated book file when not supplied.

* **Tests**
  * Added unit tests validating duration/progress inference, book-type inference, and persistence behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1194)

### DIFF
--- a/backend/src/main/java/org/booklore/service/ReadingSessionService.java
+++ b/backend/src/main/java/org/booklore/service/ReadingSessionService.java
@@ -14,6 +14,7 @@ import org.booklore.model.entity.BookEntity;
 import org.booklore.model.entity.BookLoreUserEntity;
 import org.booklore.model.entity.CategoryEntity;
 import org.booklore.model.entity.ReadingSessionEntity;
+import org.booklore.model.enums.BookFileType;
 import org.booklore.model.enums.ReadStatus;
 import org.booklore.repository.BookRepository;
 import org.booklore.repository.ReadingSessionRepository;
@@ -26,11 +27,7 @@ import org.springframework.security.core.userdetails.UsernameNotFoundException;
 import org.springframework.stereotype.Service;
 import org.springframework.transaction.annotation.Transactional;
 
-import java.time.DayOfWeek;
-import java.time.Instant;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.ZoneId;
+import java.time.*;
 import java.time.format.TextStyle;
 import java.time.temporal.ChronoUnit;
 import java.time.temporal.TemporalAdjusters;
@@ -90,32 +87,69 @@ public class ReadingSessionService {
         return (dow.getValue() % 7) + 1;
     }
 
+    private String getDurationFormatted(int duration) {
+        int hours = duration / 3600;
+        int minutes = (duration % 3600) / 60;
+        int seconds = duration % 60;
+
+        if (hours > 0) {
+            return String.format("%dh %dm %ds", hours, minutes, seconds);
+        } else if (minutes > 0) {
+            return String.format("%dm %ds", minutes, seconds);
+        } else {
+            return String.format("%ds", seconds);
+        }
+    }
+
     @Transactional
     public void recordSession(ReadingSessionRequest request) {
         BookLoreUser authenticatedUser = authenticationService.getAuthenticatedUser();
         Long userId = authenticatedUser.getId();
 
         BookLoreUserEntity userEntity = userRepository.findById(userId).orElseThrow(() -> new UsernameNotFoundException("User not found with ID: " + userId));
-        BookEntity book = bookRepository.findById(request.getBookId()).orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(request.getBookId()));
+        BookEntity book = bookRepository.findByIdWithBookFiles(request.getBookId()).orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(request.getBookId()));
+
+        BookFileType bookType = request.getBookType();
+        if (bookType == null && book.getPrimaryBookFile() != null) {
+            bookType = book.getPrimaryBookFile().getBookType();
+        }
+
+        Integer durationSeconds = request.getDurationSeconds();
+        if (durationSeconds == null) {
+            // We don't need to check if request start / end time are non-null because
+            // that's enforced by the request model.
+            Duration duration = request.getStartTime().until(request.getEndTime());
+            durationSeconds = Math.abs(Math.toIntExact(duration.getSeconds()));
+        }
+
+        String durationFormatted = request.getDurationFormatted();
+        if (durationFormatted == null) {
+            durationFormatted = getDurationFormatted(durationSeconds);
+        }
+
+        Float progressDelta = request.getProgressDelta();
+        if (progressDelta == null && request.getStartProgress() != null && request.getEndProgress() != null) {
+            progressDelta = Math.abs(request.getEndProgress() - request.getStartProgress());
+        }
 
         ReadingSessionEntity session = ReadingSessionEntity.builder()
                 .user(userEntity)
                 .book(book)
-                .bookType(request.getBookType())
+                .bookType(bookType)
                 .startTime(request.getStartTime())
                 .endTime(request.getEndTime())
-                .durationSeconds(request.getDurationSeconds())
-                .durationFormatted(request.getDurationFormatted())
+                .durationSeconds(durationSeconds)
+                .durationFormatted(durationFormatted)
                 .startProgress(request.getStartProgress())
                 .endProgress(request.getEndProgress())
-                .progressDelta(request.getProgressDelta())
+                .progressDelta(progressDelta)
                 .startLocation(request.getStartLocation())
                 .endLocation(request.getEndLocation())
                 .build();
 
         readingSessionRepository.save(session);
 
-        log.info("Reading session persisted successfully: sessionId={}, userId={}, bookId={}, duration={}s", session.getId(), userId, request.getBookId(), request.getDurationSeconds());
+        log.info("Reading session persisted successfully: sessionId={}, userId={}, bookId={}, duration={}s", session.getId(), userId, request.getBookId(), durationSeconds);
     }
 
     public List<ReadingSessionHeatmapResponse> getSessionHeatmapForYear(int year) {
--- a/backend/src/main/java/org/booklore/service/ReadingSessionService.java
+++ b/backend/src/main/java/org/booklore/service/ReadingSessionService.java
@@ -119,7 +119,7 @@ public class ReadingSessionService {
             // We don't need to check if request start / end time are non-null because
             // that's enforced by the request model.
             Duration duration = request.getStartTime().until(request.getEndTime());
-            durationSeconds = Math.abs(Math.toIntExact(duration.getSeconds()));
+            durationSeconds = Math.toIntExact(Math.abs(duration.getSeconds()));
         }
 
         String durationFormatted = request.getDurationFormatted();
--- a/backend/src/main/java/org/booklore/service/ReadingSessionService.java
+++ b/backend/src/main/java/org/booklore/service/ReadingSessionService.java
@@ -103,6 +103,17 @@ public class ReadingSessionService {
         }
     }
 
+    private int getDuration(ReadingSessionRequest request) {
+        if (request.getDurationSeconds() == null) {
+            // We don't need to check if request start / end time are non-null because
+            // that's enforced by the request model.
+            Duration duration = request.getStartTime().until(request.getEndTime());
+            return Math.toIntExact(Math.abs(duration.getSeconds()));
+        }
+
+        return request.getDurationSeconds();
+    }
+
     @Transactional
     public void recordSession(ReadingSessionRequest request) {
         BookLoreUser authenticatedUser = authenticationService.getAuthenticatedUser();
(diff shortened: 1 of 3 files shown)


backend/src/test/java/org/booklore/service/ReadingSessionServiceTest.java
```diff
@@ -0,0 +1,247 @@
...
+
+        service.recordSession(request);
+
+        ArgumentCaptor<ReadingSessionEntity> captor = ArgumentCaptor.forClass(ReadingSessionEntity.class);
+        verify(readingSessionRepository).save(captor.capture());
+
+        ReadingSessionEntity saved = captor.getValue();
+        assertEquals(2.3f, saved.getProgressDelta(), 0.001f);
+    }
+}
```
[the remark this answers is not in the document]
imnotjames: 1. That does not fail, it just gives negative.  Why would it fail?
2. No, it does not, it's an absolute delta.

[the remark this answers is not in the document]
imnotjames: I'm still going to add these to the tests.

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -90,32 +87,70 @@ private static int toSundayFirstDow(DayOfWeek dow) {
...
+
+        String durationFormatted = request.getDurationFormatted();
+        if (durationFormatted == null) {
+            durationFormatted = getDurationFormatted(durationSeconds);
+        }
+
+        Float progressDelta = request.getProgressDelta();
+        if (progressDelta == null && request.getStartProgress() != null && request.getEndProgress() != null) {
+            progressDelta = Math.abs(request.getEndProgress() - request.getStartProgress());
+        }
```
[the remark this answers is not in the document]
imnotjames: Says who?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -90,32 +87,70 @@ private static int toSundayFirstDow(DayOfWeek dow) {
...
+
+        Integer durationSeconds = request.getDurationSeconds();
+        if (durationSeconds == null) {
+            // We don't need to check if request start / end time are non-null because
+            // that's enforced by the request model.
+            durationSeconds = Math.toIntExact(
+                    request.getStartTime().until(request.getEndTime()).toSeconds()
+            );
+        }
```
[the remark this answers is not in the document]
imnotjames: Who says this isn't valid?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -90,32 +87,70 @@ private static int toSundayFirstDow(DayOfWeek dow) {
...
+            bookType = book.getPrimaryBookFile().getBookType();
+        }
+
+        Integer durationSeconds = request.getDurationSeconds();
+        if (durationSeconds == null) {
+            // We don't need to check if request start / end time are non-null because
+            // that's enforced by the request model.
+            durationSeconds = Math.toIntExact(
+                    request.getStartTime().until(request.getEndTime()).toSeconds()
+            );
```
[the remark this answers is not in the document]
imnotjames: That's wrong.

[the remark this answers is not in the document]
imnotjames: I do like `ChronoUnit.SECONDS.between` though

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -90,32 +87,69 @@ private static int toSundayFirstDow(DayOfWeek dow) {
...
+            bookType = book.getPrimaryBookFile().getBookType();
+        }
+
+        Integer durationSeconds = request.getDurationSeconds();
+        if (durationSeconds == null) {
+            // We don't need to check if request start / end time are non-null because
+            // that's enforced by the request model.
+            Duration duration = request.getStartTime().until(request.getEndTime());
+            durationSeconds = Math.abs(Math.toIntExact(duration.getSeconds()));
+        }
```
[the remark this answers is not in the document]
imnotjames: I would prefer the global exception handler in this case.

A: Your error message is confusing
B: That's a 4xx level error when it should be 5xx since it's a problem with our server

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -90,32 +87,69 @@ private static int toSundayFirstDow(DayOfWeek dow) {
...
+        int seconds = duration % 60;
+
+        if (hours > 0) {
+            return String.format("%dh %dm %ds", hours, minutes, seconds);
+        } else if (minutes > 0) {
+            return String.format("%dm %ds", minutes, seconds);
+        } else {
+            return String.format("%ds", seconds);
+        }
+    }
```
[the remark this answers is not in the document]
imnotjames: I am not changing the API more than we have.

zachyale (commented):
I'm fine with inferring duration formatted, but maybe we should hold off on allowing duration to be nullable until we've come to consensus on the final target shape of sessions?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
zachyale, line 104: Frankly, why are we even formatting duration on the B/E? I know this is maybe a V3 concern, so I'm nost saying we should drop it now in V1, but it's so weird

imnotjames: I don't know why - I am only trying to keep the API consistent. :sweat_smile:

 It's definitely weird - it would be better to do on the frontend & could then be done with localization.

imnotjames:
> I'm fine with inferring duration formatted, but maybe we should hold off on allowing duration to be nullable until we've come to consensus on the final target shape of sessions?

I think I understand what you're saying here:

The duration is only nullable for the submission to the API - it's not nullable when being written to the database.  This means that you can omit it when making the "record" request for the API but it never can be null on the response side.

Does that address your concern?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: you can use the duration API for this, and for if/else stuff, you can use simple pattern matching. Then it'd look slightly less awkward

backend/src/main/java/org/booklore/service/ReadingSessionService.java
```diff
@@ -90,32 +87,69 @@ private static int toSundayFirstDow(DayOfWeek dow) {
...
         BookLoreUserEntity userEntity = userRepository.findById(userId).orElseThrow(() -> new UsernameNotFoundException("User not found with ID: " + userId));
-        BookEntity book = bookRepository.findById(request.getBookId()).orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(request.getBookId()));
+        BookEntity book = bookRepository.findByIdWithBookFiles(request.getBookId()).orElseThrow(() -> ApiError.BOOK_NOT_FOUND.createException(request.getBookId()));
+
+        BookFileType bookType = request.getBookType();
+        if (bookType == null && book.getPrimaryBookFile() != null) {
+            bookType = book.getPrimaryBookFile().getBookType();
+        }
+
+        Integer durationSeconds = request.getDurationSeconds();
```
balazs-szucs: i do not believe this should be a object

balazs-szucs: a more idiomatic way to solve this is using optional + int

imnotjames: If I auto-unbox it it will raise a NPE in cases where `durationSeconds` is omitted.

balazs-szucs:
```java
int durationSeconds = request.getDurationSeconds() != null
    ? request.getDurationSeconds()
    : Math.toIntExact(Math.abs(request.getStartTime().until(request.getEndTime()).getSeconds()));
```

or Optional.ofNullable.

backend/src/main/java/org/booklore/service/ReadingSessionService.java
imnotjames: > you can use the duration API for this,

Sure, something like this?

```
var d = Duration.ofSeconds(123456789)
String.format("%dh %dm %ds", d.toHoursPart(), d.toMinutesPart(), d.toSecondsPart());

// "21h 33m 9s"
```

I had initially tried with this but ran into problems with days.  My example is `123456789` seconds which is 1428 days, 21 hours, 33 minutes, and 9 seconds.  The existing format is expected to only be hours, months, and days.

I could update this to try to calculate out the hours from days -but at that point I'm still doing math with numbers and it felt like it defeated the purpose of using the `Duration` helper.

Which direction should I go with that, you think?  (Or am I missing something with the duration API?)

> for if/else stuff, you can use simple pattern matching.

Could you elaborate?  Do you mean something like `.replace("^(0[hm] )+", "")`?

backend/src/main/java/org/booklore/service/ReadingSessionService.java
imnotjames: If i use `Optional` I don't think I can use `int`.

But sure, I'll update the code to avoid storing an `Integer`

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: Pattern matching is:

```java
return switch (0) {
    case 0 when h > 0 -> "%dh %dm %ds".formatted(h, m, s);
    case 0 when m > 0 -> "%dm %ds".formatted(m, s);
    default           -> "%ds".formatted(s);
};
```

it's a borowed construct from functional languages generally preferred over "bunch of if/else"

```java
Duration d = Duration.ofSeconds(durationSeconds);
long h = d.toHours();        // total hours, e.g. 34293 for 1428d 21h
int  m = d.toMinutesPart();  // 0–59, safe
int  s = d.toSecondsPart();  // 0–59, safe
```

for duration

backend/src/main/java/org/booklore/service/ReadingSessionService.java
balazs-szucs: apologies i had some time to spare and it popped out to me, its not really a big deal.

backend/src/main/java/org/booklore/service/ReadingSessionService.java
imnotjames: Got it - thanks!  My java knowledge is stuck in 2010

backend/src/main/java/org/booklore/service/ReadingSessionService.java
imnotjames: Resolved in ab2cccb6b

backend/src/main/java/org/booklore/service/ReadingSessionService.java
imnotjames: Added the switch to `Duration` in 433518b78fd578818ffb6431da9e477083f42f7b

---

I tried the following:

```java
        return switch (0) {
            case 0 when hours > 0 -> String.format("%dh %dm %ds", hours, minutes, seconds);
            case 0 when minutes > 0 -> String.format("%dm %ds", minutes, seconds);
            default -> String.format("%ds", seconds);
        };
    }
```

But it ended up with a compilation error:

```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':compileJava'.
> Compilation failed; see the compiler output below.
  grimmory/backend/src/main/java/org/booklore/service/ReadingSessionService.java:98: error: guards are only allowed for case with a pattern
              case 0 when hours > 0 -> String.format("%dh %dm %ds", hours, minutes, seconds);
                     ^
  grimmory/backend/src/main/java/org/booklore/service/ReadingSessionService.java:99: error: guards are only allowed for case with a pattern
              case 0 when minutes > 0 -> String.format("%dm %ds", minutes, seconds);
                     ^
  2 errors
```

It seems it needs to be...

```java
        return switch (0) {
            case int _ when hours > 0 -> String.format("%dh %dm %ds", hours, minutes, seconds);
            case int _ when minutes > 0 -> String.format("%dm %ds", minutes, seconds);
            default -> String.format("%ds", seconds);
        };
```

Is this more readable than `if`/`else` blocks?

zachyale: a future problem that can be addressed in v3

balazs-szucs: apologies forgot about this PR

I believe so, i think people prefer these functional programming inspired patterns more. It's the new fancy thing around the block, see the so-called modern languages e.g., Go/Rust all have lot of features inspired similarly from that part of programming

imnotjames: Some people may prefer but do you?  It feels like it's a misuse of the feature given it's a `switch` on `0` and still using `if`/`else` with a slightly different language structure.

Regardless, I don't really have the time to work through this further & am closing it out.  It's a low priority change.

PR #1253, opened 2026-05-10
fix(metadata): update EPUB manifest and file extensions when replacing covers

### LINKED ISSUE #1252: Cover replacement writes WebP image but retains JPEG filename/media type
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1077

<sup>Originally posted by **Shaglock** May  3, 2026</sup>
### What went wrong?

When I replace a book cover with a WebP image, Grimmory saves the new WebP file into the EPUB but the OPF manifest still says the cover is a JPEG. This causes EPUBCheck to report corrupted image errors because the file doesn't match the declared media type

#### Reference
Discord discussion: https://discord.com/channels/1483208318720675860/1488232624047522033

### How can we reproduce it?

1. Open an EPUB file that has a JPEG cover in Grimmory
2. Replace the cover with a WebP format image
3. Save the modified EPUB
4. Run EPUBCheck validation on the updated file

**Expected:** The OPF entry is updated to match the new image format:
```xml
<item href="images/cover.webp" id="img-cover" media-type="image/webp" properties="cover-image"/>
```
**Actual:** OPF still declares the old JPEG metadata:
```xml
<item href="images/cover.jpg" id="img-cover" media-type="image/jpeg" properties="cover-image"/>
```
EPUBCheck throws:
```
The file "OEBPS/images/cover.jpg" does not appear to match the media type image/jpeg
Corrupted image file encountered
```

Stable

### Your setup

Grimmory version 2.3.0,  running in Docker via Coolify.

_No response_

### Before submitting

balazs-szucs:
## Description

Updated the EPUB manifest and file extensions to ensure proper handling when replacing covers. This fixes inconsistencies that occurred during cover updates

## Linked Issue

Fixes: #1252

## Manual Testing Steps

1. Replace a cover in an EPUB file.
2. Verify the manifest is updated correctly.
3. Ensure file extensions for cover images are appropriate.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Extended cover image support to JPEG, PNG, WebP and AVIF.
  * Cover replacement can change cover format and filename when needed; EPUBs and internal references are updated accordingly.

* **Improvements**
  * Validation errors now include the detected image MIME for clearer feedback.
  * Cover cleanup and manifest/content updates handle additional filename/extension variants and ensure safe path handling.

* **Tests**
  * Added tests covering WebP and AVIF uploads and EPUB cover rewrite.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1253)

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/writer/AudiobookMetadataWriter.java
+++ b/backend/src/main/java/org/booklore/service/metadata/writer/AudiobookMetadataWriter.java
@@ -181,6 +181,10 @@ public class AudiobookMetadataWriter implements MetadataWriter {
         try {
             String detectedMime = MimeDetector.detect(new ByteArrayInputStream(coverData));
             String extension = getExtensionForMime(detectedMime);
+            if (extension == null) {
+                log.warn("Unsupported audiobook cover MIME type: {}", detectedMime);
+                return;
+            }
             String filename = "cover" + extension;
             Path coverPath = folderPath.resolve(filename);
 
@@ -203,10 +207,11 @@ public class AudiobookMetadataWriter implements MetadataWriter {
 
     private String getExtensionForMime(String mime) {
         return switch (mime) {
+            case "image/jpeg" -> ".jpg";
             case "image/png" -> ".png";
             case "image/webp" -> ".webp";
             case "image/avif" -> ".avif";
-            default -> ".jpg";
+            default -> null;
         };
     }
 
--- a/backend/src/main/java/org/booklore/service/metadata/writer/AudiobookMetadataWriter.java
+++ b/backend/src/main/java/org/booklore/service/metadata/writer/AudiobookMetadataWriter.java
@@ -19,13 +19,20 @@ import org.jaudiotagger.tag.images.ArtworkFactory;
 import org.springframework.stereotype.Component;
 import org.springframework.web.multipart.MultipartFile;
 
-import java.io.*;
+import java.io.ByteArrayInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+
+import org.booklore.util.MimeDetector;
+import org.jaudiotagger.audio.AudioFile;
 
 @Slf4j
 @Component
@@ -129,7 +136,7 @@ public class AudiobookMetadataWriter implements MetadataWriter {
                         tag.deleteArtworkField();
                         Artwork artwork = ArtworkFactory.getNew();
                         artwork.setBinaryData(coverData);
-                        artwork.setMimeType(detectMimeType(coverData));
+                        artwork.setMimeType(MimeDetector.detect(new ByteArrayInputStream(coverData)));
                         tag.setField(artwork);
                         hasChanges[0] = true;
                     } catch (Exception e) {
@@ -172,7 +179,8 @@ public class AudiobookMetadataWriter implements MetadataWriter {
         }
 
         try {
-            String extension = detectImageExtension(coverData);
+            String detectedMime = MimeDetector.detect(new ByteArrayInputStream(coverData));
+            String extension = getExtensionForMime(detectedMime);
             String filename = "cover" + extension;
             Path coverPath = folderPath.resolve(filename);
 
@@ -186,18 +194,20 @@ public class AudiobookMetadataWriter implements MetadataWriter {
     }
 
     private void deleteExistingCovers(Path folderPath) throws IOException {
-        String[] coverNames = {"cover.jpg", "cover.jpeg", "cover.png", "folder.jpg", "folder.jpeg", "folder.png"};
+        String[] coverNames = {"cover.jpg", "cover.jpeg", "cover.png", "cover.webp", "cover.avif", "folder.jpg", "folder.jpeg", "folder.png", "folder.webp", "folder.avif"};
         for (String name : coverNames) {
             Path existing = folderPath.resolve(name);
             Files.deleteIfExists(existing);
         }
     }
 
-    private String detectImageExtension(byte[] data) {
-        if (data.length > 3 && data[0] == (byte) 0x89 && data[1] == (byte) 0x50 && data[2] == (byte) 0x4E && data[3] == (byte) 0x47) {
-            return ".png";
-        }
-        return ".jpg";
+    private String getExtensionForMime(String mime) {
+        return switch (mime) {
+            case "image/png" -> ".png";
+            case "image/webp" -> ".webp";
+            case "image/avif" -> ".avif";
+            default -> ".jpg";
+        };
     }
 
     private void setTagField(Tag tag, FieldKey key, String value, boolean[] hasChanges) {
@@ -280,7 +290,7 @@ public class AudiobookMetadataWriter implements MetadataWriter {
             tag.deleteArtworkField();
             Artwork artwork = ArtworkFactory.getNew();
             artwork.setBinaryData(coverData);
-            artwork.setMimeType(detectMimeType(coverData));
+            artwork.setMimeType(MimeDetector.detect(new ByteArrayInputStream(coverData)));
             tag.setField(artwork);
             f.commit();
 
@@ -335,12 +345,5 @@ public class AudiobookMetadataWriter implements MetadataWriter {
             log.warn("Failed to load image from {}: {}", pathOrUrl, e.getMessage());
             return null;
         }
-    }
-
-    private String detectMimeType(byte[] data) {
-        if (data.length > 3 && data[0] == (byte) 0x89 && data[1] == (byte) 0x50 && data[2] == (byte) 0x4E && data[3] == (byte) 0x47) {
-            return "image/png";
-        }
-        return "image/jpeg";
     }
 }
--- a/backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
+++ b/backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
@@ -2,6 +2,7 @@ package org.booklore.service.metadata.writer;
 
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.booklore.model.MetadataClearFlags;
 import org.booklore.model.dto.settings.MetadataPersistenceSettings;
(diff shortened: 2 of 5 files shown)


backend/src/test/java/org/booklore/service/metadata/BookCoverServiceTest.java
```diff
@@ -500,6 +501,48 @@ void acceptsPngFile() {
...
+                avifHeader[4] = 'f'; avifHeader[5] = 't'; avifHeader[6] = 'y'; avifHeader[7] = 'p';
+                avifHeader[8] = 'a'; avifHeader[9] = 'v'; avifHeader[10] = 'i'; avifHeader[11] = 'f';
+                when(file.getInputStream()).thenReturn(new ByteArrayInputStream(avifHeader));
+                when(file.getBytes()).thenReturn(new byte[]{1, 2, 3});
+            } catch (Exception _) {}
+
+            when(bookQueryService.findAllWithMetadataByIds(any())).thenReturn(List.of());
+
+            service.updateCoverFromFileForBooks(Set.of(1L), file);
+        }
```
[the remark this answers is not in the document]
balazs-szucs: Out of scope for this PR

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,97 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+                    try {
+                        String content = Files.readString(path, StandardCharsets.UTF_8);
+                        if (content.contains(oldFilename)) {
+                            String newContent = content.replace(oldFilename, newFilename);
+                            Files.writeString(path, newContent, StandardCharsets.UTF_8);
+                            log.debug("Updated references in file: {}", path);
+                        }
+                    } catch (IOException e) {
+                        log.warn("Failed to update references in file {}: {}", path, e.getMessage());
+                    }
```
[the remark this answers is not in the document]
balazs-szucs: I feel like in-practice this is _very much_ rare edge and not really something that should be common... but sure.

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+            }
+
+            if (changed) {
+                saveXmlDocument(doc, file);
+                log.debug("Updated XML references in file: {}", file);
+            }
+        } catch (Exception e) {
+            log.warn("Failed to parse XML file {} for reference update, falling back to regex: {}", file, e.getMessage());
+            updateReferencesRegex(file, oldCoverPath, newCoverPath);
+        }
```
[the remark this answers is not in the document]
balazs-szucs: Not in purview of this PR

backend/src/main/java/org/booklore/service/metadata/writer/AudiobookMetadataWriter.java
```diff
@@ -19,14 +19,21 @@
...
+import java.io.InputStream;
 import java.net.URI;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.StandardCopyOption;
 import java.util.logging.Level;
 import java.util.logging.Logger;
 
+import org.booklore.util.MimeDetector;
+import org.jaudiotagger.audio.AudioFile;
```
imnotjames, line 35: This seems like a duplicate import from line 13 - is it?

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -464,8 +468,11 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
             throw new IOException("No cover item found in manifest");
         }
 
+        String detectedMime = MimeDetector.detect(new ByteArrayInputStream(coverData));
```
imnotjames, line 471: Would it make sense to add an overload to `MimeDetctor.detect` which accepts `byte[]`?

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+            throw new IOException("Unsupported cover MIME type for EPUB: " + detectedMime);
+        }
+
+        String newHref = coverHref;
+        Path newCoverFilePath = currentCoverFilePath;
+
+        if (!newExt.equalsIgnoreCase(currentExt)) {
+            String newDecodedHref = decodedCoverHref.substring(0, decodedCoverHref.length() - currentExt.length()) + newExt;
+            // Simple URL encoding for the new href, focusing on keeping it clean and preserving forward slashes
+            newHref = URLEncoder.encode(newDecodedHref, StandardCharsets.UTF_8).replace("+", "%20").replace("%2F", "/");
```
imnotjames, line 517: Do we need to encode the `href` here?

I know we need to decode it when reading.  That seems to be because we are trying to normalize all the oddities we might find in the wild.  I can't think of a case where this is a problem - especially given we're using actual XML serialization.

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+            newCoverFilePath = opfDir.resolve(newDecodedHref).normalize();
+
+            // Security check: ensure new cover path is within the extracted EPUB directory
+            if (!newCoverFilePath.startsWith(tempDir)) {
+                throw new IOException("Updated cover href resolves outside extracted EPUB: " + newHref);
+            }
+
+            log.info("Updating cover href from {} to {}", coverHref, newHref);
+
+            if (Files.exists(currentCoverFilePath)) {
```
imnotjames, line 528: Why only delete if the ext is different?  Should this be moved down to line 540 before the `Files.createDirectories` call?

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+    }
 
-        Files.createDirectories(coverFilePath.getParent());
-        Files.write(coverFilePath, coverData);
+    private String getExtensionForMimeEPUB(String mime) {
+        return switch (mime) {
+            case "image/jpeg" -> ".jpg";
+            case "image/png" -> ".png";
+            case "image/webp" -> ".webp";
+            default -> null;
```
imnotjames, line 752: We allow avif but don't support it here.

```suggestion
            case "image/avif" -> ".avif";
            default -> null;
```

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+
+            for (Path file : files) {
+                updateReferencesInFile(file, oldCoverPath, newCoverPath);
+            }
+        }
+    }
+
+    private void updateReferencesInFile(Path file, Path oldCoverPath, Path newCoverPath) throws IOException {
+        String name = file.getFileName().toString().toLowerCase();
+        if (name.endsWith(".css")) {
```
imnotjames, line 595: Should we use the mime detector to check this?

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+            updateCssReferences(file, oldCoverPath, newCoverPath);
+        } else {
```
imnotjames, line 597: We probably want to validate this is for sure an XML-parsable document via an `else if`

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
```diff
@@ -478,10 +485,272 @@ private void applyCoverImageToEpub(Path tempDir, Document opfDoc, byte[] coverDa
...
+            if (Files.exists(currentCoverFilePath)) {
+                Files.delete(currentCoverFilePath);
+            }
+
+            // Update guide/reference hrefs (and any other in-OPF refs) in memory so the
+            // later transformer.transform of opfDoc preserves them.
+            updateOpfHrefReferences(opfDoc, opfPath, currentCoverFilePath, newCoverFilePath);
+
+            // Update all references in the EPUB to the new cover filename
+            updateInternalReferences(tempDir, currentCoverFilePath, newCoverFilePath);
```
imnotjames, line 537: I feel conflicted with this approach.

On one hand, yeah - from a technical perspective it makes sense that if you replace a file with another file, you should update references.

On the other hand - from an author's perspective, I don't know it makes as much sense.  From the perspective of a curated document that was decided on by a human for various reasons, we cannot change this.  If there are references to the cover in CSS or in various HTML / XML files, changing it to a different file could lead to breakage in flow, meaning, or intent of the author.

I am concerned that we're no longer writing just the metadata - we are also rewriting the data.  This can lead to a LARGE amount of edge cases and failure scenarios which are hard to account for.  If we were writing a copy of the EPUB that's intended for another use case (eg, for sending to an ebook reader like kepub), sure - what's the worst that could happen?  We'd still be able to go back and get the old file and try again after a bug fix.  But this is modifying the original file.

Instead, a less destructive approach could be to scan the files for internal references (which can be done in a simple "does this string exist" way?) and in those cases we can leave the original cover image along-side the new cover which is now referenced by the metadata.

If users find that this is a problem, then we can follow up again?

balazs-szucs: They are finding this problematic, because this is blocking send kindle to feature... E.g., these EPUBs fail EPUBCHECK that is leading to blocking of send to kindle.

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
balazs-szucs: Sure

backend/src/main/java/org/booklore/service/metadata/writer/EpubMetadataWriter.java
imnotjames: Yes, users are finding the metadata being incorrect a problem.  

This is updating every other internal reference throughout the file - meaning we could be modifying any XML, XHTML, HTML, or SVG.

This could lead to corrupting the file in other ways we don't expect if there's any unintended behaviors, right?

balazs-szucs: Well, this was coderabbit suggestion here: https://github.com/grimmory-tools/grimmory/pull/1253#discussion_r3214951911

to be clear, i am not actually that attached to that particular thing. If we focus _solely_ on the .opf file, that would resolve the main problem: that is the send to kindle problems.

balazs-szucs: I guess we can also do a switcheroo? E.g., convert the cover file itself to the proper cover name and cover file extension?

balazs-szucs: So instead of rewriting the references, convert the downloaded cover to be 1:1 replacement to the old cover?

imnotjames: It's not wrong - these things do need to be handled in _some_ way.

And yeah - not saying you are attached to it.  I just like to show up and throw wrenches into plans. 😓 

An example I can think of off the top of my head where this could work incorrectly would be:

```html
<html>
<body>
    <p>
      To add a cover to your epub you can use the
      attribute <code>src=images/cover.jpg</code> to provide an image in your HTML.
</body>
</html>
```

Is it contrived?  Sure.  But it's technically valid HTML we'd be breaking.. :sob:

balazs-szucs: Okey so how about the conversion plan? Yay or nay?

balazs-szucs:
As discussed this will be done via conversion and NOT via rewriting the content's of the file

PR #1256, opened 2026-05-10
fix(metadata): handle non-standard OPF locations

### LINKED ISSUE #1250: Upload fails for epub containing package.opf
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1172

<sup>Originally posted by **altotin** May  6, 2026</sup>
### What went wrong?

an epub that named its `.opf` file `package.opf` fails to upload. grimmory is looking for `content.opf`

```
[booklore-api] [t-handler-11415] o.b.s.m.extractor.EpubMetadataExtractor  : Failed to read metadata from EPUB file upload-4268772709178541794.epub: File not found: OEBPS/content.opf

[booklore-api] [t-handler-11415] o.b.exception.GlobalExceptionHandler     : Unhandled exception: Cannot invoke "org.booklore.model.dto.BookMetadata.getTitle()" because "metadata" is null
```

META-INF/container.xml :
```
<?xml version='1.0' encoding='utf-8'?>
<container xmlns='urn:oasis:names:tc:opendocument:xmlns:container' version='1.0'>
<rootfiles>
<rootfile full-path='OEBPS/package.opf' media-type='application/oebps-package+xml'/>
</rootfiles>
</container>
```

### How can we reproduce it?

upload a similar .epub file

Stable

### Your setup

3.0.3

_No response_

### Before submitting

balazs-szucs:
## Description

Updated the EPUB metadata extraction logic to better handle non-standard OPF (Open Packaging Format) file locations and improves XML parsing security. The main changes include introducing a robust method for locating the OPF file via `container.xml`, updating XML parsing to use secure utilities, and adding a test to verify support for OPF files at non-standard paths.

## Linked Issue

Fixes #1250

## Manual Testing Steps

1. Replace a cover in an EPUB file to a non standard location
2. Verify the metadata was read in

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed metadata extraction for EPUBs with metadata files in non-standard locations.
  * Improved cover image handling by increasing fallback stream buffering for more reliable extraction.
  * Strengthened XML parsing with a secure, stricter parsing approach.

* **Tests**
  * Added tests for EPUBs with alternative metadata file locations and unconventional XML quoting.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1256)

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
+++ b/backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
@@ -590,7 +590,7 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
     private String findOpfPath(EpubContainer container) throws IOException, ParserConfigurationException, SAXException {
         String containerXmlPath = "META-INF/container.xml";
         if (!container.exists(containerXmlPath)) {
-            throw new IOException("META-INF/container.xml not found");
+            return "OEBPS/content.opf";
         }
 
         Document containerDoc = parseXmlFromContainer(container, containerXmlPath);
--- a/backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
+++ b/backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
@@ -18,11 +18,15 @@ import org.apache.commons.io.FilenameUtils;
 import org.apache.commons.lang3.StringUtils;
 import org.booklore.model.dto.BookMetadata;
 import org.booklore.service.metadata.BookLoreMetadata;
+import org.booklore.util.SecureXmlUtils;
 import org.springframework.stereotype.Component;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
+import org.xml.sax.SAXException;
 
+import javax.xml.parsers.ParserConfigurationException;
+import java.io.ByteArrayInputStream;
 import java.io.ByteArrayOutputStream;
 import java.io.File;
 import java.io.IOException;
@@ -115,8 +119,8 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
 
         // Last resort: scan container for cover-like images
         try (EpubContainer container = EpubContainers.open(epubFile.toPath())) {
-            Document opf = container.parseOpf();
-            String opfName = container.getOpfName();
+            String opfName = findOpfPath(container);
+            Document opf = parseXmlFromContainer(container, opfName);
 
             // Try OPF manifest for cover-image property
             NodeList items = opf.getElementsByTagName("item");
@@ -127,7 +131,7 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
                     String href = URLDecoder.decode(item.getAttribute("href"), StandardCharsets.UTF_8);
                     String fullPath = resolvePath(opfName, href);
                     if (container.exists(fullPath)) {
-                        ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                        ByteArrayOutputStream baos = new ByteArrayOutputStream(4096);
                         container.streamTo(fullPath, baos);
                         return baos.toByteArray();
                     }
@@ -146,7 +150,7 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
                         String decodedHref = URLDecoder.decode(href, StandardCharsets.UTF_8);
                         String fullPath = resolvePath(opfName, decodedHref);
                         if (container.exists(fullPath)) {
-                            ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                            ByteArrayOutputStream baos = new ByteArrayOutputStream(4096);
                             container.streamTo(fullPath, baos);
                             return baos.toByteArray();
                         }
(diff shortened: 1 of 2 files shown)


backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -582,6 +587,38 @@ private byte[] getImageFromEpubResource(Resource res) {
         }
     }
 
+    private String findOpfPath(EpubContainer container) throws IOException, ParserConfigurationException, SAXException {
+        String containerXmlPath = "META-INF/container.xml";
+        if (!container.exists(containerXmlPath)) {
+            throw new IOException("META-INF/container.xml not found");
+        }
```
imnotjames, line 596: epub4j implements this as a default of `OEBPS/content.opf` - should we do the same here?

backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -582,6 +587,38 @@ private byte[] getImageFromEpubResource(Resource res) {
...
+            throw new IOException("No <rootfile> found in container.xml");
+        }
+
+        // EPUB spec §3.5.1: first rootfile is the default rendition
+        String opfPath = ((Element) rootfiles.item(0)).getAttribute("full-path");
+        if (StringUtils.isBlank(opfPath)) {
+            throw new IOException("Empty full-path in container.xml");
+        }
+
+        return URLDecoder.decode(opfPath, StandardCharsets.UTF_8);
```
imnotjames, line 610: Aside from doing actual XML parsing, what's the difference between this and epub4j?

https://github.com/grimmory-tools/epub4j/blob/main/epub4j-core/src/main/java/org/grimmory/epub4j/archive/ZipEpubContainer.java#L155-L163

Based on #1250 the example file is:

```
<?xml version='1.0' encoding='utf-8'?>
<container xmlns='urn:oasis:names:tc:opendocument:xmlns:container' version='1.0'>
<rootfiles>
<rootfile full-path='OEBPS/package.opf' media-type='application/oebps-package+xml'/>
</rootfiles>
</container>
```

So the issue seems to be that we are trying to parse XML without an XML parser, and while the XML is valid it's using a single-quote for attribute definition - but our code doesn't handle it.

```java
    int fullPathIndex = content.indexOf("full-path=\"");
```

If epub4j was updated to support single-quotes this change would not be needed.

Are we wanting to move the logic out of epub4j?  Or should we update epub4j to fix this?

imnotjames (commented):
As far as I can tell, this is replicating the logic that's already in epub4j - so a few questions.

backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
balazs-szucs: i have _still_ gotten to actually migrating everything, since this is problem now, i though quick fix is better.

Last time the migrating got interrupted due to the SIGSEV stuff

backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
balazs-szucs: It's the mismatch between standard _vs_ reality, META-INF/container.xml is technically required for a valid EPUB. container.xml is used to tell "us" where the metadata .opf file is. The most "well known" configuration is, yes: OEBPS/content.opf, but there is some variance in the real world, hence this PR. But this should yes probably backported to the library once

imnotjames: Sure, a backport would make sense but what I mean is we match this code to "reality", too - similar [to what epub4j does today](https://github.com/grimmory-tools/epub4j/blob/main/epub4j-core/src/main/java/org/grimmory/epub4j/archive/ZipEpubContainer.java#L150-L153):

```suggestion
        if (!container.exists(containerXmlPath)) {
            return "EOBPS/content.opf";
        }
```

backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
imnotjames: Okay - if this is a quick fix that's not meant to be long term, a comment [on line 122](https://github.com/grimmory-tools/grimmory/pull/1256/changes#diff-4cb1a0799b11f3128aedb0fdc07d26124e67d87ba2d70be43eb9da968eb08952R122-R123) describing the issue and what to do once it's fixed in the upstream library would make sense.

imnotjames: We may also want to create an issue on the epub4j library so we know to fix this bug there, too.

balazs-szucs: Addressed here: [990949b](https://github.com/grimmory-tools/grimmory/pull/1256/commits/990949bc425def685b7c75864e1ed6245eefb338)

backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
balazs-szucs: Addressed here: [1502f13](https://github.com/grimmory-tools/grimmory/pull/1256/commits/1502f1315304c92672c4519c524dd6d0ae42fc26)

imnotjames (approved):

PR #1258, opened 2026-05-11
perf(api): initialize StringBuilder with specified capacities for performance

### LINKED ISSUE #1257: Specify initial capacity for StringBuilder instances to improve performance
StringBuilder instances throughout the codebase are constructed without an initial capacity, which causes the JVM to default to a capacity of 16 characters. When that limit is exceeded, the internal char array is resized (doubling in size each time), leading to unnecessary heap allocations and memory copies, particularly costly in frequently called methods or when building larger strings.

Expected Outcome

Fewer heap allocations, reduced GC pressure, and improved throughput in string-heavy code paths.

balazs-szucs:
## Description

This pull request optimizes memory usage and performance by pre-allocating StringBuilder instances with an initial capacity throughout various parts of the codebase. By specifying an appropriate initial size for each StringBuilder, the code reduces the need for dynamic resizing during string concatenation, which can improve efficiency, especially in frequently called methods or when dealing with large strings.

As noted by Intellij:

Reports attempts to instantiate a new StringBuffer or StringBuilder object without specifying its initial capacity.
If no initial capacity is specified, a default capacity is used, which will rarely be optimal. Failing to specify the initial capacity for StringBuffer may result in performance issues if space needs to be reallocated and memory copied when the initial capacity is exceeded.

## Linked Issue
Fixes #1257

## Changes

- Updated `CoverImageGenerator` to reuse a single `StringBuilder` instance by resetting its length instead of reallocating objects.
- Minor improvement tp performance by initializing `StringBuilder` instances with appropriate capacities.

## Manual Testing Steps

1. Verify that the `CoverImageGenerator` class generates images correctly without errors.
2. Check the MD5 hash generation test go through without failure

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Optimized internal memory allocation and string operations across backend services (pre-sized builders, reduced temporary allocations), including metadata, parsing, filename/date pattern handling, audio/chapter processing, and PDF/CBX writing.
  * Reduced allocations in cover image text layout.
  * Switched MD5 formatting to a more efficient hex conversion.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1258)


imnotjames (commented):
Given this is meant to improve memory usage and performance do we have measurements before and after of the improvement?

balazs-szucs:
> Given this is meant to improve memory usage and performance do we have measurements before and after of the improvement?

It's a micro-optimization. Also i will not spin up a benchmarking right for this. I would be surprised if this yields results that may be measured using "real-life" tests. Perhaps, we can make JMH tests that's about it.

Issue #300, opened 2026-03-31
Configurable OIDC Mobile Redirect URI Whitelist


zachyale:
### What's Your Idea?

A change prior to the fork hardcoded the OIDC mobile redirect URI validation to `booklore://` in [commit `3ff8f0f`](https://github.com/grimmory-tools/grimmory/commit/3ff8f0f69f232253d23dc18d2421977aa25bb75b#diff-6d27a07cad2f86a97e5503af7850aac946dbfbfac26160217a75fc9095141997), which effectively binds mobile auth to the a first party app scheme.

Prior to this, the existing behaviour was to accept whatever redirect URI was provided. We should instead implement functionality to allow admins to edit the list of whitelisted mobile redirect URIs for their instance.

Proposed change:
- Add a new app setting for OIDC mobile redirect URI whitelisting
- Use this whitelist when validating:
  - `redirect_uri` in OIDC callback/token exchange flows
  - `app_redirect_uri` in redirect flows
- Keep secure defaults for new installs (e.g. `grimmory://` scheme by default), but allow admins to add/edit client schemes
- Backfill existing installs to avoid regressions (e.g., any setup w/ OIDC currently enabled will backfill to include support for the legacy redirect URI + the new official scheme)
- Optionally support a permissive mode (“allow any mobile redirect URI”) for advanced setups

### Why Would This Be Helpful?

- Improves compatibility with:
  - community clients
  - forks/white-label apps
- Removes vendor lockin to a single official app URI scheme, allows user choice in client selection
- Keeps security posture strong by default via explicit whitelisting

### Anything Else? (Optional)

Implementation idea:
- Introduce an `OIDC_MOBILE_REDIRECT_URIS` setting in `app_settings`
- Validate mobile redirect URIs against this setting in `OidcAuthService`
- Backfill existing OIDC-enabled installs with both legacy + default scheme to avoid regressions
- Resetting to default settings should only restore the default scheme for new installs

### Want to Help Out?

Yes! I'd love to help implement this

### Have You Considered Any Alternatives? (Optional)

- Keep hardcoded scheme and require downstream patching: rejected (poor interoperability)
- Allow any URI always: more flexible, weaker security default
- Best compromise: explicit admin-managed whitelist, with optional open schema setting

### Before Submitting

- [x] I've searched existing issues and confirmed this feature hasn't been requested yet

PR #1268, opened 2026-05-12
chore(oidc): backfill + add mobile redirect URI whitelist setting to the backend

zachyale:
## Description

This PR updates the backend OIDC mobile redirect URI logic to support an admin configurable list of mobile redirect URIs, instead of just using a hardcoded legacy value. It also defaults/backfills the initial redirect URI `grimmory://oauth2-callback`, with silent backwards compatibility for the legacy redirect URI for now (so long as the default URI is present and the legacy one is not also added manually)

## Linked Issue

Part of https://github.com/grimmory-tools/grimmory/issues/300

## Changes

- Added a new `OIDC_MOBILE_REDIRECT_URIS` app setting as a JSON string array
- Added backend validations for various edge cases, including:
  - blank values
  - duplicates
  - requiring wildcards to be the only element, if present
  - reject http/https URIs
  - reject URIs w/ fragments
  - require URIs to have valid schemes
- replaced the hardcoded legacy redirect URI w/ proper handling/checking against the new array
- add support for existing apps using the legacy URI by allowing the legacy URI if the default URI is still present (legacy compatibility is marked w/ a standard comment format that already exists in repo for a breaking change update in the future)
- adds b/e coverage for valid/illegal cases

## Manual Testing Steps

1. open app settings within a table viewer, confirm the backfill value was set
2. test the API surface w/ sample green/red tests to ensure validations work as tested
3. test w/ a sample callback for both the new default, old legacy, old legacy when new default is not present, etc.

## Screenshots (Optional)

No UI changes in this PR.

## AI Disclosure

Codex for spec generation

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Configurable mobile app redirect URIs with a default value and implicit legacy fallback.
  * Settings accept a wildcard to allow flexible mobile redirect configuration (only when sole entry).

* **Bug Fixes**
  * Stricter mobile redirect validation and normalization: rejects http/https, fragments, blanks, duplicates, mixed wildcard lists; app redirect checks now require allowed mobile URIs for app flows.

* **Tests**
  * Added tests covering redirect validation and settings update/validation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1268)

### DIFF
--- a/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
+++ b/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
@@ -42,6 +42,7 @@ public class OidcAuthService {
     private static final String DEFAULT_MOBILE_REDIRECT_URI = "grimmory://oauth2-callback";
     // TODO(grimmory-cleanup): Remove after the Booklore to Grimmory migration window closes.
     private static final String LEGACY_MOBILE_REDIRECT_URI = "booklore://oauth2-callback";
+    private static final String WILDCARD_REDIRECT_URI = "*";
     private static final String OAUTH2_CALLBACK_PATH = "/oauth2-callback";
     private static final Pattern TRAILING_SLASH_PATTERN = Pattern.compile("/+$");
 
@@ -179,7 +180,7 @@ public class OidcAuthService {
 
     private boolean isAllowedMobileRedirectUri(String redirectUri) {
         List<String> effectiveRedirectUris = getEffectiveMobileRedirectUris();
-        return effectiveRedirectUris.contains("*") || effectiveRedirectUris.contains(redirectUri);
+        return effectiveRedirectUris.contains(WILDCARD_REDIRECT_URI) || effectiveRedirectUris.contains(redirectUri);
     }
 
     private List<String> getEffectiveMobileRedirectUris() {
@@ -211,7 +212,7 @@ public class OidcAuthService {
                     && !"http".equalsIgnoreCase(scheme)
                     && !"https".equalsIgnoreCase(scheme)
                     && uri.getFragment() == null;
-        } catch (IllegalArgumentException e) {
+        } catch (IllegalArgumentException _) {
             return false;
         }
     }
--- a/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
+++ b/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
@@ -185,11 +185,9 @@ public class OidcAuthService {
 
     private List<String> getEffectiveMobileRedirectUris() {
         List<String> configuredRedirectUris = appSettingService.getAppSettings().getOidcMobileRedirectUris();
-        List<String> effectiveRedirectUris = new ArrayList<>(
-                configuredRedirectUris == null || configuredRedirectUris.isEmpty()
-                        ? List.of(DEFAULT_MOBILE_REDIRECT_URI)
-                        : configuredRedirectUris
-        );
+        List<String> effectiveRedirectUris = configuredRedirectUris == null || configuredRedirectUris.isEmpty()
+                ? new ArrayList<>(List.of(DEFAULT_MOBILE_REDIRECT_URI))
+                : new ArrayList<>(configuredRedirectUris);
 
         if (effectiveRedirectUris.contains(DEFAULT_MOBILE_REDIRECT_URI)
                 && !effectiveRedirectUris.contains(LEGACY_MOBILE_REDIRECT_URI)) {
--- a/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
+++ b/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
@@ -173,7 +173,7 @@ public class OidcAuthService {
     }
 
     public void validateAppRedirectUri(String appRedirectUri) {
-        if (!isMobileRedirectUri(appRedirectUri)) {
+        if (!isMobileRedirectUri(appRedirectUri) || !isAllowedMobileRedirectUri(appRedirectUri)) {
             throw ApiError.OIDC_INVALID_REDIRECT_URI.createException();
         }
     }
--- a/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
+++ b/backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
@@ -26,6 +26,8 @@ import jakarta.servlet.http.HttpServletRequest;
 
 import java.net.URI;
 import java.text.ParseException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentMap;
@@ -37,7 +39,9 @@ import java.util.regex.Pattern;
 @AllArgsConstructor
 public class OidcAuthService {
 
-    private static final String MOBILE_SCHEME = "booklore://";
+    private static final String DEFAULT_MOBILE_REDIRECT_URI = "grimmory://oauth2-callback";
+    // TODO(grimmory-cleanup): Remove after the Booklore to Grimmory migration window closes.
+    private static final String LEGACY_MOBILE_REDIRECT_URI = "booklore://oauth2-callback";
     private static final String OAUTH2_CALLBACK_PATH = "/oauth2-callback";
     private static final Pattern TRAILING_SLASH_PATTERN = Pattern.compile("/+$");
 
@@ -123,8 +127,8 @@ public class OidcAuthService {
             throw ApiError.OIDC_INVALID_REDIRECT_URI.createException();
         }
 
-        if (redirectUri.startsWith(MOBILE_SCHEME)) {
-            if (!redirectUri.equals(MOBILE_SCHEME + "oauth2-callback")) {
+        if (isMobileRedirectUri(redirectUri)) {
+            if (!isAllowedMobileRedirectUri(redirectUri)) {
                 throw ApiError.OIDC_INVALID_REDIRECT_URI.createException();
             }
             return;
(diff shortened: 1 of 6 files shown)


zachyale:
fwiw, I think stuff like this starts to better highlight the kind of infra we should have, and could have if we had b/e localization. it would be really nice to instead of having to also do f/e validation to have proper localization on errors, if users had locale on the b/e, and if our put/post response structs could be structured to return a json struct w/ none, one or many key value mappings for inputs to errors

backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
```diff
@@ -168,11 +172,50 @@ private String resolveRequestOrigin(HttpServletRequest request) {
...
     public void validateAppRedirectUri(String appRedirectUri) {
-        if (appRedirectUri == null || !appRedirectUri.startsWith(MOBILE_SCHEME)) {
+        if (!isMobileRedirectUri(appRedirectUri)) {
             throw ApiError.OIDC_INVALID_REDIRECT_URI.createException();
         }
     }
 
+    private boolean isAllowedMobileRedirectUri(String redirectUri) {
+        List<String> effectiveRedirectUris = getEffectiveMobileRedirectUris();
+        return effectiveRedirectUris.contains("*") || effectiveRedirectUris.contains(redirectUri);
```
balazs-szucs:
```suggestion
        return effectiveRedirectUris.contains(WILDCARD_REDIRECT_URI) || effectiveRedirectUris.contains(redirectUri);
```

at the top:

private static final String WILDCARD_REDIRECT_URI = "*";

backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
```diff
@@ -168,11 +172,50 @@ private String resolveRequestOrigin(HttpServletRequest request) {
...
+
+        try {
+            URI uri = URI.create(redirectUri);
+            String scheme = uri.getScheme();
+            return scheme != null
+                    && !scheme.isBlank()
+                    && !"http".equalsIgnoreCase(scheme)
+                    && !"https".equalsIgnoreCase(scheme)
+                    && uri.getFragment() == null;
+        } catch (IllegalArgumentException e) {
```
balazs-szucs:
```suggestion
        } catch (IllegalArgumentException _) {
```

balazs-szucs: Shouldn't we cache this somewhere?

backend/src/main/java/org/booklore/service/appsettings/AppSettingService.java
```diff
@@ -107,6 +115,56 @@ private void validatePermission(AppSettingKey key, BookLoreUser user) {
...
+            if (scheme == null || scheme.isBlank()) {
+                throw ApiError.GENERIC_BAD_REQUEST.createException("Redirect URI must include a scheme");
+            }
+            if ("http".equalsIgnoreCase(scheme) || "https".equalsIgnoreCase(scheme)) {
+                throw ApiError.GENERIC_BAD_REQUEST.createException("Redirect URI must use a custom mobile scheme");
+            }
+            if (uri.getFragment() != null) {
+                throw ApiError.GENERIC_BAD_REQUEST.createException("Redirect URI must not contain a fragment");
+            }
+        } catch (IllegalArgumentException e) {
```
balazs-szucs:
```suggestion
        } catch (IllegalArgumentException _) {
```

backend/src/main/java/org/booklore/service/appsettings/AppSettingService.java
```diff
@@ -107,6 +115,56 @@ private void validatePermission(AppSettingKey key, BookLoreUser user) {
...
+        }
+        if (!(val instanceof List<?> rawValues)) {
+            throw ApiError.GENERIC_BAD_REQUEST.createException("OIDC mobile redirect URIs must be an array");
+        }
+
+        List<String> redirectUris = rawValues.stream()
+                .map(value -> value == null ? null : String.valueOf(value).trim())
+                .toList();
+
+        if (redirectUris.contains("*") && redirectUris.size() > 1) {
```
balazs-szucs:
```suggestion
        if (redirectUris.contains(WILDCARD_REDIRECT_URI) && redirectUris.size() > 1) {
```

at the top:

private static final String WILDCARD_REDIRECT_URI = "*";

backend/src/main/java/org/booklore/service/appsettings/AppSettingService.java
```diff
@@ -107,6 +115,56 @@ private void validatePermission(AppSettingKey key, BookLoreUser user) {
...
+            }
+            if (!"*".equals(redirectUri)) {
+                validateMobileRedirectUriShape(redirectUri);
+            }
+            if (!uniqueUris.add(redirectUri)) {
+                throw ApiError.GENERIC_BAD_REQUEST.createException("Duplicate redirect URI: " + redirectUri);
+            }
+        }
+
+        return redirectUris;
```
balazs-szucs: This smells, i would go something like this:


```suggestion
        Set<String> uniqueUris = new LinkedHashSet<>(redirectUris.size());
        for (String redirectUri : redirectUris) {
            if (redirectUri == null || redirectUri.isBlank()) {
                throw ApiError.GENERIC_BAD_REQUEST.createException("Redirect URI cannot be blank");
            }
            if (!WILDCARD_REDIRECT_URI.equals(redirectUri)) {
                validateMobileRedirectUriShape(redirectUri);
            }
            if (!uniqueUris.add(redirectUri)) {
                throw ApiError.GENERIC_BAD_REQUEST.createException("Duplicate redirect URI: " + redirectUri);
            }
        }

        return List.copyOf(uniqueUris);
    }
```

Set<String> uniqueUris = new LinkedHashSet<>(redirectUris.size()); is kinda still strange

backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
```diff
@@ -168,11 +172,50 @@ private String resolveRequestOrigin(HttpServletRequest request) {
...
+        return effectiveRedirectUris.contains("*") || effectiveRedirectUris.contains(redirectUri);
+    }
+
+    private List<String> getEffectiveMobileRedirectUris() {
+        List<String> configuredRedirectUris = appSettingService.getAppSettings().getOidcMobileRedirectUris();
+        List<String> effectiveRedirectUris = new ArrayList<>(
+                configuredRedirectUris == null || configuredRedirectUris.isEmpty()
+                        ? List.of(DEFAULT_MOBILE_REDIRECT_URI)
+                        : configuredRedirectUris
+        );
```
balazs-szucs:
```suggestion
        List<String> configuredRedirectUris = appSettingService.getAppSettings().getOidcMobileRedirectUris();
        List<String> effectiveRedirectUris = configuredRedirectUris == null || configuredRedirectUris.isEmpty()
                ? new ArrayList<>(List.of(DEFAULT_MOBILE_REDIRECT_URI))
                : new ArrayList<>(configuredRedirectUris);
```

Something like this would be more idiomatic. It avoids the build‑to‑check anti‑pattern.

balazs-szucs: or even better but not sure if:

```java
List<String> effective = Optional.ofNullable(
        appSettingService.getAppSettings().getOidcMobileRedirectUris())
    .filter(l -> !l.isEmpty())
    .map(ArrayList::new)
    .orElseGet(() -> new ArrayList<>(List.of(DEFAULT_MOBILE_REDIRECT_URI)));
```

backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
```diff
@@ -37,7 +39,10 @@
 @AllArgsConstructor
 public class OidcAuthService {
 
-    private static final String MOBILE_SCHEME = "booklore://";
+    private static final String DEFAULT_MOBILE_REDIRECT_URI = "grimmory://oauth2-callback";
+    // TODO(grimmory-cleanup): Remove after the Booklore to Grimmory migration window closes.
+    private static final String LEGACY_MOBILE_REDIRECT_URI = "booklore://oauth2-callback";
+    private static final String WILDCARD_REDIRECT_URI = "*";
     private static final String OAUTH2_CALLBACK_PATH = "/oauth2-callback";
```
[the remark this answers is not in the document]
zachyale: This is fine since this is hosted locally on user homelabs, they would need to already whitelist the given redirect URI for their SSO

backend/src/main/java/org/booklore/config/security/oidc/OidcAuthService.java
zachyale: Cache the result of `isMobileRedirectUri()`? I don't see what that would buy us- the method is a small validation check on a given URI string- there's no IO, settings fetched, etc- can you explain further?

backend/src/main/java/org/booklore/service/appsettings/AppSettingService.java
zachyale: I've updated it to return a list copy of the unique URIs- don't think we need the size init

imnotjames (commented):
Same comment as other side of the PR - is there a reason we're restricting this to just mobile redirect URIs?

balazs-szucs (approved):

balazs-szucs:
> Same comment as other side of the PR - is there a reason we're restricting this to just mobile redirect URIs?

I believe, even if there is no reason, perhaps we can make an issue to that, and deal with it later. This is about mobile stuff, so i guess let it be for mobile stuff?

backend/src/main/java/org/booklore/model/dto/settings/AppSettings.java
```diff
@@ -27,6 +27,7 @@ public class AppSettings {
     private boolean metadataDownloadOnBookdrop;
     private boolean oidcEnabled;
     private OidcProviderDetails oidcProviderDetails;
+    private List<String> oidcMobileRedirectUris;
```
imnotjames:
```suggestion
    private List<String> oidcRedirectUris;
```

If we find we can use `http`/`https` in the future (which would be good) this will make things less confusing.

The rest is fine.

backend/src/main/java/org/booklore/model/dto/settings/AppSettingKey.java
```diff
@@ -10,6 +10,7 @@ public enum AppSettingKey {
     // @formatter:off
     // ADMIN only (public settings)
     OIDC_PROVIDER_DETAILS               ("oidc_provider_details",                true,  true,  List.of(PermissionType.ADMIN)),
+    OIDC_MOBILE_REDIRECT_URIS           ("oidc_mobile_redirect_uris",            true,  false, List.of(PermissionType.ADMIN)),
```
imnotjames: Same thing here.


```suggestion
    OIDC_REDIRECT_URIS           ("oidc_redirect_uris",            true,  false, List.of(PermissionType.ADMIN)),
```

PR #1287, opened 2026-05-13
feat(oidc): add configurable mobile redirect URI to the admin settings UI

### LINKED ISSUE #300: Configurable OIDC Mobile Redirect URI Whitelist

zachyale:
## Description

This PR adds frontend support for configuring OIDC mobile redirect URIs on the OIDC settings page. It introduces a chip-based redirect URI input, aligns frontend validation with the backend, and updates the UI to use Prime chips so the redirect URI pills behave consistently with the current theme

## Linked Issue

Closes #300

## Changes

- Added `OIDC_MOBILE_REDIRECT_URIS` to the frontend app settings model and settings wiring.
- Added a mobile redirect URI input/copy to Settings
- Added frontend validation and validation copy for:
  - blank values already present in the saved array
  - duplicate values
  - `*` mixed with other values
  - missing scheme
  - `http` and `https` URIs
  - fragment-bearing URIs
  - otherwise malformed URIs
- Added frontend specs

## Manual Testing Steps

1. Open Settings -> OIDC, and verify the mobile redirect URI input appears with the expected helper copy
2. Add and remove redirect URIs, including valid custom-scheme values and invalid cases such as duplicates, `https://...`, missing-scheme values, fragments, and mixed `*` arrays, and verify the correct inline validation message is shown
3. Save valid redirect URI values, reload the page, and verify the saved values are restored and rendered correctly

## Screenshots (Optional)

Screenshot 2026-05-12 at 1 08 25 AM

Sample inline validation:
Screenshot 2026-05-13 at 2 02 02 AM

I err towards f/e inline validations- I think it's a better UX, but I understand the risk of duplicating validation logic across FE/BE. Happy to kill the FE validations and rely on the BE validations in the previous PR, but if so, I think we should get better alignment as a team on our philosophy towards surfacing validation errors served from the BE in forms- I think there's a lot of value in doing so

## AI Disclosure

Codex - helped fix the prime ng chip styling + helped gen boilerplate specs

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added mobile redirect URI management to OAuth2/OIDC authentication settings. Users can now add, view, and remove redirect URI entries through an intuitive input interface.

* **Tests**
  * Added comprehensive test coverage for mobile redirect URI functionality and validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1287)

### DIFF
--- a/frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
+++ b/frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
@@ -217,7 +217,7 @@
         </p>
         <div class="config-item full-width">
           <label for="mobileRedirectUriInput">{{ t('mobileRedirectUris.label') }}</label>
-          <div class="redirect-uri-input" [class.has-error]="mobileRedirectUriErrorKey">
+          <div class="redirect-uri-input">
             <div class="redirect-uri-chip-list">
               @for (redirectUri of mobileRedirectUris; track redirectUri; let index = $index) {
                 <p-chip
@@ -245,9 +245,6 @@
             {{ '. ' + t('mobileRedirectUris.hintSuffix') }}
             <code>{{ wildcardMobileRedirectUri }}</code>
           </span>
-          @if (mobileRedirectUriErrorKey) {
-            <span class="field-error">{{ t(mobileRedirectUriErrorKey) }}</span>
-          }
         </div>
       </div>
 
--- a/frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
+++ b/frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
@@ -243,7 +243,7 @@
             {{ t('mobileRedirectUris.hintPrefix') }}
             <code>{{ defaultMobileRedirectUri }}</code>
             {{ '. ' + t('mobileRedirectUris.hintSuffix') }}
-            <code>*</code>
+            <code>{{ wildcardMobileRedirectUri }}</code>
           </span>
           @if (mobileRedirectUriErrorKey) {
             <span class="field-error">{{ t(mobileRedirectUriErrorKey) }}</span>
--- a/frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
+++ b/frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
@@ -210,6 +210,47 @@
         </div>
       </div>
 
+      <div class="claims-section">
+        <h4 class="subsection-title">{{ t('mobileRedirectUris.title') }}</h4>
+        <p class="subsection-description">
+          {{ t('mobileRedirectUris.description') }}
+        </p>
+        <div class="config-item full-width">
+          <label for="mobileRedirectUriInput">{{ t('mobileRedirectUris.label') }}</label>
+          <div class="redirect-uri-input" [class.has-error]="mobileRedirectUriErrorKey">
+            <div class="redirect-uri-chip-list">
+              @for (redirectUri of mobileRedirectUris; track redirectUri; let index = $index) {
+                <p-chip
+                  class="redirect-uri-chip"
+                  [label]="redirectUri"
+                  [removable]="true"
+                  [attr.aria-label]="t('mobileRedirectUris.removeLabel', {uri: redirectUri})"
+                  (onRemove)="removeMobileRedirectUri(index)">
+                </p-chip>
+              }
+              <input
+                pInputText
+                pSize="small"
+                id="mobileRedirectUriInput"
+                [(ngModel)]="mobileRedirectUriInput"
+                [placeholder]="mobileRedirectUris.length === 0 ? t('mobileRedirectUris.placeholder') : ''"
+                (keydown)="onMobileRedirectUriKeydown($event)"
+                (blur)="addMobileRedirectUriFromInput()"
+              />
+            </div>
+          </div>
+          <span class="field-hint">
+            {{ t('mobileRedirectUris.hintPrefix') }}
+            <code>{{ defaultMobileRedirectUri }}</code>
+            {{ '. ' + t('mobileRedirectUris.hintSuffix') }}
+            <code>*</code>
+          </span>
+          @if (mobileRedirectUriErrorKey) {
+            <span class="field-error">{{ t(mobileRedirectUriErrorKey) }}</span>
+          }
+        </div>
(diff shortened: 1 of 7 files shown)


frontend/src/i18n/en/settings-auth.json
```diff
@@ -47,6 +47,24 @@
...
+    "placeholder": "Add a redirect URI or *",
+    "hintPrefix": "The default redirect URI is",
+    "hintSuffix": "To allow any URI, just use",
+    "removeLabel": "Remove {{uri}}",
+    "validation": {
+      "blank": "Redirect URI cannot be blank.",
+      "invalid": "Enter a valid mobile redirect URI.",
+      "schemeRequired": "Redirect URI must include a scheme.",
+      "customSchemeRequired": "Redirect URI must use a custom mobile scheme.",
+      "fragmentNotAllowed": "Redirect URI must not contain a fragment.",
```
imnotjames:
```suggestion
      "fragmentNotAllowed": "Redirect URI must not contain a fragment (#).",
```

imnotjames: People should know what a fragment is but they don't always.

frontend/src/app/core/security/oauth2-management/authentication-settings.component.ts
```diff
@@ -259,6 +279,112 @@ export class AuthenticationSettingsComponent {
...
+
+  private getMobileRedirectUriShapeError(value: string): string | null {
+    const hasScheme = /^[a-zA-Z][a-zA-Z\d+.-]*:/.test(value);
+    if (!hasScheme) {
+      return 'mobileRedirectUris.validation.schemeRequired';
+    }
+
+    try {
+      const uri = new URL(value);
+      if (uri.protocol === 'http:' || uri.protocol === 'https:') {
```
imnotjames: Is there a reason we can't allow other web apps?

zachyale: It was an intentional restriction- the scope of this work, and the security concerns thought through were limited to a feature specifically for allowing customizable mobile deep link redirects, not generic web apps. Allowing those protocols would expand the scope of this and the security model- if there's a relevant use case or compelling argument for allowing this in the future we can address it then, but I think for now it's best to limit the scope of the security changes as much as possible

frontend/src/i18n/en/settings-auth.json
```diff
@@ -47,6 +47,15 @@
     "backchannelLogoutUri": "Back-Channel Logout URI",
     "backchannelLogoutHint": "Configure this URI in your OIDC provider to enable server-initiated session revocation."
   },
+  "mobileRedirectUris": {
```
imnotjames, line 50: Do we want to match these with the generalization we added elsewhere?  So contributions of lang strings can be easier?

frontend/src/app/core/security/oauth2-management/authentication-settings.component.html
```diff
@@ -210,6 +210,44 @@ <h4 class="subsection-title">{{ t('providerConfig.claimMappings') }}</h4>
         </div>
       </div>
 
+      <div class="claims-section">
```
imnotjames, line 213: Is it possible for this to be a separate card for the mobile redirect that makes it distinct from provider config?  If we can't or it's a lot of effort, I won't block on it.

The long version:

There's a bit of a conflation within our API which is being reflected in the UI.  I don't think we can fully fix it right now without far too much blocking on something that can be changed in the future.  We aren't adding APIs, aren't adding any real state that locks us into any approach.

There's a difference between being an OAuth client and an OAuth server.  The way everything is set up right now, Grimmory is operating as both an OAuth client & server when OIDC mode is on.

It's a client authenticating to the provider & also a server providing tokens back to the client (mobile device, here).  We have to follow the standards when interacting with an external server.  We never followed them when implementing an app because of a lack of.. craft, time, care.  Choose any or all.

The OIDC provider configuration is part of the "Client" part of Grimmory.  The OIDC redirect URIs are used as part of the "server".  Given these are two distinct things I believe they should be distinct anyway - and for cases like supporting QR Code auth or the like, that's going to become more important.

Like I said: I sure as heck don't want to change the actual way auth works now, but little things could lead to less churn in the future.

imnotjames (approved):

PR #1317, opened 2026-05-14
refactor(api): resolve static collection memory leak warnings and enforce safety rules

### LINKED ISSUE #1315: Investigate & remediate static mutable collection fields (memory leak risk)
Intellij's static analysis tool flagged several fields of the form:

```java
static List<String> list = new ArrayList<>();
```

Static fields are bound to the class loader, not any object instance, meaning they live for the entire lifetime of the JVM process. When the field holds a mutable, growable collection, any objects added to it can never be garbage collected, even if they are no longer needed by the application. This is a classic memory leak: the GC is working correctly, but the code holds a permanent reference that prevents collection.

Tangible risks with these are:

- In a long-running server process, unbounded static collections grow continuously and are never reclaimed, eventually causing OutOfMemoryError
- Global mutable state can be modified from any class, making behavior unpredictable and hard to test

Generally, these are considered code smells when it comes long running applications (such as this)

balazs-szucs:
## Description

Improved memory safety by replacing mutable static collections with immutable or weakly-referenced alternatives, in line with new static collection and memory leak guidelines. It also refactors some static initializations for better readability and thread safety, and updates related test code accordingly.

## Linked Issue

Fixes: #1315

## Changes

- Static Collections: Converted mutable MEDIA_TYPES, systemDirs, and PROBES to immutable List.of()/Map.of()/Stream.toList().
- Locking: Replaced ConcurrentHashMap with Caffeine weak-value caches in OidcAuthService and DefaultSettingInitializer to ensure lock objects are garbage collected.

## Manual Testing Steps

1. Run just api check to confirm all unit tests pass.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Reduced memory-leak risk by switching per-user locks and initialization state to weak-valued, expiring caches and updating transaction-aware initialization flow.

* **Chores**
  * Replaced mutable static collections with immutable alternatives and standardized lazy/immutable media-type handling.
  * Updated review guidelines to require immutable static collections and safer cache patterns.

* **Tests**
  * Adjusted tests to reset cache-backed lock/state between runs.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1317)

### DIFF
--- a/backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
+++ b/backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
@@ -25,39 +25,31 @@ public final class NativeLibraries {
         EPUB4J_NATIVE
     }
 
-    private static final Map<Library, Probe> PROBES;
-
-    static {
-        Map<Library, Probe> probes = new LinkedHashMap<>();
-
-        probes.put(Library.PDFIUM, new Probe("PDFium", () -> {
-            Boolean clean = tryInvokeStaticBoolean("org.grimmory.pdfium4j.PdfiumLibrary");
-            if (clean != null) {
-                return clean;
-            }
-            PdfiumLibrary.initialize();
-            return true;
-        }));
-
-        probes.put(Library.LIBARCHIVE, new Probe("libarchive", Archive::isAvailable));
-
-        probes.put(Library.EPUB4J_NATIVE, new Probe("epub4j-native", () -> {
-            Boolean clean = tryInvokeStaticBoolean(
-                    "org.grimmory.epub4j.native_parsing.EpubNativeLibrary"
-            );
-            if (clean != null) {
-                return clean;
-            }
-            Class.forName(
-                    "org.grimmory.epub4j.native_parsing.PanamaConstants",
-                    true,
-                    NativeLibraries.class.getClassLoader()
-            );
-            return true;
-        }));
-
-        PROBES = Collections.unmodifiableMap(probes);
-    }
+    private static final Map<Library, Probe> PROBES = Map.of(
+            Library.PDFIUM, new Probe("PDFium", () -> {
+                Boolean clean = tryInvokeStaticBoolean("org.grimmory.pdfium4j.PdfiumLibrary");
+                if (clean != null) {
+                    return clean;
+                }
+                PdfiumLibrary.initialize();
+                return true;
+            }),
+            Library.LIBARCHIVE, new Probe("libarchive", Archive::isAvailable),
+            Library.EPUB4J_NATIVE, new Probe("epub4j-native", () -> {
+                Boolean clean = tryInvokeStaticBoolean(
+                        "org.grimmory.epub4j.native_parsing.EpubNativeLibrary"
+                );
+                if (clean != null) {
+                    return clean;
+                }
+                Class.forName(
+                        "org.grimmory.epub4j.native_parsing.PanamaConstants",
+                        true,
+                        NativeLibraries.class.getClassLoader()
+                );
+                return true;
+            })
+    );
 
     private final Map<Library, Boolean> status;
 
--- a/backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
+++ b/backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
@@ -25,31 +25,35 @@ public final class NativeLibraries {
         EPUB4J_NATIVE
     }
 
-    private static final Map<Library, Probe> PROBES = Map.of(
-            Library.PDFIUM, new Probe("PDFium", () -> {
-                Boolean clean = tryInvokeStaticBoolean("org.grimmory.pdfium4j.PdfiumLibrary");
-                if (clean != null) {
-                    return clean;
-                }
-                PdfiumLibrary.initialize();
-                return true;
-            }),
-            Library.LIBARCHIVE, new Probe("libarchive", Archive::isAvailable),
-            Library.EPUB4J_NATIVE, new Probe("epub4j-native", () -> {
-                Boolean clean = tryInvokeStaticBoolean(
-                        "org.grimmory.epub4j.native_parsing.EpubNativeLibrary"
-                );
-                if (clean != null) {
-                    return clean;
-                }
-                Class.forName(
-                        "org.grimmory.epub4j.native_parsing.PanamaConstants",
-                        true,
-                        NativeLibraries.class.getClassLoader()
-                );
-                return true;
-            })
-    );
+    private static final Map<Library, Probe> PROBES;
+
+    static {
+        var probes = new LinkedHashMap<Library, Probe>();
+        probes.put(Library.PDFIUM, new Probe("PDFium", () -> {
+            Boolean clean = tryInvokeStaticBoolean("org.grimmory.pdfium4j.PdfiumLibrary");
+            if (clean != null) {
+                return clean;
+            }
+            PdfiumLibrary.initialize();
+            return true;
+        }));
+        probes.put(Library.LIBARCHIVE, new Probe("libarchive", Archive::isAvailable));
+        probes.put(Library.EPUB4J_NATIVE, new Probe("epub4j-native", () -> {
+            Boolean clean = tryInvokeStaticBoolean(
+                    "org.grimmory.epub4j.native_parsing.EpubNativeLibrary"
+            );
+            if (clean != null) {
+                return clean;
+            }
+            Class.forName(
+                    "org.grimmory.epub4j.native_parsing.PanamaConstants",
+                    true,
+                    NativeLibraries.class.getClassLoader()
+            );
+            return true;
+        }));
+        PROBES = Collections.unmodifiableSequencedMap(probes);
+    }
 
     private final Map<Library, Boolean> status;
 
--- a/backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
+++ b/backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
@@ -48,19 +48,11 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
     private static final Pattern YEAR_ONLY_PATTERN = Pattern.compile("^\\d{4}$");
     private static final String OPF_NS = "http://www.idpf.org/2007/opf";
 
-    // List of all media types that epub4j has so we can lazy load them.
-    // Note that we have to add in null to handle files without extentions like mimetype.
-    private static final List<MediaType> MEDIA_TYPES = new ArrayList<>();
     private static final Pattern ISBN_SEPARATOR_PATTERN = Pattern.compile("[- ]");
 
     private static final Set<Integer> VALID_AGE_RATINGS = Set.of(0, 6, 10, 13, 16, 18, 21);
 
     private static final ObjectMapper OBJECT_MAPPER = JsonMapper.builder().build();
-
-    static {
-        MEDIA_TYPES.addAll(Arrays.asList(MediaTypes.mediaTypes));
-        MEDIA_TYPES.add(null);
-    }
 
     private static final Map<String, BiConsumer<BookMetadata.BookMetadataBuilder, String>> CALIBRE_IDENTIFIER_PREFIXES = Map.of(
             "amazon", BookMetadata.BookMetadataBuilder::asin,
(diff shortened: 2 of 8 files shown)


backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
```diff
@@ -48,20 +48,12 @@ public class EpubMetadataExtractor implements FileMetadataExtractor {
     private static final Pattern YEAR_ONLY_PATTERN = Pattern.compile("^\\d{4}$");
     private static final String OPF_NS = "http://www.idpf.org/2007/opf";
 
-    // List of all media types that epub4j has so we can lazy load them.
-    // Note that we have to add in null to handle files without extentions like mimetype.
-    private static final List<MediaType> MEDIA_TYPES = new ArrayList<>();
```
imnotjames: This was just never used?

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
```diff
@@ -32,14 +32,17 @@ public class DefaultSettingInitializer {
             .maximumSize(1000)
             .expireAfterWrite(Duration.ofHours(24))
             .build();
-    private static final ConcurrentHashMap<Long, Lock> userLocks = new ConcurrentHashMap<>();
+    private static final Cache<Long, Lock> userLocks = Caffeine.newBuilder()
+            .weakValues()
+            .expireAfterAccess(Duration.ofMinutes(10))
+            .build();
```
imnotjames: Doesn't this drastically change the behavior of the locks?  eg, locks will now unlock after 10 minutes, which they never did before?

imnotjames: Also, you're saying these would not get garbage collected even after they were removed from the hashmap?

backend/src/main/java/org/booklore/service/metadata/extractor/EpubMetadataExtractor.java
balazs-szucs: Yes it was artifact of old pre fork

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
balazs-szucs: > Also, you're saying these would not get garbage collected even after they were removed from the hashmap?

Yes it does. From what i am seeing in the code, this is not problematic.

balazs-szucs: I guess if it helps i can see if i can recreate old functionality, that change is not important.

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
```diff
@@ -32,28 +29,19 @@ public class DefaultSettingInitializer {
...
             }
             patchPerBookSetting(user);
             userRepository.save(user);
-            initializedUsers.put(bookLoreUser.getId(), Boolean.TRUE);
-        } finally {
-            lock.unlock();
-            userLocks.remove(bookLoreUser.getId());
-        }
+            return Boolean.TRUE;
+        });
```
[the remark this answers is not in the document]
balazs-szucs: oh... thats not good i guess.

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
balazs-szucs: Okey i cannot make it work otherwise. From a pure functionality perspective this is better because that 10 mins is for _stale_ locks. So if there lock that is being used, that will not be removed after 10 mins.

backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
```diff
@@ -28,8 +28,7 @@ public enum Library {
     private static final Map<Library, Probe> PROBES;
 
     static {
-        Map<Library, Probe> probes = new LinkedHashMap<>();
-
+        var probes = new LinkedHashMap<Library, Probe>();
```
imnotjames, line 31: This is just a stylistic decision, right?

backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
```diff
@@ -55,8 +52,7 @@ public enum Library {
             );
             return true;
         }));
-
-        PROBES = Collections.unmodifiableMap(probes);
+        PROBES = Collections.unmodifiableSequencedMap(probes);
```
imnotjames, line 55: Based on the description I'm guessing this is a thread safety thing?  How so?

backend/src/main/java/org/booklore/util/epub/EpubContentReader.java
```diff
@@ -16,22 +16,22 @@
...
 
 @Slf4j
 public class EpubContentReader {
 
-    private static final List<MediaType> MEDIA_TYPES = new ArrayList<>();
-    static {
-        MEDIA_TYPES.addAll(Arrays.asList(MediaTypes.mediaTypes));
-        MEDIA_TYPES.add(null);
-    }
+    private static final List<MediaType> LAZY_MEDIA_TYPES = Stream.concat(
```
imnotjames, line 24: What makes these lazy vs just media types?

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
imnotjames: If you have a lot of users or a malicious OIDC server this would be a problem.  It's not so much otherwise.  Given we're concerned about that case - as that's the only non-hypothetical case where we'd have a memory leak here in a real world scenario - this only means "how many requests can you make in a 10 minute period and will that be a memory leak".

Then reading the default setting initializer, this actually doesn't protect any of the concurrency issues.  Because we remove the lock from the concurrency hash map after we are done with it, that means that a third request that comes in while the second is processing wouldn't see any lock at all.  This is incorrect.

Looking at this and thinking about it:

These locks are meant to prevent multiple concurrent user initialization calls or OIDC requests from occurring at the same time.  Given `provisionOidcUser()` is always called in a transaction and we have proper constraints on the `username` field, the actual data integrity isn't a concern.  So the lock usage is to prevent a failed request because of a race condition.

With all that in mind, instead of using an unbounded concurrent hash map we could take two approaches and still maintain safety:

* Just use one Reentrant Lock
* Use multiple re-entrant locks similar to `ConcurrentHashMap` does with a concurrency level.

One reentrant lock would be easiest.  You just replace the hash map with a single lock.  Pretty straightforward.  The drawback is on the performance side.  Only one request could be processed at a time, limiting the login rate.  Given we are building a self-hosted utility to support at most tens or hundreds of users per instance, this is probably okay.

If it's not, or if we run into this in the future, the alternative would be to use multiple re-entrant locks.

```java
class HashedReentrantLocks {
    private final boolean fair;
    private final int concurrencyLevel;
    private final ReentrantLock[] locks;

    HashedReentrantLocks(int concurrencyLevel) {
        this(concurrencyLevel, false);
    }

    HashedReentrantLocks(int concurrencyLevel, boolean fair) {
        this.fair = fair;
        this.concurrencyLevel = concurrencyLevel;
        locks = new ReentrantLock[concurrencyLevel];
    }

    public ReentrantLock getLock(Object obj) {
        int bucket = obj.hashCode() & (concurrencyLevel - 1);
        ReentrantLock lock = locks[bucket];
        if (lock == null) {
            lock = locks[bucket] = new ReentrantLock(this.fair);
        }

        return lock;
    }
}
```

Here we have a consistent number of locks, can control the concurrency level, and still provide the safety of a lock.

As long as the object's hash code is properly distributed you wouldn't get hot spots and could increase concurrency - but I think this might be overkill.  We don't even need more than one lock.

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
```diff
@@ -32,27 +33,44 @@ public class DefaultSettingInitializer {
...
-            initializedUsers.put(bookLoreUser.getId(), Boolean.TRUE);
+
+            if (TransactionSynchronizationManager.isSynchronizationActive()) {
+                TransactionSynchronizationManager.registerSynchronization(new TransactionSynchronization() {
+                    @Override
+                    public void afterCommit() {
+                        initializedUsers.put(userId, Boolean.TRUE);
+                    }
+                });
+            } else {
```
imnotjames, line 69: Why was this not needed before but is now?

balazs-szucs: It was a suggestion from the rabbit: https://github.com/grimmory-tools/grimmory/pull/1317#discussion_r3247811195

i think it's pretty nice. Looks idiomatic for me.

backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
balazs-szucs: It was response to: https://github.com/grimmory-tools/grimmory/pull/1317#discussion_r3244196631

Specifically: 

> If stable log ordering is desired, consider Map.ofEntries() with **explicit ordering** or document that the order may vary.

`unmodifiableSequencedMap` preserves order so callers can rely on insertion/sort order through the API

backend/src/main/java/org/booklore/nativelib/NativeLibraries.java
balazs-szucs: Yes

backend/src/main/java/org/booklore/util/epub/EpubContentReader.java
balazs-szucs: The media types are being concatenated with mediatype null, and only matched to their actual media type lazily :)

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
balazs-szucs: > Then reading the default setting initializer, this actually doesn't protect any of the concurrency issues. Because we remove the lock from the concurrency hash map after we are done with it, that means that a third request that comes in while the second is processing wouldn't see any lock at all. This is incorrect.

The lock stays alive while it is being strongly referenced e.g., by logging attempts, so a concurrent request should still get the same lock instance. That is the point of `weakValues()`

balazs-szucs: > One reentrant lock would be easiest. You just replace the hash map with a single lock.

I'd actually argue the opposite, a Caffeine cache with `weakValues()` handles the lifecycle automatically, whereas a bare ReentrantLock pushes that complexity onto the caller. Less moving parts to reason about -> less complexity on caller side -> easiest solution

imnotjames: > > Then reading the default setting initializer, this actually doesn't protect any of the concurrency issues. Because we remove the lock from the concurrency hash map after we are done with it, that means that a third request that comes in while the second is processing wouldn't see any lock at all. This is incorrect.
> 
> The lock stays alive while it is being strongly referenced e.g., by logging attempts, so a concurrent request should still get the same lock instance. That is the point of `weakValues()`

Then why do you need the time aspect?  And isn't this still unbounded but in a different way?

> > One reentrant lock would be easiest. You just replace the hash map with a single lock.
> 
> I'd actually argue the opposite, a Caffeine cache with `weakValues()` handles the lifecycle automatically, whereas a bare ReentrantLock pushes that complexity onto the caller. Less moving parts to reason about -> less complexity on caller side -> easiest solution

I don't understand.  How is a single lock & unlock across the entirety of OIDC less complex than getting a lock and having to reason about garbage collection when dealing with concurrency?

imnotjames: Whatever.  I don't care.  It's all a micro-optimization for a scenario that won't happen anyway.  If it breaks stuff I have to fix it either way.  Resolving.

backend/src/main/java/org/booklore/service/user/DefaultSettingInitializer.java
imnotjames: If this happens after the transaction commit doesn't this mean that we could be running the user settings code twice?

1. Thread A starts transaction
1. Thread A enters lock
1. Thread B starts transaction
1. Thread B waits for lock
1. Thread A checks that user has not been initialized yet.
2. Thread A initializes user.
3. Thread A schedules post-commit task
1. Thread A exits lock
1. Thread B enters lock
2. Thread B checks that user has not been initialized yet
3. Thread A commits transaction
4. Thread A inserts `true` for user
5. Thread B initializes user.
3. Thread B schedules post-commit task
7. Thread B exits lock
8. Thread B commits
9. Thread B inserts `true` for user

How does this resolve the concurrency issue there?

PR #1378, opened 2026-05-18
fix(email): pass correct mail properties for SSL transport

### LINKED ISSUE #1301: Email send to SMTPS-only providers (e.g. GMail port 465) silently fails
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1289

<sup>Originally posted by **monsagri** May 13, 2026</sup>
### What went wrong?

When an email provider is configured with port 465 + SSL (i.e. SMTPS / TLS-on-connect, the alternative if port 587 is blocked at router/ISP level), Grimmory connects to the upstream server but sends the message un-authenticated.  
Gmail (and any provider that requires auth) drops the message. The connection is accepted and no error is logged. There's no failure in the UI either, but the email never arrives.

I'm happy to open a PR if this is promoted to an issue, but want to stick to contribution guidelines so I'm just providing a bit more detail below either way!

**Workaround**
I've been able to work around this via running an SMTP relay in docker and passing the unauthenticated message from Grimmory through that and forwarding it after forwarding the message to Gmail with proper SMTP authentication.

**Root cause**

[SendEmailV2Service.setupMailSender](https://github.com/grimmory-tools/grimmory/blob/v3.0.3/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java#L128) configures auth on the `mail.smtp.* namespace`. [configureConnectionType](https://github.com/grimmory-tools/grimmory/blob/v3.0.3/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java#L154) selects the smtps transport for SSL connections.  
When `mail.transport.protocol=smtps`, Jakarta Mail's transport class is `org.eclipse.angus.mail.smtp.SMTPSSLTransport`, which reads its properties under the `mail.smtps.*` prefix and `mail.smtp.*` keys are ignored.

There is a [notification + error-log path](https://github.com/grimmory-tools/grimmory/blob/v3.0.3/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java#L84-L88)) on send failure, but it never fires in this bug's reproduction case because timeouts set by [configureTimeouts](https://github.com/grimmory-tools/grimmory/blob/v3.0.3/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java#L185-L187) are also mail.smtp.* keys  and get ignored by SMTPS transport.

**Suggested fix**
In `configureConnectionType`, the SSL branch needs to write properties under the mail.smtps.* prefix (or both prefixes). Writing to both prefixes unconditionally would not be an issue as the unneeded config would be ignored. Same thing for `configureTimeouts`

### How can we reproduce it?

1. Configure an email provider in Grimmory
    - Host: smtp.gmail.com
    - Port: 465
    - Encryption: SSL (i.e. start_tls=false in DB, auth=true)
    - Username: a Gmail address
    - Password: a valid Gmail app password
  2. Add a recipient (any address)
  3. Send any book via the email feature
  4. Observe 'Send initiated' log in grimmory
  5. Don't receive email

For more debug information, set `JAVA_TOOL_OPTIONS=-Dmail.debug=true -Dmail.smtps.debug=true` to observe issue in debug logs.

Stable

### Your setup

- Grimmory version: 3.0.3
- Running via: Docker using ghcr.io/grimmory-tools/grimmory:v3.0.3
- Browser: Chrome (Any browser)
- Operating System: Ubuntu
- Outbound port 587 blocked by ISP

Debug logs from enabling email debugging as mentioned in repro steps

```
  INFO ... Email dispatch initiated for book: <book> to <addr>
  INFO ... Email configuration: Host=smtp.gmail.com, Port=465, Type=SSL, Timeouts=60s
  DEBUG: Jakarta Mail version 2.1.5
  DEBUG: getProvider() returning jakarta.mail.Provider[TRANSPORT,smtps,org.eclipse.angus.mail.smtp.SMTPSSLTransport,Oracle]
  DEBUG SMTP: useEhlo true, useAuth false      <-- this is the bug
  DEBUG SMTP: trying to connect to host "smtp.gmail.com", port 465, isSSL true
```

### Before submitting

monsagri:
## Description

Email sending via SSL providers (e.g. Gmail on port 465) silently failed because Jakarta Mail's `SMTPSSLTransport` reads properties only from the `mail.smtps.*` namespace, while `SendEmailV2Service` was writing all properties under `mail.smtp.*`. The transport ignored those properties and providers dropped the unauthenticated message with no UI feedback.

## Linked Issue

Fixes #1301 

## Changes

* Updates `configureConnectionType` and `configureTimeouts`  to accept `prefix` input to support `mail.smtp` and `mail.smtps` prefixes
* Updates `setupMailSender` to pass appropriate prefix based on `ConnectionType` 
* Adds unit tests to ensure that (only) appropriate fields are populated 

## Manual Testing Steps

1. Ran grimmory locally via `just dev-up` (off `develop` branch) 
2. Reproduced issue as described in repro steps in #1301
3. Ran locally from fix branch
4. Observed email being sent successfully

* Includes a related fix by ensuring `ssl.protocols` are space-separated. Previous comma-separated syntax was incorrect but never being hit due to the original bug we're fixing here.
 
## AI Disclosure

Cursor 4.7 used for local code review and writing unit tests.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved email delivery configuration to pick the correct SMTP vs SMTPS settings per provider, write SSL-specific properties under the secure namespace, honor JVM timeout overrides per connection type, and normalize TLS protocol handling for more reliable secure connections.
* **Tests**
  * Added unit tests for SMTP over SSL, SMTP with STARTTLS, and plain SMTP to validate provider-specific behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1378?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
+++ b/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
@@ -117,6 +117,15 @@ public class SendEmailV2Service {
                 .orElseThrow(() -> ApiError.FILE_NOT_FOUND.createException(bookFileId));
     }
 
+    /**
+     * Builds a configured {@link JavaMailSenderImpl} for the given provider.
+     *
+     * <p>Selects the Jakarta Mail property namespace ({@code mail.smtp.} or {@code mail.smtps.})
+     * from the resolved {@link ConnectionType} and writes authentication, SSL/STARTTLS,
+     * and timeout properties under that prefix so the active transport actually reads them.
+     *
+     * <p>Package-private to allow same-package unit tests to assert the resulting property map.
+     */
     JavaMailSenderImpl setupMailSender(EmailProviderV2Entity emailProvider) {
         JavaMailSenderImpl dynamicMailSender = new JavaMailSenderImpl();
         dynamicMailSender.setHost(emailProvider.getHost());
@@ -154,6 +163,13 @@ public class SendEmailV2Service {
         }
     }
 
+    /**
+     * Writes transport-protocol, SSL, and STARTTLS properties for the given connection type.
+     *
+     * <p>{@code mail.transport.protocol} is namespace-independent (it selects the transport
+     * implementation). All other keys are written under {@code prefix}, which must match the
+     * namespace the active transport reads from.
+     */
     private void configureConnectionType(Properties mailProps, ConnectionType connectionType, EmailProviderV2Entity emailProvider, String prefix) {
         switch (connectionType) {
             case SSL -> {
@@ -162,7 +178,6 @@ public class SendEmailV2Service {
                 mailProps.put(prefix + "ssl.trust", emailProvider.getHost());
                 mailProps.put(prefix + "starttls.enable", "false");
                 mailProps.put(prefix + "ssl.protocols", "TLSv1.2 TLSv1.3");
-                mailProps.put(prefix + "ssl.checkserveridentity", "false");
                 mailProps.put(prefix + "ssl.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
                 mailProps.put(prefix + "ssl.socketFactory.fallback", "false");
             }
@@ -180,10 +195,22 @@ public class SendEmailV2Service {
         }
     }
 
+    /**
+     * Writes connection, socket, and write timeout properties under {@code prefix}.
+     *
+     * <p>Default is 60s per timeout. JVM operators can override per-transport via
+     * {@code -D<prefix>connectiontimeout=...} or fall back to the legacy {@code mail.smtp.*}
+     * keys for backward compatibility with deployments that predate transport-aware overrides.
+     */
     private void configureTimeouts(Properties mailProps, String prefix) {
-        String connectionTimeout = System.getProperty("mail.smtp.connectiontimeout", "60000");
-        String socketTimeout = System.getProperty("mail.smtp.timeout", "60000");
-        String writeTimeout = System.getProperty("mail.smtp.writetimeout", "60000");
+        // Prefer transport-namespaced JVM overrides (e.g. -Dmail.smtps.connectiontimeout=...)
+        // and fall back to the unprefixed mail.smtp.* form for backward compatibility.
+        String connectionTimeout = System.getProperty(prefix + "connectiontimeout",
+                System.getProperty("mail.smtp.connectiontimeout", "60000"));
+        String socketTimeout = System.getProperty(prefix + "timeout",
+                System.getProperty("mail.smtp.timeout", "60000"));
+        String writeTimeout = System.getProperty(prefix + "writetimeout",
+                System.getProperty("mail.smtp.writetimeout", "60000"));
 
         mailProps.put(prefix + "connectiontimeout", connectionTimeout);
         mailProps.put(prefix + "timeout", socketTimeout);
--- a/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
+++ b/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
@@ -130,7 +130,7 @@ public class SendEmailV2Service {
         // SMTPTransport reads only mail.smtp.*, SMTPSSLTransport reads only mail.smtps.*.
         // Pick the prefix matching the active transport so auth/ssl/timeout settings apply.
         String prefix = connectionType == ConnectionType.SSL ? "mail.smtps." : "mail.smtp.";
-        mailProps.put(prefix + "auth", emailProvider.isAuth());
+        mailProps.put(prefix + "auth", String.valueOf(emailProvider.isAuth()));
         configureConnectionType(mailProps, connectionType, emailProvider, prefix);
         configureTimeouts(mailProps, prefix);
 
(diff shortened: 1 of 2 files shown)


backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
```diff
@@ -151,40 +154,40 @@ private ConnectionType determineConnectionType(EmailProviderV2Entity emailProvid
...
-                mailProps.put("mail.smtp.ssl.trust", emailProvider.getHost());
-                mailProps.put("mail.smtp.starttls.enable", "false");
-                mailProps.put("mail.smtp.ssl.protocols", "TLSv1.2,TLSv1.3");
-                mailProps.put("mail.smtp.ssl.checkserveridentity", "false");
-                mailProps.put("mail.smtp.ssl.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
-                mailProps.put("mail.smtp.ssl.socketFactory.fallback", "false");
+                mailProps.put(prefix + "ssl.enable", "true");
+                mailProps.put(prefix + "ssl.trust", emailProvider.getHost());
+                mailProps.put(prefix + "starttls.enable", "false");
+                mailProps.put(prefix + "ssl.protocols", "TLSv1.2 TLSv1.3");
```
monsagri: Mentioned this in the PR description - switching from comma separated to space separated here is a secondary fix for an issue that only surfaced once the SSL switch case was actually being excercised.

backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
```diff
@@ -151,40 +154,40 @@ private ConnectionType determineConnectionType(EmailProviderV2Entity emailProvid
...
+                mailProps.put(prefix + "ssl.checkserveridentity", "false");
```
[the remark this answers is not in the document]
monsagri: ~~This is pre-existing.~~

~~Happy to pull that change in, but not sure if there's historical context/reasoning behind the "false" setting~~

Changed my mind on this, the underlying default is `true` and since this config was actually never being read, that's the real behaviour prior to this PR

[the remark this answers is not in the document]
monsagri: Addressed in follow-up commit

backend/src/test/java/org/booklore/service/email/SendEmailV2ServiceTest.java
```diff
@@ -307,6 +310,78 @@ void emailBookQuick_sendEmailFailure_logsError() {
...
+        // so auth/ssl/timeout properties must be visible under that prefix.
+        assertThat(props.get("mail.smtps.auth")).isEqualTo("true");
+        assertThat(props.get("mail.smtps.ssl.enable")).isEqualTo("true");
+        assertThat(props.get("mail.smtps.ssl.trust")).isEqualTo("smtp.gmail.com");
+        // Whitespace-separated, not comma-separated: Angus Mail's SocketFetcher
+        // passes the raw string to SSLSocket.setEnabledProtocols which splits on whitespace.
+        assertThat(props.get("mail.smtps.ssl.protocols")).isEqualTo("TLSv1.2 TLSv1.3");
+        assertThat(props.get("mail.smtps.connectiontimeout")).isEqualTo("60000");
+        assertThat(props.get("mail.smtps.timeout")).isEqualTo("60000");
+        assertThat(props.get("mail.smtps.writetimeout")).isEqualTo("60000");
```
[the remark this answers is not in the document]
monsagri: Addressed in follow-up commit

backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java
```diff
@@ -151,40 +163,58 @@ private ConnectionType determineConnectionType(EmailProviderV2Entity emailProvid
...
             case SSL -> {
                 mailProps.put("mail.transport.protocol", "smtps");
-                mailProps.put("mail.smtp.ssl.enable", "true");
-                mailProps.put("mail.smtp.ssl.trust", emailProvider.getHost());
-                mailProps.put("mail.smtp.starttls.enable", "false");
-                mailProps.put("mail.smtp.ssl.protocols", "TLSv1.2,TLSv1.3");
-                mailProps.put("mail.smtp.ssl.checkserveridentity", "false");
-                mailProps.put("mail.smtp.ssl.socketFactory.class", "javax.net.ssl.SSLSocketFactory");
-                mailProps.put("mail.smtp.ssl.socketFactory.fallback", "false");
+                mailProps.put(prefix + "ssl.enable", "true");
```
imnotjames: In what case would this not be `mail.smtps`?  I'm trying to follow the logic and I think I've got a disconnect where I'm not seeing that

balazs-szucs: I believe: https://github.com/monsagri/grimmory/blob/f245b497d8186424941340250dee82ea3e46380f/backend/src/main/java/org/booklore/service/email/SendEmailV2Service.java#L141 is the relevant snippet here.

I think it looks like it derives a single prefix from the ConnectionType. Should be good. But I am not a mail-expert.

balazs-szucs: https://github.com/grimmory-tools/grimmory/pull/1378/changes#diff-fcf2704422c6c6a8f9dc26ad090bae5d909887bc203d74b804374c756cfe58b3R329

also seems to be testing this question 🤔

imnotjames: Right but on this specific line:

The case is for connection type SSL.  The prefix is derived by connection type SSL - is there any case where the prefix on the line I noted is ever different?

The tests seem to say no.

monsagri: Sorry about the delay, was out for a while. 

You're right that the prefix calculation isn't strictly necessary in the smtp or smptps blocks, I've simplified to have those be hardcoded and only use the prefix var where there actually variable values involved. 

Hope I interpreted the questions in the thread correctly, lmk!

imnotjames: You did!  Thank you.

imnotjames (approved):
Looks good to me, builds, I can still send email, and seems to change the values as expected.  I don't have any mail servers that operate on different ports but can see the mail sending attempt to connect to the ports I define.

imnotjames:
Thanks for the contribution!

PR #1407, opened 2026-05-19
refactor(metadata): Normalize language field to ISO 639-1

### LINKED ISSUE #1383: Normalize language field to ISO 639-1 across all metadata providers
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1382

<sup>Originally posted by **ghostdiverr** May 18, 2026</sup>

Add a `LanguageNormalizer` utility that maps the raw `language` value returned by each metadata provider to a consistent ISO 639-1 two-letter code at ingestion time, before it reaches the database.

Providers currently return the language field in incompatible formats:
  | Provider | Example output | Format |
  |---|---|---|
  | Google, Hardcover, RanobeDB, Audible, LubimyCzytac | `"fr"`, `"en"` | ISO 639-1 ✓ |
  | Amazon | `"French"`, `"English"` | Full English name ✗ |
  | GoodReads | `"English"`, `"German"` | Full English name ✗ |

The normalizer would handle: ISO 639-1 passthrough, ISO 639-2 codes (`"fre"` → `"fr"`), full English names (`"French"` → `"fr"`), native names (`"Français"` → `"fr"`), and BCP 47 tags (`"fr-FR"` → `"fr"`). Unknown values would be lowercased and returned as-is — no data loss.

Note: `GoogleParser` already has a local `normalizeLanguage()` method that does a subset of this. The proposal is to centralize and generalize it.

The language filter in the library view breaks silently across providers: a book imported from Amazon with `"French"` won't match one from Google with `"fr"`, even though they're the same language. As a user with a mixed-provider library, filtering by language produces incomplete results depending on which provider sourced each book.

A consistent language field also unlocks other features. For example, the DuckDuckGo cover search could automatically use the book's language to localize search terms (searching `"livre"` instead of `"book"` for French titles), which significantly improves cover relevance for non-English libraries.

### Anything else?

The change is fully backward-compatible. No schema change, no migration, existing data untouched. I can work on a PoC if interested.

As a follow-up, a one-shot "normalize existing library languages" action could be added to let users clean up data imported before this fix. That's entirely optional and independent of the core change.

### Want to help out?

### Before submitting

ghostdiverr:
## Description

Normalises book metadata language values to ISO 639-1 two-letter codes. All parsers (Amazon, GoodReads, Google, Audible, Douban, RanobeDb, LubimyCzytac, Hardcover) now go through a shared `LanguageNormalizer` utility that handles ISO 639-1/639-2 codes, full English names, BCP 47 tags, and localized names via JDK CLDR fallback.

## Linked Issue

Fixes #1383

## Changes

**Backend**
 - Added `LanguageNormalizer` (nv-i18n + JDK CLDR): ISO 639-1 passthrough, ISO 639-2 alpha-3, full English names, BCP 47 tags, CLDR fallback for localized names (e.g. "français" → `fr`, "Französisch" → `fr`)
 - Wired `LanguageNormalizer` into all 8 parsers (Amazon, GoodReads, Google, Audible, Douban, RanobeDb, LubimyCzytac, Hardcover)
 - Removed duplicate `mapLanguage()` from `LubimyCzytacParser` (covered by CLDR)
 - CLDR name map is lazily initialized. No startup cost if the normalizer is never hit with a localized name

**Tests**
 - New `LanguageNormalizerTest.java`: null/blank, ISO 639-1 passthrough, ISO 639-2 alpha-3, full English names, BCP 47 tags, French/German/Spanish localized names → ISO codes, unknown fallback
  
## Manual Testing Steps

1. Fetch metadata from a provider that returns a raw language value (e.g. `"French"`, `"français"`, `"fre"`) and verify the value is saved as the normalised ISO 639-1 code (e.g. `fr`).   
 2. Run `just api check`.

## Screenshots (Optional)

N/A — backend-only change.

The CLDR map covers localized names across 9 display locales (French, German, Italian, Spanish, Portuguese, Dutch, Russian, Polish, Japanese). English names are handled by nv-i18n's `LanguageCode.findByName()` and don't require the CLDR fallback.

## AI Disclosure
  
I used Claude Code which helped me design some aspects of `LanguageNormalizer`, generate and fix tests, and draft this PR description. All changes were reviewed, validated and tested.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Implemented standardized language normalization across all book sources (Amazon, Audible, Douban, GoodReads, Google, Hardcover, and others) for consistent language code representation.

* **Chores**
  * Added internationalization library dependency to support language code resolution and normalization.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1407?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/parser/RanobeDbParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/RanobeDbParser.java
@@ -10,6 +10,7 @@ import org.booklore.model.dto.response.ranobedbapi.RanobedbSearchResponse;
 import org.booklore.model.enums.MetadataProvider;
 import org.booklore.service.appsettings.AppSettingService;
 import org.booklore.util.BookUtils;
+import org.booklore.util.LanguageNormalizer;
 import org.springframework.stereotype.Service;
 import org.springframework.web.util.UriComponentsBuilder;
 import tools.jackson.databind.ObjectMapper;
@@ -259,7 +260,7 @@ public class RanobeDbParser implements BookParser {
                     .publisher(englishPublisher != null ? englishPublisher.getName() : null)
                     .thumbnailUrl(book.getImage() != null ? RANOBEDB_IMAGE_URL + book.getImage().getFilename() : null)
                     .description(book.getDescription())
-                    .language(englishRelease != null ? englishRelease.getLang() : book.getLang())
+                    .language(LanguageNormalizer.normalize(englishRelease != null ? englishRelease.getLang() : book.getLang()))
                     .seriesName(book.getSeries() != null ? book.getSeries().getTitle() : null)
                     .seriesNumber(seriesIndex != -1 ? seriesIndex + 1.0f : null)
                     .seriesTotal(seriesBooks.isEmpty() ? null : seriesBooks.size())
--- a/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
+++ b/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
@@ -30,7 +30,7 @@ public class LanguageNormalizer {
             result = resolveCode(trimmed);
         }
 
-        return result != null ? result : trimmed.toLowerCase().strip();
+        return result != null ? result : trimmed.toLowerCase(Locale.ROOT).strip();
     }
 
     private static String resolveCode(String input) {
--- a/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
+++ b/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
@@ -1,0 +1,57 @@
+package org.booklore.util;
+
+import com.neovisionaries.i18n.LanguageAlpha3Code;
+import com.neovisionaries.i18n.LanguageCode;
+
+import java.util.List;
+import java.util.regex.Pattern;
+
+public class LanguageNormalizer {
+
+    private LanguageNormalizer() {}
+
+    public static String normalize(String input) {
+        if (input == null || input.isBlank()) {
+            return null;
+        }
+        String trimmed = input.trim();
+
+        // Handle BCP 47 / IETF tags ("fr-FR", "fr_FR") — take the primary subtag
+        String primary = trimmed;
+        if (trimmed.contains("-") || trimmed.contains("_")) {
+            primary = trimmed.split("[-_]")[0];
+        }
+
+        String result = resolveCode(primary);
+        if (result == null && !primary.equals(trimmed)) {
+            result = resolveCode(trimmed);
+        }
+
+        return result != null ? result : trimmed.toLowerCase().strip();
+    }
+
+    private static String resolveCode(String input) {
+        // ISO 639-1 (case-insensitive): "fr", "FR"
+        LanguageCode byCode = LanguageCode.getByCode(input, false);
+        if (byCode != null) {
+            return byCode.name();
+        }
+
+        // ISO 639-2 alpha-3 (case-insensitive): "fre", "fra", "eng"
+        LanguageAlpha3Code byAlpha3 = LanguageAlpha3Code.getByCode(input, false);
+        if (byAlpha3 != null) {
+            LanguageCode alpha2 = byAlpha3.getAlpha2();
+            if (alpha2 != null) {
+                return alpha2.name();
+            }
+        }
+
+        // English name match (case-insensitive): "French", "German", etc.
+        List<LanguageCode> byName = LanguageCode.findByName("(?i)" + Pattern.quote(input));
+        if (!byName.isEmpty()) {
+            return byName.get(0).name();
+        }
+
+        return null;
+    }
+}
--- a/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
+++ b/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
@@ -13,17 +13,15 @@ import java.util.regex.Pattern;
 @UtilityClass
 public class LanguageNormalizer {
 
-    public static String normalize(String input) {
+    private static final Pattern BCP47_SEPARATOR = Pattern.compile("[-_]");
+
+    public String normalize(String input) {
         if (input == null || input.isBlank()) {
             return null;
         }
         String trimmed = input.trim();
 
-        // BCP 47 / IETF tags ("fr-FR", "fr_FR") — take primary subtag first
-        String primary = trimmed;
-        if (trimmed.contains("-") || trimmed.contains("_")) {
-            primary = trimmed.split("[-_]")[0];
-        }
+        String primary = BCP47_SEPARATOR.split(trimmed)[0];
 
         String result = resolveCode(primary);
         if (result == null && !primary.equals(trimmed)) {
@@ -33,7 +31,7 @@ public class LanguageNormalizer {
         return result != null ? result : trimmed.toLowerCase(Locale.ROOT).strip();
     }
 
-    private static String resolveCode(String input) {
+    private String resolveCode(String input) {
         LanguageCode byCode = LanguageCode.getByCode(input, false);
         if (byCode != null) return byCode.name();
 
@@ -44,7 +42,7 @@ public class LanguageNormalizer {
         }
 
         List<LanguageCode> byName = LanguageCode.findByName("(?i)" + Pattern.quote(input));
-        if (!byName.isEmpty()) return byName.get(0).name();
+        if (!byName.isEmpty()) return byName.getFirst().name();
 
         return LocalizedNameMap.MAP.get(input.toLowerCase(Locale.ROOT).trim());
     }
@@ -54,14 +52,14 @@ public class LanguageNormalizer {
 
         private static Map<String, String> build() {
             Locale[] displayLocales = {
-                Locale.FRENCH, Locale.GERMAN, Locale.ITALIAN,
-                new Locale("es"), new Locale("pt"), new Locale("nl"),
-                new Locale("ru"), new Locale("pl"), new Locale("ja")
+                Locale.of("fr"), Locale.of("de"), Locale.of("it"), Locale.of("ja"),
+                Locale.of("es"), Locale.of("pt"), Locale.of("nl"),
+                Locale.of("ru"), Locale.of("pl")
             };
             Map<String, String> map = new HashMap<>();
             for (LanguageCode code : LanguageCode.values()) {
                 if (code == LanguageCode.undefined) continue;
-                Locale locale = new Locale(code.name());
+                Locale locale = Locale.of(code.name());
                 for (Locale displayLocale : displayLocales) {
                     String displayName = locale.getDisplayLanguage(displayLocale).toLowerCase(Locale.ROOT);
                     if (displayName.length() > 2) {
--- a/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
+++ b/backend/src/main/java/org/booklore/util/LanguageNormalizer.java
@@ -2,13 +2,16 @@ package org.booklore.util;
 
 import com.neovisionaries.i18n.LanguageAlpha3Code;
 import com.neovisionaries.i18n.LanguageCode;
+import lombok.experimental.UtilityClass;
 
+import java.util.HashMap;
 import java.util.List;
+import java.util.Locale;
+import java.util.Map;
 import java.util.regex.Pattern;
 
+@UtilityClass
 public class LanguageNormalizer {
-
-    private LanguageNormalizer() {}
 
     public static String normalize(String input) {
         if (input == null || input.isBlank()) {
@@ -16,7 +19,7 @@ public class LanguageNormalizer {
         }
         String trimmed = input.trim();
 
-        // Handle BCP 47 / IETF tags ("fr-FR", "fr_FR") — take the primary subtag
+        // BCP 47 / IETF tags ("fr-FR", "fr_FR") — take primary subtag first
         String primary = trimmed;
         if (trimmed.contains("-") || trimmed.contains("_")) {
             primary = trimmed.split("[-_]")[0];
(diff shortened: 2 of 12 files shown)


backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,79 @@
...
+import com.neovisionaries.i18n.LanguageAlpha3Code;
+import com.neovisionaries.i18n.LanguageCode;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+public class LanguageNormalizer {
```
balazs-szucs, line 14: Use lombok utilityclass

balazs-szucs:
Sonarcloud is complaning because of code duplication at:

frontend/src/app/features/metadata/component/book-metadata-center/metadata-picker/metadata-picker.component.html

balazs-szucs:
> A POST /metadata/manage/normalize-languages endpoint allows migrating all existing values in a single pass.


Isn't this largely overkill? I feel like it is. In-case of migration people can just re-fetch language there is no need for dedicated endpoint, and this PR becomes probably sub-200 LOC in that case.

balazs-szucs (commented):
I feel like the worthwhile changes here: are the utility class + parser changes and that's it. The rest of code I feel like is overkill even if it was perfect, but it also smells 😢

ghostdiverr:
> > A POST /metadata/manage/normalize-languages endpoint allows migrating all existing values in a single pass.
> 
> Isn't this largely overkill? I feel like it is. In-case of migration people can just re-fetch language there is no need for dedicated endpoint, and this PR becomes probably sub-200 LOC in that case.

You're right, it is a one-shot operation. My main concern with forcing a mass re-fetch was triggering provider rate limits for users with large libraries. 
To keep the PR focused, I will remove the endpoint and the UI button right away.

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,79 @@
...
+            if (code == LanguageCode.undefined) continue;
+            Locale locale = new Locale(code.name());
+            for (Locale displayLocale : displayLocales) {
+                String displayName = locale.getDisplayLanguage(displayLocale).toLowerCase(Locale.ROOT);
+                // skip trivial cases where the display name is the ISO code itself
+                if (displayName.length() > 2) {
+                    map.putIfAbsent(displayName, code.name());
+                }
+            }
+        }
```
balazs-szucs: I would be much happier if we could load on demand and lazily not at startup time at once everything. I do not know how big LanguageCode actually is, but I would think, it is rather large.

ghostdiverr:
> I feel like the worthwhile changes here: are the utility class + parser changes and that's it. The rest of code I feel like is overkill even if it was perfect, but it also smells 😢

Fair enough! I definitely went down a rabbit hole with the UI and over-engineered it. 😅

My main worry was that leaving it as a free-text input would let users manually break the normalization we just fixed. But you're right, if someone really wants to type a custom value manually, that's their choice.

I'll strip the PR down to the bare minimum: just the LanguageNormalizer and the parsers. Let's get the core backend fixed first, and we can always chat about the UI later if it becomes an issue.

I'll update the PR shortly!

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
ghostdiverr: Done

backend/src/main/java/org/booklore/controller/LanguageController.java
```diff
@@ -0,0 +1,36 @@
...
+@RequestMapping("/api/v1/languages")
+@Tag(name = "Languages", description = "ISO 639-1 language list")
+public class LanguageController {
+
+    public record LanguageDto(String code, String name) {}
+
+    @Operation(summary = "Get all supported languages", description = "Returns all ISO 639-1 language codes with names localized to the requested language.")
+    @GetMapping
+    public List<LanguageDto> getLanguages(@RequestParam(defaultValue = "en") String lang) {
+        Locale displayLocale = Locale.forLanguageTag(lang);
```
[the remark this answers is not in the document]
ghostdiverr: I removed this endpoint

backend/src/main/java/org/booklore/service/metadata/MetadataManagementService.java
```diff
@@ -284,6 +286,28 @@ private void consolidatePublishers(List<String> targetValues, List<String> value
...
+            String normalized = LanguageNormalizer.normalize(metadata.getLanguage());
+            if (normalized != null && !normalized.equals(metadata.getLanguage())) {
+                metadata.setLanguage(normalized);
+                changed.add(metadata);
+            }
+        }
+
+        if (!changed.isEmpty()) {
+            bookMetadataRepository.saveAll(changed);
+            writeMetadataToFile(changed, moveFile);
```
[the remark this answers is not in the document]
ghostdiverr: I removed this endpoint

frontend/src/app/features/metadata/component/book-metadata-center/metadata-editor/metadata-editor.component.ts
```diff
@@ -400,6 +404,7 @@ export class MetadataEditorComponent implements OnInit {
       this.metadataCenterViewMode = user.userSettings.metadataCenterViewMode ?? 'route';
       this.autoSaveEnabled = user.userSettings.autoSaveMetadata ?? false;
     }
+    this.languageService.getLanguages().subscribe(langs => this.languageOptions = langs);
```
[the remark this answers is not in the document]
ghostdiverr: No more frontend features in this PR.

frontend/src/app/features/metadata/component/book-metadata-center/metadata-viewer/metadata-viewer.component.ts
```diff
@@ -442,7 +444,15 @@ export class MetadataViewerComponent implements OnInit, OnChanges, AfterViewChec
...
 
+  private languageOptions: LanguageOption[] = [];
+
+  getLanguageName(code: string): string {
+    const found = this.languageOptions.find(l => l.code === code);
+    return found ? found.name : code;
+  }
+
   ngOnInit(): void {
+    this.languageService.getLanguages().subscribe(langs => this.languageOptions = langs);
```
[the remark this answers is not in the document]
ghostdiverr: No more frontend features in this PR.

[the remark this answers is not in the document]
ghostdiverr: No more frontend features in this PR.

frontend/src/app/features/metadata/component/bulk-metadata-update/bulk-metadata-update-component.ts
```diff
@@ -123,6 +129,7 @@ export class BulkMetadataUpdateComponent implements OnInit {
   ngOnInit(): void {
     this.bookIds = this.config.data?.bookIds ?? [];
     this.books = this.bookService.getBooksByIds(this.bookIds);
+    this.languageService.getLanguages().subscribe(langs => this.languageOptions = langs);
```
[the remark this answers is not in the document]
ghostdiverr: No more frontend features in this PR.

frontend/src/app/features/metadata/component/metadata-manager/metadata-manager.component.html
```diff
@@ -182,13 +203,25 @@ <h2 class="settings-title">
...
+          <p-select
+            [(ngModel)]="mergeTarget"
+            [options]="languageOptions"
+            optionLabel="name" optionValue="code"
+            [filter]="true" filterBy="name,code"
+            [showClear]="true"
+            appendTo="body"
+            styleClass="full-width-input"
+            [placeholder]="t('singleValuePlaceholder')"/>
+        } @else {
```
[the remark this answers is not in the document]
ghostdiverr: No more frontend features in this PR.

frontend/src/app/features/metadata/component/metadata-manager/metadata-manager.component.ts
```diff
@@ -142,6 +150,10 @@ export class MetadataManagerComponent implements OnInit, OnDestroy {
   ];
 
   ngOnInit() {
+    this.languageService.getLanguages().subscribe(langs => {
+      this.languageOptions = langs;
+      this.cdRef.detectChanges();
+    });
```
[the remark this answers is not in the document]
ghostdiverr: No more frontend features in this PR.

backend/src/main/java/org/booklore/service/metadata/MetadataManagementService.java
```diff
@@ -284,6 +286,28 @@ private void consolidatePublishers(List<String> targetValues, List<String> value
...
+            log.info("Normalized {} book language values", changed.size());
```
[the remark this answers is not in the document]
ghostdiverr: I removed this endpoint

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,75 @@
...
+    public static String normalize(String input) {
+        if (input == null || input.isBlank()) {
+            return null;
+        }
+        String trimmed = input.trim();
+
+        // BCP 47 / IETF tags ("fr-FR", "fr_FR") — take primary subtag first
+        String primary = trimmed;
+        if (trimmed.contains("-") || trimmed.contains("_")) {
+            primary = trimmed.split("[-_]")[0];
```
balazs-szucs: Use pattern variable for this. 

Afterwards you can remove redundant comment also

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,75 @@
...
+        if (byCode != null) return byCode.name();
+
+        LanguageAlpha3Code byAlpha3 = LanguageAlpha3Code.getByCode(input, false);
+        if (byAlpha3 != null) {
+            LanguageCode alpha2 = byAlpha3.getAlpha2();
+            if (alpha2 != null) return alpha2.name();
+        }
+
+        List<LanguageCode> byName = LanguageCode.findByName("(?i)" + Pattern.quote(input));
+        if (!byName.isEmpty()) return byName.get(0).name();
```
balazs-szucs:
```suggestion
        if (!byName.isEmpty()) return byName.getFirst().name();
```

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    private static final class LocalizedNameMap {
+        static final Map<String, String> MAP = build();
+
+        private static Map<String, String> build() {
+            Locale[] displayLocales = {
+                Locale.FRENCH, Locale.GERMAN, Locale.ITALIAN,
+                new Locale("es"), new Locale("pt"), new Locale("nl"),
+                new Locale("ru"), new Locale("pl"), new Locale("ja")
```
balazs-szucs: Locale(java.lang.String)' is deprecated since version 19, refactor to it's "modern" equivalent, please

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,75 @@
...
+            for (LanguageCode code : LanguageCode.values()) {
+                if (code == LanguageCode.undefined) continue;
+                Locale locale = new Locale(code.name());
+                for (Locale displayLocale : displayLocales) {
+                    String displayName = locale.getDisplayLanguage(displayLocale).toLowerCase(Locale.ROOT);
+                    if (displayName.length() > 2) {
+                        map.putIfAbsent(displayName, code.name());
+                    }
+                }
+            }
```
balazs-szucs, line 69: I think this is still not loading this lazily

backend/src/test/java/org/booklore/util/LanguageNormalizerTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class LanguageNormalizerTest {
+
+    // --- null / blank ---
```
balazs-szucs: Unnecessary comments

backend/src/test/java/org/booklore/util/LanguageNormalizerTest.java
```diff
@@ -0,0 +1,99 @@
+package org.booklore.util;
+
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.CsvSource;
+
+import static org.junit.jupiter.api.Assertions.*;
+
+class LanguageNormalizerTest {
```
balazs-szucs, line 9: Most of these are testing the underlying library, no?

The current suite is essentially verifying that your lookup table is populated correctly, that does not seem very useful. Even then there are some more obvious test cases missing: e.g., numeric input, empty string etc

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
```diff
@@ -0,0 +1,75 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Locale;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+@UtilityClass
+public class LanguageNormalizer {
+
+    public static String normalize(String input) {
```
balazs-szucs:
```suggestion
    public String normalize(String input) {
```

static is redundant, i believe:` @UtilityClass` already marks methods static.

balazs-szucs: (this is for methods in this)

backend/src/main/java/org/booklore/service/metadata/parser/RanobeDbParser.java
```diff
@@ -259,7 +260,7 @@ private BookMetadata searchResultToBookMetadata(int bookId) {
                     .publisher(englishPublisher != null ? englishPublisher.getName() : null)
                     .thumbnailUrl(book.getImage() != null ? RANOBEDB_IMAGE_URL + book.getImage().getFilename() : null)
                     .description(book.getDescription())
-                    .language(englishRelease != null ? englishRelease.getLang() : book.getLang())
+                    .language(LanguageNormalizer.normalize(englishRelease != null ? englishRelease.getLang() : book.getLang()))
```
balazs-szucs, line 263: Should utility class null check be enough for this? Any particular reason we want to null check here too?

backend/src/test/java/org/booklore/util/LanguageNormalizerTest.java
ghostdiverr: The tests have been slimmed down to focus only on our logic:
- Input guards: null, empty string, blank, numeric
- BCP 47 splitting: our parsing of fr-FR / fr_FR → fr
- CLDR localized names: the map we built (French, German, Spanish names → ISO codes)
- Unknown fallback: our lowercase-and-return behaviour

Tests that were only verifying nv-i18n internals (ISO 639-1 passthrough, ISO 639-2 mapping, full English names) have been removed.

balazs-szucs: Thanks! I'll take a look shortly, from this point i think if the comments are addressed there shouldn't be much hiccups, but i want to manually test 1 more time. (Just in-case)

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
ghostdiverr: Done, replaced all new Locale(String) calls with Locale.of(String)

backend/src/main/java/org/booklore/service/metadata/parser/RanobeDbParser.java
ghostdiverr: `englishRelease` is assigned via `.orElse(null)`. It's legitimately null when no English release exists. 

The ternary selects which object to read the language from to avoid a NPE. It's not a null-guard on the language string itself. 

`normalize()`'s null check handles the case where getLang() returns null, but can't protect us against calling a method on null.

backend/src/main/java/org/booklore/util/LanguageNormalizer.java
ghostdiverr: Maybe I'm missing something, but since `LocalizedNameMap` is a private inner class and only referenced at the very end of `resolveCode()`, it won't be loaded (and `build()` won't run) until that fallback line is actually reached.

That said, did you mean we should compute the values on-demand per locale instead of pre-building the map on the first access?

backend/src/main/java/org/booklore/service/metadata/parser/RanobeDbParser.java
balazs-szucs: Okey

balazs-szucs (approved):
Thanks, works well.

PR #1413, opened 2026-05-20
Stable mod

podalanga:
## Manual Testing Steps

1.
2.
3.

## Checklist

- [ ] I have disclosed any AI usage as per the organization AI Policy above.
- [ ] I understand all of my submitted changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -85,7 +85,7 @@ ENV JAVA_TOOL_OPTIONS="-XX:+UseShenandoahGC \
     --enable-native-access=ALL-UNNAMED \
     --enable-preview"
 
-RUN apk add --no-cache su-exec libstdc++ libgcc libarchive && \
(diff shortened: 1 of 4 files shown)


balazs-szucs (commented):
1. This is (at least for now) outside purview
2. We do not want to add binaries because it breaks user deployments that are non-docker (so only when it is absolutely necessary)
3. We particularly do not want to add qpdf because it already pulled in via org.grimmory.pdfium4j.natives, so we would be downloading this twice
4. The changes themselves, are also somewhat questionable.
5. Reader vs qpdf changes seem to be unrelated.

balazs-szucs:
Feel free to start a GH discussion if you think this is worth talking about. That said, I do not think this PR solves this issue in way that would be appropriate for this codebase, so I do not see this going anywhere good. Apologies.

PR #1489, opened 2026-05-25
fix(file-imports): Properly group multi-file audiobooks imported via the folder watcher

### LINKED ISSUE #1007: Watch folder imports split multi-file audiobooks into separate books in Book Per File mode
### What went wrong?

When a library is configured with `Organization Mode: Book Per File` and `Watch folders` is enabled, adding a chapter-based audiobook folder causes Grimmory to import each `.mp3` track as its own separate book entry.

This does not match the documented behavior for organization modes. The docs say that chaptered audiobooks with 2 or more audio files in a folder should still be grouped into a single audiobook before organization mode is applied.

From investigating the code, this appears to be a mismatch between two ingestion paths:

- The full scan/rescan path collapses multi-file audiobook folders into a single folder-based audiobook before grouping.
- The watch-folder event path does not do that in `Book Per File` mode, and instead processes every audio file individually.

As a result, behavior depends on how the files are discovered:
- library scan/rescan path: grouped correctly
- watched folder create path: split incorrectly

This makes watched imports inconsistent with both the docs and the non-watcher scan behavior.

### How can we reproduce it?

1. Create or use a library with these settings:
   - `Watch folders`: enabled
   - `Metadata source`: `Embedded only`
   - `Organization mode`: `Book Per File`
   - all formats allowed

2. Use a library path that contains a chapter-based audiobook folder with multiple `.mp3` files inside a single book folder, for example:

```text
James Comey/
└── A Higher Loyalty/
    ├── A Higher Loyalty - 1 - Opening Credits.mp3
    ├── A Higher Loyalty - 2 - Dedication.mp3
    ├── A Higher Loyalty - 3 - Preface.mp3
    ├── ...
    └── A Higher Loyalty - 21 - End Credits.mp3
```

3. Let Grimmory detect the folder via the watch-folder flow, or copy/move that folder into the watched library location while Grimmory is running.

4. Open the library after the import finishes.

5. Observe that each `.mp3` file is imported as its own separate book entry instead of a single audiobook.

Expected result:
- the entire `A Higher Loyalty` folder should become one audiobook entry

Actual result:
- each chapter/track is created as its own book

Additional note:
- this seems specific to the watcher ingestion path
- the code investigation suggests a normal full scan/rescan path may behave differently and collapse the folder correctly

Nightly

### Your setup

- Grimmory version 3.0.0
- Running via Docker on macOS 26
- Accessed via Firefox

From code investigation, the likely cause is:
- watch-folder create handling in `Book Per File` mode processes files individually
- per-file creation logic in `Book Per File` mode explicitly avoids attaching/grouping files into an existing book

### Before submitting

zachyale:
## Description

`Organization Mode: Book Per File` and `Watch folders` is enabled, adding a chapter-based audiobook folder causes Grimmory to import each `.mp3` track as its own separate book entry.

This PR fixes this behaviour by moving the audiobook handling before the behaviour for `BOOK_PER_FILE` libraries is executed.

## Linked Issue

Fixes https://github.com/grimmory-tools/grimmory/issues/1007

## Changes

- Move audiobook handling before `BOOK_PER_FILE` handling
- Move audiobook handling into a cleaner helper function
- Improved test coverage for red/green tests for these cases

## Manual Testing Steps

As per the issue;

1. Create or use a library with these settings:
   - `Watch folders`: enabled
   - `Metadata source`: `Embedded only`
   - `Organization mode`: `Book Per File`
   - all formats allowed
2. Use a library path that contains a chapter-based audiobook folder with multiple `.mp3` files inside a single book folder, for example:
```text
Sample Author/
└── Sample Audiobook/
    ├── Sample Audiobook - 1 - Opening Credits.mp3
    ├── Sample Audiobook - 2 - Dedication.mp3
    ├── ...
    └── Sample Audiobook - 21 - End Credits.mp3
```

3. Let Grimmory detect the folder via the watch-folder flow, or copy/move that folder into the watched library location while Grimmory is running.
4. Open the library after the import finishes.
5. Observe that each `.mp3` file is imported as its own separate book entry instead of a single audiobook.

## Screenshots (Optional)

N/A

N/A

## AI Disclosure

Codex was used to generate improved spec coverage

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved folder-based audiobook handling to better detect and process folders containing multiple audio files as single audiobooks

* **Tests**
  * Added test coverage for audiobook subfolder detection and collapsing behavior
  * Verified folder-based audiobook processing across multiple library organization modes

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1489?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/backend/src/main/java/org/booklore/service/watcher/LibraryFileEventProcessor.java
+++ b/backend/src/main/java/org/booklore/service/watcher/LibraryFileEventProcessor.java
@@ -251,6 +251,12 @@ public class LibraryFileEventProcessor implements SmartLifecycle {
         var mode = library.getOrganizationMode() != null
                 ? library.getOrganizationMode() : LibraryOrganizationMode.AUTO_DETECT;
 
+        FolderAnalysis analysis = analyzeFolderForAudiobook(folderPath);
+        if (analysis.isFolderBasedAudiobook()) {
+            processFolderAudiobook(library, folderPath, analysis);
+            return;
+        }
+
         if (mode == LibraryOrganizationMode.BOOK_PER_FILE) {
             processFilesInFolderIndividually(library, folderPath);
             return;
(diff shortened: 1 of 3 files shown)


backend/src/main/java/org/booklore/service/watcher/LibraryFileEventProcessor.java
```diff
@@ -251,6 +251,12 @@ private void handleFolderCreate(LibraryEntity library, Path folderPath) {
         var mode = library.getOrganizationMode() != null
                 ? library.getOrganizationMode() : LibraryOrganizationMode.AUTO_DETECT;
 
+        FolderAnalysis analysis = analyzeFolderForAudiobook(folderPath);
+        if (analysis.isFolderBasedAudiobook()) {
+            processFolderAudiobook(library, folderPath, analysis);
+            return;
+        }
```
imnotjames, line 258: So we're going to look for this auto-detect no matter what?  When before we were only doing it if the value was auto-detect.

imnotjames: Per OOB comms:

Documentation says that file / folder mode doesn't really affect this because it's a "book" per folder / file, and the "book" is the group of files.

imnotjames (approved):

PR #1534, opened 2026-05-27
fix(auth): centralize token handling in auth service

### LINKED ISSUE #1480: Users are regularly logged out
### What went wrong?

After some period of time, users are getting logged out and require re-authentication.

### How can we reproduce it?

1. Log in
2. Wait a while
3. Try to do anything in the app

Nightly

### Your setup

Grimmory nightly
Firefox 149.0.2 (64-bit)

_No response_

### Before submitting

zachyale:
## Description

Fixes a race condition for handling expired tokens, where route guards could log users out before the interceptor had a chance to handle the session. Achieves this by refactoring auth service to be the source of truth for token related behaviour via `ensureAccessToken`. 

## Linked Issue

Fixes https://github.com/grimmory-tools/grimmory/issues/1480

## Changes

- Refactored access token refresh logic and validation into `AuthService` (`ensureAccessToken(...)`), and refactored the auth guard + interceptor to use ensure access token
- Added fail closed handling for incomplete auth responses
- Moved shell route auth to `canActivateChild` so guards are consistently run on all routes
- Updated the auth service, guard, interceptor, and route tests

## Manual Testing Steps

1. Reset access tokens to 60 seconds, start the app + log in
2. After token has expired, navigate to a protected route, confirm the app stays logged in
3. After confirming navigation works, repeat on an auth-protected action (for ex., saving changed metadata)
4. Verify in the network tab the first request could be a 401, but then refresh runs, and original request is retried w/ the new token
5. With an expired access token, manually update refresh token to an invalid token, then verify on navigation/action the refresh fails, and the app logs our + redirects to the login page
6. With an expired access token, manually clear the refresh token, then verify on navigation/action the refresh fails, and the app logs our + redirects to the login page

## Screenshots (Optional)

N/A

**Note:** We moved the auth guard check to a child guard so every route is guaranteed to run the same session refresh + auth checks before loading for consistency. This wasn't necessary to fix the race condition, but made sense to do in furtherance of the goal of this PR to improve consistency around auth behaviour on the F/E. Before it was inconsistent which routes had the guard on them, this helps resolve that

## AI Disclosure

Spec generation + updates

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fewer unexpected logouts and clearer handling when token refresh fails or returns incomplete credentials.
  * Reduced duplicate token-refresh attempts during concurrent requests.

* **Improvements**
  * Centralized session/token handling for more reliable authentication and websocket initialization.
  * Route authentication now yields consistent redirects (login or change-password) and inherited protection for nested pages.

* **New Features**
  * Added child-route authentication behavior so child pages inherit the shell’s auth policy.

### DIFF
--- a/frontend/src/app/core/security/auth-interceptor.service.ts
+++ b/frontend/src/app/core/security/auth-interceptor.service.ts
@@ -1,12 +1,9 @@
 import { HttpErrorResponse, HttpEvent, HttpHandlerFn, HttpInterceptorFn, HttpRequest } from '@angular/common/http';
 import { inject } from '@angular/core';
-import { catchError, filter, switchMap, take } from 'rxjs/operators';
-import { BehaviorSubject, Observable, throwError, defer } from 'rxjs';
+import { catchError, switchMap } from 'rxjs/operators';
+import { Observable, throwError, defer } from 'rxjs';
 import { AuthService } from '../../shared/service/auth.service';
 import { API_CONFIG } from '../config/api-config';
-
-let isRefreshing = false;
-const refreshTokenSubject = new BehaviorSubject<string | null>(null);
 
 export const AuthInterceptorService: HttpInterceptorFn = (req, next: HttpHandlerFn) => {
   const authService = inject(AuthService);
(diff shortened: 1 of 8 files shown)


frontend/src/app/core/security/auth-interceptor.service.ts
```diff
@@ -42,36 +39,14 @@ export const AuthInterceptorService: HttpInterceptorFn = (req, next: HttpHandler
...
-          return throwError(() => err);
-        })
-      );
-    }
-
-    return refreshTokenSubject.pipe(
-      filter(token => token !== null),
-      take(1),
-      switchMap(token =>
+    return authService.ensureAccessToken({forceRefresh: true}).pipe(
```
imnotjames, line 42: This can't happen multiple times, can it?  Since we got rid of the `isRefreshing` check.

frontend/src/app/shared/service/auth.service.ts
```diff
@@ -78,6 +118,19 @@ export class AuthService {
     this.token.set(accessToken);
   }
 
+  private saveInternalTokenResponse(response: AccessTokenResponse, expectedRefreshToken?: string): void {
+    if (expectedRefreshToken !== undefined && expectedRefreshToken !== this.getInternalRefreshToken()) {
+      throw new Error('Refresh response belongs to a stale session');
+    }
+
+    if (!response.accessToken || !response.refreshToken) {
```
imnotjames, line 139: For our use case it's fine since it should never happen - but there are cases where an access token may be granted but the refresh token won't - which would still be valid.

I think the code here is just "based on how our auth works today, this is correct?"

Not saying it needs to change, just making sure I follow.

imnotjames (approved):

frontend/src/app/shared/service/auth.service.ts
zachyale: yep, the intent it just to reflect the auth contract today

PR #1539, opened 2026-05-27
feat(kobo): Add initial implementation for reading services proxy

black-dragon74:
## Description

This PR adds the skeleton for Kobo reading services proxy. This proxy just logs for now but can be used
to store annotations in Grimmory or to sync to Hardcover.

The reading services (and related transparent proxy for UserStorage) use a set of user specified Device IDs for auth. We cannot use token based approach as the reading services endpoint may be called in cases where we do not have an active token. User can specify multiple device IDs delimited by comma (`,`)

## Linked Issue

#669 

## Changes

- Added a new security filter to check Kobo Device ID in incoming requests to reading services endpoint.
- Added reading services controller
- Added user storage controller (omitting this leads to sync failures on Kobo), acts as a transparent proxy.
- Added DTOs for kobo annotations.
- Added migrations for allowed device IDs
- Modified KoboUserSetting repo, dto and entity for allowed device IDs.
- Modified KoboServerProxy to add `proxyToReadingServices`
- Exposed the allowed device ID nob to front-end.

## Manual Testing Steps

1. Ensure your Kobo device is connected to WiFi (required for testing)
2. Highlight a text on your Kobo device.
3. Observe the logs for:

#### A highlight is added
```log
[grimmory] 2026-05-27T20:54:46.638021000Z 2026-05-28T02:24:46.637+05:30  INFO 1 --- [booklore-api] [cat-handler-155] o.b.c.ReadingServicesController          : [ReadingServices]   

Annotation id=18228563-4770-41bb-bf37-7d582e69d916 
type=highlight 
color=#F6F3B3 
chapter=OEBPS/Text/part0010.html 
progress=0.8421052631578947
text=Why wouldn’t I use my own money for that?” I sounded quite reasonable, I thought
```

#### A note is added
```log
[grimmory] 2026-05-27T20:58:15.621270000Z 2026-05-28T02:28:15.621+05:30  INFO 1 --- [booklore-api] [cat-handler-176] o.b.c.ReadingServicesController          : [ReadingServices]   

Annotation id=92e841ae-1667-4154-9b58-5f703966f730 
type=note 
color=#F6F3B3 
chapter=OEBPS/Text/part0010.html 
progress=0.8421052631578947 
text=Why wouldn’t I use my own money for that? 
note=This is a test
```

## Screenshots (Optional)
Screenshot 2026-05-28 at 2 58 38 AM

This PR is aimed at discussing the Kobo sync implementation and modifications would be made during the course of review.

Please feel free to suggest alternative approaches. Having the support for Annotation/Highlights would be an amazing addition :)

## AI Disclosure

Claude was used to understand the codebase and generate the DTOs for annotations. Each change has been thoroughly reviewed by me.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Device-ID authentication added for Kobo Reading Services and UserStorage endpoints.
  * New /api/v3 and /api/UserStorage proxy endpoints forward request/response payloads and selected headers to Reading Services.
  * Server logs concise annotation summaries when annotations are synced while still forwarding original payloads.
  * Settings UI, API, and DB migration support configuring and persisting allowed device IDs; frontend form includes debounced save.

[Review Change Stack](https://app.coderabbit.ai/change-stack/grimmory-tools/grimmory/pull/1539?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


imnotjames:
> The reading services (and related transparent proxy for UserStorage) use a set of user specified Device IDs for auth. We cannot use token based approach as the reading services endpoint may be called in cases where we do not have an active token. User can specify multiple device IDs delimited by comma (,)

Sorry, when are you calling this when there's not a valid token that's used for auth?  This API does not exist without a token.

backend/src/main/java/org/booklore/controller/ReadingServicesController.java
```diff
@@ -0,0 +1,97 @@
...
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+import tools.jackson.databind.ObjectMapper;
+
+import java.util.List;
+
+@Slf4j
+@RequiredArgsConstructor
+@RestController
+@RequestMapping("/api/v3")
```
imnotjames, line 19: There is a set of kobo paths for this & a kobo specific controller for this.

black-dragon74:
> > The reading services (and related transparent proxy for UserStorage) use a set of user specified Device IDs for auth. We cannot use token based approach as the reading services endpoint may be called in cases where we do not have an active token. User can specify multiple device IDs delimited by comma (,)
> 
> Sorry, when are you calling this when there's not a valid token that's used for auth? This API does not exist without a token.

Unlike the `api_endpoint` which has a `token` in its path segment, the Kobo devices use only the `Host` part when making requests to the reading services endpoint. Due to that we cannot leverage the token in the manner that we do for regular endpoints.

And in cases where the device has been offline for quite a while (very common with e-Readers), the reading services endpoint might be the first one being called, without init/handshake with the API. That would lead to flaky behavior.

We need a way for the user to ID themselves, I could only think of using a Device ID as a whitelist.

backend/src/main/java/org/booklore/controller/ReadingServicesController.java
black-dragon74: Umm. Do you mean in the existing codebase? There isn't. We only had `api/{v1,v2,kobo}`. Could you please rephrase?

imnotjames: You're making a new API version and new endpoints that are going to be versioned and understood as separate from the Kobo controller.

It's not something we will be officially supporting as any kind of contract and will change as Kobo devices need it to.

Kobo endpoints live under `/api/kobo`

imnotjames:
I think this Kobo devices use oauth bearer tokens for authentication with the reading services host, right?  So if we supported intercepting oauth with kobo we'd be able to do auth that way rather than an approach that's much more difficult, right?

backend/src/main/java/org/booklore/controller/ReadingServicesController.java
black-dragon74: >It's not something we will be officially supporting as any kind of contract and will change as Kobo devices need it to.

Ah, I see. That's sad as the Kobo device cannot be configured to include URL segments (without any fw patching). It just picks up the Host from the specified `reading_services_host` and forms the complete endpoint. Say, we configure it to `grimmory/api/kobo/xxxx` it will always form the endpoint as `grimmory/api/v3/xxx`.

>It's not something we will be officially supporting

Does that mean Grimmory will not support Kobo reading services as until the kobo logic remains like the above? I will then close this PR and refactor reading services into a little sidecar project/plugin that stores the annotations and links them to Grimmory via its API.

black-dragon74:
> So if we supported intercepting oauth with kobo we'd be able to do auth that way rather than an approach that's much more difficult, right?

The issue is not authentication but identification. Reading services does not send any user identification in its payload. 

Say we manage to authenticate against Kobo API by intercepting OAuth, I don't know if Kobo implements OIDC into its OAuth implementation, leading to the same issue around identification. Also, a user's credentials might differ across Grimmory and Kobo.

We could use Cookies but this will create flaky behavior in cases where the device has been offline for an extended period of time. CWA has it and trust me, it's a different kind of pain :)

imnotjames:
> > So if we supported intercepting oauth with kobo we'd be able to do auth that way rather than an approach that's much more difficult, right?
> 
> The issue is not authentication but identification. Reading services does not send any user identification in its payload.

Both authentication and identification are sent in the authorization header with a JWT that has roughly this payload:

```
{
  "nbf": 1780431177,
  "exp": 1780434777,
  "iss": "https://auth.kobobooks.com",
  "aud": [
    "https://auth.kobobooks.com/resources",
    "profileauth",
    "public_api"
  ],
  "client_id": "nickel",
  "sub": "00000000-0000-0000-0000-000000000000",
  "auth_time": 1780294514,
  "idp": "Kobo.Profile.Auth",
  "kobo_user_id": "00000000-0000-0000-0000-000000000000",
  "scope": [
    "openid",
    "profile",
    "kobo_profile",
    "profileauth_client_authd_referral",
    "public_api_authenticated",
    "public_api_anonymous",
    "offline_access"
  ]
}
```

This is retrieved by the Kobo devices through the oauth host.

I've been investigating how - if at all - we could use this to perform identification / verification in a way that does not require an extra auth mechanism.

> Say we manage to authenticate against Kobo API by intercepting OAuth, I don't know if Kobo implements OIDC into its OAuth implementation, leading to the same issue around identification. Also, a user's credentials might differ across Grimmory and Kobo.
> 
> We could use Cookies but this will create flaky behavior in cases where the device has been offline for an extended period of time. CWA has it and trust me, it's a different kind of pain :)

Kobo does implement OIDC, I didn't think Kobo's HTTP client supported cookies but I guess that's possible.

imnotjames:
Given the above - that we don't want to introduce another authentication mechanism - I'll be closing this.

PR #1661, opened 2026-06-06
chore(helm): add extra env to deployment

geneliu:
## Description
Small Helm chart fix: expose extraEnv for optional container environment variables, and render startupProbe from values (it was defined but not applied).

## Linked Issue
n/a
Need a way to set runtime UID/GID (e.g. for NFS/hostPath book mounts) without forking the chart. extraEnv follows the same pattern as extraVolumes / extraVolumeMounts.

startupProbe was already in values.yaml but missing from the deployment template, so overrides had no effect.

## Changes
values.yaml

Add extraEnv: [] with a commented example for USER_ID, GROUP_ID, and APP_USER.
No chart-level defaults for those vars; the image entrypoint already falls back its defaults.

deployment.yaml

Render .Values.extraEnv after the built-in DATABASE_* env vars.
Render .Values.startupProbe, matching the existing liveness/readiness probe pattern.

## Manual Testing Steps

helm template with defaults: startupProbe present; no extra env vars.

helm template with extraEnv set: additional env entries appear on the app container.

helm template with startupProbe overridden: deployment reflects the change.

helm install for a new installation

helm upgrade upon existing installation

## AI Disclosure

None

## Checklist

- [x ] This PR is a single focused change.

## Summary by CodeRabbit

* **New Features**
  * Added ability to configure additional environment variables for the application container (includes commented example name/value entries).

### DIFF
--- a/deploy/helm/grimmory/templates/deployment.yaml
+++ b/deploy/helm/grimmory/templates/deployment.yaml
@@ -133,6 +133,13 @@ spec:
               {{- else }}
               value: {{ include "grimmory.dbName" . | quote }}
               {{- end }}
+            {{- with .Values.extraEnv }}
+            {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- with .Values.startupProbe }}
+          startupProbe:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
           {{- with .Values.livenessProbe }}
           livenessProbe:
             {{- toYaml . | nindent 12 }}
--- a/deploy/helm/grimmory/templates/deployment.yaml
+++ b/deploy/helm/grimmory/templates/deployment.yaml
@@ -136,10 +136,6 @@ spec:
             {{- with .Values.extraEnv }}
             {{- toYaml . | nindent 12 }}
             {{- end }}
-          {{- with .Values.startupProbe }}
-          startupProbe:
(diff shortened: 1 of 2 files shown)


imnotjames:
Can you split the startup probe change to its own PR?  That can be applied immediately.

@thibaultamartin what do you think of extra env?

thibaultamartin:
The extraEnv is a pattern I've already seen used in the wild and that can be useful.

I don't see a use case for Grimmory off the top of my head. Maybe to build a very specific db connection string?

Like extraObject, it's an additional footgun for admin so it must be documented as a last resort, but it doesn't strike me as something particularly exotic.

I would consider both extraEnv and extraObject as "off the beaten path": you can use it if you know what you're doing, but the chart won't add safeguards for you.

geneliu:
@thibaultamartin @imnotjames 

The example env settings in the default values.yaml explains the use cases exactly. (extraVolume like hostPath owned by a specific uid/gid, the uid/gid need to be aligned. it align with what docker-compse does).

It will help extra securityContext settings too if needed.

imnotjames:
I'm not particularly against it.

Once this is split I can review and merge it in.

zachyale (approved):

imnotjames:
Thanks for your contribution!

Issue #1751, opened 2026-06-15
PDF Reader causes service worker to crash


imnotjames:
> Failed to load ‘https://example.com/api/v1/books/1560/content’. A ServiceWorker intercepted the request and encountered an unexpected error.

When debugging further:

> TypeError: Response.clone: Body has already been consumed.

This causes the service worker to fail and is often caused by not awaiting `fetch` calls.

When this happens in the PDF reader we see a failure of the PDF content.

imnotjames:
https://github.com/angular/angular/blob/40f7fbdccbfadd92e02623f5e5f3933d6772a68b/packages/service-worker/worker/src/assets.ts#L172

Related code?

imnotjames:
I think this is still an issue, right?

alexhb1:
Nevermind, just going through some web reader issues and couldn't reproduce initially. I think i've caught it happening again though

imnotjames:
> Nevermind, just going through some web reader issues and couldn't reproduce initially. I think i've caught it happening again though

Yeah it's really hard to replicate.  I only managed to make it happen once. 😬

alexhb1:
Nevermind i've found it, it's this: https://github.com/grimmory-tools/grimmory/blob/e6ac37f03f65ffeb42d97ac6f82628237fd221bf/frontend/ngsw-config.json#L39-L45

Remove this api-passthrough block and no more crash. I don't see the point in keeping it, it doesn't actually do what it says and angular is still doing its caching regardless. Any thoughts? 

https://angular.dev/ecosystem/service-workers/config#strategy

imnotjames:
I'm fine with it, yeah.

PR #1754, opened 2026-06-15
chore(pdf-reader): drop redundant pdf in-memory caching

imnotjames:
## Description

the PDF reader's in-memory caching is redundant and may be causing issues with the service worker's caching operation / cache-storage service.  given we would be pulling data from an in-memory blob either way, it seems unlikely that this provides any perceivable performance gain after testing with multiple workflows / network configurations

## Linked Issue

related to #1751

## Changes

drops memoized PDF blob cache

## Manual Testing Steps

1. open PDF reader in application
2. open storage console
3. clear cached values
4. open network console
5. disable cache
6. set simulated network to 2G
7. reload PDF reader

It's possible that the `fetch` not being awaited caused an issue with the response but it's unclear if that's the case here?  It's hard to replicate and reproduce the failure scenario because it's a race condition.

## AI Disclosure

None.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Refactored PDF document viewer initialization and caching mechanism to optimize memory usage and streamline cache refresh operations after document updates.

* **Tests**
  * Updated unit tests to reflect changes in the PDF viewer caching behavior.


imnotjames:
smoke test failure is related to jitpack outage

EDIT: all good now

zachyale (approved):

PR #1795, opened 2026-06-18
fix(metadata): require Google Books API key

### LINKED ISSUE #1435: Enforce using the Google Books API Key
Google Books has implemented a much more restrictive (eg, 429 on first request) rate limit.  This is either a bug or generally the way things are now and given it's been this way for a few weeks seems to be the latter.

We should update the UI to enforce an API key to enable the metadata.  An extra credit would be to devise a way that saving a newly-enabled-with-no-key-or-newly-empty Google books metadata integration is possible but requires checking a box that affirms it will likely not work.

terminalchai:
## Description

Require a non-empty Google Books API key before the metadata provider can be enabled. The key input remains available while the provider is disabled, clearing the key turns the provider off, and the save payload defensively prevents persisting `enabled: true` without a key.

## Linked Issue

Fixes #1435

## Changes

- Disable the Google Books toggle until a trimmed API key is present.
- Keep the API key input visible while the provider is disabled.
- Turn Google Books off when its key is cleared.
- Prevent invalid enabled state during settings hydration and persistence.
- Replace the skipped component spec with five focused behavior tests.

## Manual Testing Steps

1. Ran the focused component suite: `pnpm exec ng test --watch=false --include=src/app/features/settings/global-preferences/metadata-provider-settings/metadata-provider-settings.component.spec.ts` (`5/5` passed).
2. Ran `pnpm run typecheck`, `pnpm run lint`, and `pnpm run build`; all passed.
3. Ran the full frontend suite with a larger Node heap: `1624` tests passed and one unrelated `book-dialog-helper.service.spec.ts` test timed out; that spec passed `3/3` when rerun independently.

The local environment uses Node `22.18.0`, while the repository declares Node `24+`. The optional confirmation-checkbox behavior from the issue was not included; this PR implements the core requirement that an API key is necessary to enable Google Books.

## AI Disclosure

OpenAI Codex - substantially assisted with codebase investigation, implementation, test drafting, and review follow-up. I reviewed the changes and verified them locally using the commands above.

## Checklist

## New Features
- Google Books metadata provider toggle is now disabled until a valid Google API key is configured.
- The Google API key field label now indicates the key is required (not optional).

## Bug Fixes
- Google Books is automatically disabled when the API key is cleared or contains only whitespace.
- Saving settings no longer persists Google Books as enabled unless a valid API key is provided.

## Tests
- Added/updated component tests to verify Google Books enablement logic and settings persistence behavior.


imnotjames:
Can you match the description to the template?  Thanks

terminalchai:
Updated the PR description to match the repository template, including manual testing steps, checklist state, AI disclosure, and clarification that the optional confirmation-checkbox behavior is deferred. I also added the positive Google Books hydration test requested in review; the focused suite now passes 5/5.

frontend/src/i18n/en/settings-metadata.json
```diff
@@ -56,8 +56,8 @@
     "cookiePlaceholder": "Paste your Amazon cookie",
     "languageRestriction": "Language Restriction",
     "selectLanguage": "Select Language",
-    "apiKeyOptional": "API Key (Optional)",
-    "googleApiDesc": "Providing a Google Books API key increases rate limits and improves reliability.",
+    "apiKeyRequired": "API Key",
```
imnotjames: Maybe could be a good idea to just make this `googleApiLabel` so it's clearer / more consistent with the `googleApiDesc`?

frontend/src/app/features/settings/global-preferences/metadata-provider-settings/metadata-provider-settings.component.ts
```diff
@@ -141,6 +141,17 @@ export class MetadataProviderSettingsComponent {
     this.comicvineToken = newToken;
   }
 
+  onGoogleApiKeyChange(newApiKey: string): void {
+    this.googleApiKey = newApiKey;
+    if (!this.googleApiKeyConfigured) {
+      this.googleEnabled = false;
+    }
+  }
+
```
imnotjames: You're both setting this to `false` and also checking this new getter directly.  Perhaps choose one or the other?

I would drop this `on` event?

imnotjames (commented):
Is there a nicer way to mark the input as required?  When testing this it's kind of confusing to use.

terminalchai:
Thanks for the review. Addressed this in 6dc385dc7 by renaming the label key to googleApiLabel and removing the extra Google API key change handler, so the component now relies on the existing configured-key guard during hydration/save. Focused spec passes locally: pnpm --filter grimmory exec vitest run src/app/features/settings/global-preferences/metadata-provider-settings/metadata-provider-settings.component.spec.ts.

imnotjames (approved):
Works well enough for now.

imnotjames:
Thanks for the contribution!

PR #1800, opened 2026-06-19
refactor(metadata): improve hardcover edition selection

### LINKED ISSUE #1284: Improve Matching for Hardcover books
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1135

<sup>Originally posted by **sabrina553** May  5, 2026</sup>
### What went wrong?

auto-fetch metadata will find an isbn that can be highly inaccurate, and continue to fill out inaccuracies based on this. 

> correct book, wrong edition

For example: Project hail mary finds the finish audiobook: https://hardcover.app/books/project-hail-mary/editions/30855906

### How can we reproduce it?

1. hardcover as a P1 provider
2. autofetching with title+author only

Stable

### Your setup

- Grimmory version: 3.0.3
- Docker

hardcover1
hardcover2

### Before submitting

sabrina553:
## Description
Improve overall functionality of the hardcover parser. For faster and more consistent, reliable results, and fixes issue with 
hardcover reading tracker

## Linked Issue

Fixes #1284 

## Changes

- new searchbyID uses lists for multiple isbns/hcid's, retrives metadata score+reading format, reduce api calls
- improved booksearch query, uses weights, filters, fields etc to improve initial search for title/author only
- introduce levenshtiendistance calculations for titles/alternative-titles/author names to further filter, search results, and align best-results, including for non-english titles. 
- further filtering of editions based on PRIMARY file format (audiobook/not audiobook) , and users locale.default for languages. format filter will improve compatibility with hardcover reading tracker.. where audiobook isbn's will not correlate to pages read on hc website.
- a general restructuring of the relevant code to improve readability and where relevant introduced new tests, updated, or removed if otherwise no longer relevant

## Manual Testing Steps

1. Manually testing each test where updates were needed
2. Testing a sample library of ~200 books by title only, with a success rate of 99.5%, this increases to 100% when author is included also

This is a *good* start in regards to the general improvements here and the feasibility of using hardcover as  p1 provider, and also introducing "guardrails." However there are still inadequacies mostly on the end of hc's data. But Specifically regarding these changes. I think extra thought should go toward the filtering editions section... 

Using the user's locale is not the be all and end all to searching languages, and their will be plenty of edge cases. Looking at manga. Where this method isn't sufficient.

Furthermore, The use of alternative titles, and matching therein, does work wholly. It isn't so straightforward after this point, to return the given title and potential foreign-language/edition specific results, like the description. There is some work to weight this up between hc md quality, and 

With the audiobook working via the primary file only it doesn't register when the audiobook is there, but not a primary file. this isn't a massive to fix, but again requires some further consideration with implementation  

## AI Disclosure

Qwen-coder: local model for checking over my code.

## Checklist

## Summary by CodeRabbit

* **Refactor**
  * Improved hardcover metadata retrieval with a prioritized flow (ISBN first, then title+author, then title-only) and distance-based matching for better result ranking.
  * Enhanced edition selection by requested file format (including audiobook handling) and locale language.
  * Improved metadata normalization and mapping consistency (including ISBN normalization) and derived ISBN-based provider lookup from ISBN-13 settings.
* **Bug Fixes**
  * Invalid release/publish dates are now omitted from returned metadata.
* **Tests**
  * Refreshed and expanded hardcover parser/search tests using deterministic, fixture-based data, including edition filtering coverage.


sabrina553:
Applied and tested rabbits fixes, mostly null handling, biggest error with filterAuthor function, fixed this by flipping the "<" to ">" rather than it's own suggestion which seemed pointlessly long-winded, had to adjust some tests to account for this change and hence returned an empty list rather than previous behaviour which was to return the initially received "hits" . Which should lead to:

try title+author -> no author matches -> try title only. 

where as prior it was just continuing with the same list of hits, rather than trying again, with a title only search 

++ 

Fixed the nitpicks in the tests. 

++

Ran one more test with the above changes. where Only searching by title, then following up with searching by, title+author for same 200 library sample..


Key takeaways: 

total time - 556189ms (2.82s/book), with the initial PR I was getting around 489000ms (2.48s/book), Though I *believe* that might be accounted for with grabbing covers aswell, None of the recent commits really should have such a significant affect on performance.

Which is on-par with the code as I inherited it, if you don't account for the second time one would fetch, where it would grab all new information, totalling about (4s/book), which makes sense, most of the time is api (1 call per second), and both methods required about 2 calls. Now the Isbn search should only require one call, and the title/author still takes 2 calls, though these calls are now more refined, as is the logic within grimmory's code. Looking at about a 15-41% speed increase.

meanwhile accuracy is vastly improved where initally i was finding myself with 13 languages and 4 variations of english for my sample, it is now all "en" as expected..

Moreover, 173(/197) of these books sit within 70-79% (good) metadata score, from title search alone
- 77% is the maximum value you can achieve with one provider and accounts for books in a seires, and 72% for standalone books. 

20 of them sit in 50-69% (fair), these fall under a few categorises
- limited hc md, being the modal, with these books mostly having a metadata score w/ hardcover of ~1600 for various reasons, we start to see HC doesn't have a cover image, or reviews, a designated language. 

Finally there are 4 books in the weak (30-49%) 
- not much to say other than poor hc data, 3 don't have a cover. however all do have data, and do resolve automatically with ISBN's or atleast with a HCID and a link to their own page. Which ultimately means down the line p2->x they would resolve if (and very likely so the book can be found with another provider)

++

On covers, I didn't want to check, all of them. as this is both tedious and not something that can *really* be helped, genuinely i think the best fix would just to let some other provider take the wheel for covers. But of 100: 1 was "ugly", 12 were of poor quality, 2 were photographs, and 3 had none. I was also being *very generous* in saying so. 

++

Further notes: 

somethings i forgot to mention, but there definitely could also be some improvements with differentiating titles and subtitles. I did provide a little bit of logic in this regard, and i think if the "alternative" titles, multiple languages comes into full fruition it would be accounted for here. But it's very much a bit of a rabbit-hole and it's best imo to just get the barebones product released before being too nitpicky.

backend/src/main/java/org/booklore/model/enums/BookCategory.java
imnotjames, line 1: Is it possible to remove the whitespace changes from `BookCategory`?

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -1,24 +1,25 @@
...
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicLong;
 
 @Slf4j
 @Service
 public class HardcoverBookSearchService {
 
     public static final int DEFAULT_PER_PAGE = 10;
-    private static final long INITIAL_DELAY_MS = 1200;
+    private static final long INITIAL_DELAY_MS = 1000;
```
imnotjames, line 21: Why 1200 -> 1000 here?

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -34,17 +35,29 @@ public HardcoverBookSearchService(AppSettingService appSettingService) {
                 .build();
     }
 
-    public List<GraphQLResponse.BookWithEditions> searchBookByIsbn(String isbn) {
+    public List<GraphQLResponse.BookWithEditions> searchBookByIsbn(List<String> isbn) {
+        List<Integer> empty = new ArrayList<>();
+        empty.add(0);
+        return searchBookById(isbn, empty);
```
imnotjames: List can use `List.of(0)` - but why not just an empty array via `List.of()`

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -117,8 +146,20 @@ public List<GraphQLResponse.Hit> searchBooks(String query, int perPage) {
...
+                filter_by: $filterBy,
+                weights: $weights,
+                fields: $fields,
+                sort: "users_count:desc",
+                typos: "2",
+                query_type: "Book",
+                per_page: $limit,
+                page: 1)
+                
+                {results }}""");
```
imnotjames, line 153: Can this be formatted to be easier to read?

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -104,11 +119,25 @@ query BookSearchByIsbn($isbn: String!) {
...
+    }
+
+    public List<GraphQLResponse.Hit> searchBooks(String title, String author) {
+        author = title + " "  + author;
+        return searchBooks(author, "", "5,3,10,5", "title, alternative_titles, author_names, description", DEFAULT_PER_PAGE);
+    }
+
+    public List<GraphQLResponse.Hit> searchBooks(String title, String author, int limit) {
+        author = title + " "  + author;
+        return searchBooks(author, "", "5,3,10,5", "title, alternative_titles, author_names, description", limit);
```
imnotjames: This may be clearer if you pass the value directly in instead of reassigning onto the `author`

```suggestion
        return searchBooks(title + " "  + author, "", "5,3,10,5", "title, alternative_titles, author_names, description", limit);
```

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -104,11 +119,25 @@ query BookSearchByIsbn($isbn: String!) {
...
+    }
+
+    public List<GraphQLResponse.Hit> searchBooks(String title, int limit) {
+        return searchBooks(title, "alternative_titles: " + title, "10,5,3", "title, alternative_titles, description", limit);
+    }
+
+    public List<GraphQLResponse.Hit> searchBooks(String title, String author) {
+        author = title + " "  + author;
+        return searchBooks(author, "", "5,3,10,5", "title, alternative_titles, author_names, description", DEFAULT_PER_PAGE);
+    }
```
imnotjames, line 127:
```suggestion
    public List<GraphQLResponse.Hit> searchBooks(String title, String author) {
        return searchBooks(title, author, DEFAULT_PER_PAGE);
    }
```

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
```diff
@@ -104,11 +119,25 @@ query BookSearchByIsbn($isbn: String!) {
         return response.getData().getBooks();
     }
 
-    public List<GraphQLResponse.Hit> searchBooks(String query) {
-        return searchBooks(query, DEFAULT_PER_PAGE);
+    public List<GraphQLResponse.Hit> searchBooks(String title) {
+        return searchBooks(title, "alternative_titles: " + title, "10,5,3", "title, alternative_titles, description", DEFAULT_PER_PAGE);
+    }
```
imnotjames, line 119:
```suggestion
    public List<GraphQLResponse.Hit> searchBooks(String title) {
        return searchBooks(title, DEFAULT_PER_PAGE);
    }
```

backend/src/main/java/org/booklore/service/metadata/MetadataRefreshService.java
imnotjames, line 1: This change is probably not required

backend/src/main/java/org/booklore/service/metadata/BookMetadataService.java
```diff
@@ -136,10 +134,11 @@ private List<MetadataProvider> deriveProviderChainFromSettings() {
...
             if (options != null && options.getFieldOptions() != null) {
                 MetadataRefreshOptions.FieldProvider titleProvider = options.getFieldOptions().getTitle();
-                if (titleProvider != null) {
+                MetadataRefreshOptions.FieldProvider isbnProvider = options.getFieldOptions().getIsbn13();
+                if (isbnProvider != null) {
                     List<MetadataProvider> chain = Stream.of(
-                                    titleProvider.getP1(), titleProvider.getP2(),
-                                    titleProvider.getP3(), titleProvider.getP4())
+                                    isbnProvider.getP1(), isbnProvider.getP2(),
+                                    isbnProvider.getP3(), isbnProvider.getP4())
```
imnotjames: Is this change required for the hardcover side of things?  Could it be brought in separately?

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/HardcoverBookSearchService.java
sabrina553: It was my understanding this related to the rate-limiting. 

"API calls are rate-limited to 60 requests per minute."
https://docs.hardcover.app/api/getting-started/

backend/src/main/java/org/booklore/model/enums/BookCategory.java
sabrina553: File should just be deleted, no? was a remanent of something that didn't pass a preliminary check

backend/src/main/java/org/booklore/service/metadata/BookMetadataService.java
sabrina553: Required is a stretch, but it definitely is needed to get the best out of Hardcover as is, and makes sense in the grand scheme of things. happy to apply it elsewhere but is required ASAP in my opinon

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
```diff
@@ -1,440 +1,522 @@
...
+
+            BookMetadata metadata = new BookMetadata();
+            log.debug("Processing edition '{}' with id '{}' of book '{}'", book.getEditions().getFirst().getTitle(),
+                    book.getEditions().getFirst().getId(),
+                    book.getTitle());
+            mapBookToMetadata(book, book.getEditions().getFirst(), metadata);
+
+            if (metadata != null) {
+                results.add(metadata);
+            }
```
imnotjames: This can never be null.

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
```diff
@@ -1,440 +1,522 @@
...
+            if (book.getEditions() == null || book.getEditions().isEmpty()) {
+                continue;
+            }
+
+            BookMetadata metadata = new BookMetadata();
+            log.debug("Processing edition '{}' with id '{}' of book '{}'", book.getEditions().getFirst().getTitle(),
+                    book.getEditions().getFirst().getId(),
+                    book.getTitle());
+            mapBookToMetadata(book, book.getEditions().getFirst(), metadata);
+
```
imnotjames: You probably can move the book metadata creation into the `mapBookToMetadata` helper.

```
private BookMetadata toMetadata(GraphQLResponse.BookWithEditions book, GraphQLResponse.Edition edition) {
    var builder = BookMetadata.builder()
        .hardcoverId(book.getSlug());

    // etc

    return builder.build();
}
```

Then the mapping helpers just need to accept the builder class instead if you still want them all split up like that, and they can have void returns.

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
```diff
@@ -1,440 +1,522 @@
...
+        if (!searchAuthor.isBlank()) {
+            docs = filterAuthor(docs, searchAuthor);
+            log.debug("Filtered by author: {} ({} results)", searchAuthor, docs.size());
+        }
+
+        // Sort the search results in order of their calculated lev-dist (asc).
+        return docs.stream()
+                .sorted(Comparator.comparingDouble(GraphQLResponse.Document::getLevenshteinDistanceTitle))
+                .toList();
+    }
```
imnotjames, line 156: Is it possible to do this by using predicates/functions and a single stream instead of multiple array lists?  We could also avoid using the `GraphQLResponse` document holding state by creating a helper `record` in this class.

eg, a private record in the `HardcoverParser`

```java
    record ScoredDocument (
            GraphQLResponse.Document document,
            double score
    ) {}
```

We'd update the filter to provide a predicate:

```
    private Function<GraphQLResponse.Document, ScoredDocument> mapTitleLevenshtein(String searchTitle) {
        LevenshteinDistance levenshtein = LevenshteinDistance.getDefaultInstance();
        String searchTitleLower = searchTitle.toLowerCase();

        return (GraphQLResponse.Document doc) -> {
            if (doc.getTitle() == null || doc.getTitle().isBlank()) {
                return new ScoredDocument(doc, Double.MAX_VALUE);
            }
            // calculate the lev.dist for the Work-level title and the search provided term
            double totalScore = levenshtein.apply(searchTitleLower, doc.getTitle().toLowerCase());

            if (doc.getAlternativeTitles() != null) {
                //repeat the lev.dist calc for each of the alternative titles
                totalScore += doc.getAlternativeTitles()
                        .stream()
                        .map(String::toLowerCase)
                        .map(title -> levenshtein.apply(searchTitleLower, title))
                        .min(Double::compare)
                        .orElse(Integer.MAX_VALUE);
            }

            // (minTitleDist+minAltTitleDist)/1+userCount. best scores should approach 0
            int usersCount = doc.getUsersCount() == null ? 0 : doc.getUsersCount();
            return new ScoredDocument(doc, (totalScore + 1)/(usersCount + 1));
        };
    }
```

and then finally..

```
    private List<GraphQLResponse.Document> filterSearch(List<GraphQLResponse.Hit> hits, FetchMetadataRequest request) {
        if (hits == null || hits.isEmpty()) {
            log.info("Hardcover: No results found for title '{}'", request.getTitle());
            return Collections.emptyList();
        }

        String searchTitle = request.getTitle() != null ? request.getTitle() : "";
        String searchAuthor = request.getAuthor() != null ? formatRequestAuthor(request.getAuthor()) : "";

        //convert hits into document format.
        Stream<GraphQLResponse.Document> docs = hits.stream()
                .map(GraphQLResponse.Hit::getDocument)
                .filter(Objects::nonNull);

        log.debug("Filtering by title: '{}', author: '{}'", searchTitle, searchAuthor);

        if (!searchTitle.isBlank()) {
            docs = docs
                    .map(mapTitleLevenshtein(searchTitle))
                    .sorted(Comparator.comparingDouble(ScoredDocument::score))
                    .map(ScoredDocument::document);
            log.debug("Filtered by title: {}", searchTitle);
        }

        // Similar with the author - different because it uses "max" distance as part of a filter?

        return docs.toList();
}
```

Happy to chat through this more, but I would suggest similar throughout this change set to make for an easier to follow code path.

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
```diff
@@ -1,440 +1,522 @@
...
 
-        if (matchedDocs.isEmpty()) {
-            return Collections.emptyList();
+    private List<GraphQLResponse.BookWithEditions> filterEditions(List<GraphQLResponse.BookWithEditions> results, Book book) {
+        for (GraphQLResponse.BookWithEditions result : results) {
+            filterEditionsByFormat(results, result, book);
+            filterEditionsByLanguage(results, result);
         }
+        return results;
+    }
```
imnotjames, line 259: This could be more cleanly defined as a `map` over a stream rather than relying on mutations of objects we're passing into the functions, right?

backend/src/main/java/org/booklore/service/metadata/BookMetadataService.java
imnotjames: Doesn't this affect other providers rather than just hardcover, though?  Will it negatively affect providers that don't pull ISBN?

I suggest carving it out into its own PR.

imnotjames (commented):
A big part of my comments have to do with reworking the code to avoid creating methods that cause side effects by updating objects in place and instead try to use the various stream APIs.

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
sabrina553: HAPPY to move the instantiation inside, though I'm a little lost with regards to this builder class. 

I can easily change the functions within mapBookToMetadata() to use this model. except for the last function - mapEditions() where it's going over already obtained data, there is no way to read values committed to builder. ie this code 

``` java
    private BookMetadata mapEditions(GraphQLResponse.Edition edition, BookMetadata metadata) {

        if (metadata.getIsbn13() == null || metadata.getIsbn13().isBlank()) {
            mapIsbn(metadata, edition);
        }
```
this is fine with the current functionality, where it only takes one book and one edition


I can change metadata.getIsbn13 to book.getisbn -- But would also have to pass the book to the function aswell as the edition...

However I wrote it this way with the intention, that it could parse multiple editions. Like the P1->P4, but at an edition level... Which could be beneficial in certain cases.

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
sabrina553: updated this, not 100% confident ala 

``` java
                    result.setEditions(filteredByLanguage);
                    return result;
```
 is this the best practise?

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
sabrina553: Makes sense with the title, though i think 

```java 
    //convert hits into document format.
        Stream<GraphQLResponse.Document> docs = hits.stream()
                .map(GraphQLResponse.Hit::getDocument)
                .filter(Objects::nonNull);
```

works best defined as a List instead of a stream

bumping my head with the author though, that last little part:

```java

        // find the edition with the lowest dist
        final double best = docs.stream()
                .mapToDouble(GraphQLResponse.Document::getLevenshteinDistanceAuthor)
                .min()
                .orElse(Double.MAX_VALUE);

        double threshold = searchAuthor.length()*0.8;

        //Try to return the searches where the distance is no greater than aproximately the length of the search term
        return best > threshold
                ? Collections.emptyList()
                : docs.stream()
                .filter(doc -> doc.getLevenshteinDistanceAuthor() <= threshold)
                .toList();


```

is doing a lot of work, especially I need the empty list, so it will fallback on a title only search.. Rather than just returning "unfiltered" results and not quite sure how to achieve that with this 'record' form

imnotjames: I'm not sure I'm fully understanding what that's even doing?  Isn't all of that the same as..

```
        double threshold = searchAuthor.length()*0.8;

        //Try to return the searches where the distance is no greater than aproximately the length of the search term
        return docs.stream()
                .filter(doc -> doc.getLevenshteinDistanceAuthor() <= threshold)
                .toList();
```

At least, that's what it seems like it's accomplishing via testing.  It's the same result either way.  Like.. we _could_ do that via first sorting and then using a filter that tracks the "first" result?  But a big "why" pops into my head.  The performance gain is minimal, here.

Assuming we have the map to author levenshtein:

```java
    private Function<GraphQLResponse.Document, ScoredDocument> mapAuthorLevenshtein(String searchAuthor) {
        LevenshteinDistance levenshtein = LevenshteinDistance.getDefaultInstance();
        String searchAuthorLowercase = searchAuthor.toLowerCase();

        return (GraphQLResponse.Document doc) -> {
            if (doc.getAuthorNames() == null || doc.getAuthorNames().isEmpty()) {
                return new ScoredDocument(doc, Double.MAX_VALUE);
            }

            //Find the author in the list of authors with the closest name to the search term
            int minDistance = doc.getAuthorNames().stream()
                    .map(String::toLowerCase)
                    .map(author -> levenshtein.apply(searchAuthorLowercase, author))
                    .min(Integer::compare)
                    .orElse(Integer.MAX_VALUE);

            return new ScoredDocument(doc, minDistance);
        };
    }
```

So our filter search would look like:

```java

    private List<GraphQLResponse.Document> filterSearch(List<GraphQLResponse.Hit> hits, FetchMetadataRequest request) {
        if (hits == null || hits.isEmpty()) {
            log.info("Hardcover: No results found for title '{}'", request.getTitle());
            return Collections.emptyList();
        }

        String searchTitle = request.getTitle() != null ? request.getTitle() : "";
        String searchAuthor = request.getAuthor() != null ? formatRequestAuthor(request.getAuthor()) : "";

        //convert hits into document format.
        Stream<GraphQLResponse.Document> docs = hits.stream()
                .map(GraphQLResponse.Hit::getDocument)
                .filter(Objects::nonNull);

        log.debug("Filtering by title: '{}', author: '{}'", searchTitle, searchAuthor);

        if (!searchTitle.isBlank()) {
            docs = docs
                    .map(mapTitleLevenshtein(searchTitle))
                    .sorted(Comparator.comparingDouble(ScoredDocument::score))
                    .map(ScoredDocument::document);
            log.debug("Filtered by title: {}", searchTitle);
        }

        if (!searchAuthor.isBlank()) {
            docs = docs
                    .map(mapAuthorLevenshtein(searchAuthor))
                    .sorted(Comparator.comparingDouble(ScoredDocument::score))
                    .filter(doc -> doc.score <= (searchAuthor.length() * 0.8))
                    .map(ScoredDocument::document);
        }

        return docs.toList();
    }
```

Though it's possible I'm misunderstanding?

backend/src/main/java/org/booklore/service/metadata/parser/HardcoverParser.java
imnotjames: I don't think I understand the concern here.  With the current code there is no way that this code path could get called multiple times, could it?  Even if it could, you'd overwrite, so you make sure that the "priority" goes from lowest to highest.

> I can change metadata.getIsbn13 to book.getisbn -- But would also have to pass the book to the function aswell as the edition...

There's no `book.getISBN()` calls I can see?

sabrina553:
with regards to [BookMetadataService.java](https://github.com/grimmory-tools/grimmory/pull/1800/files/04bcc8fca3d30f9ab08d468f6fbc73c653385354#diff-d910ecdf79a19ea557b10e437d45f8709b7ab17f5ac68c4385df963aa1dbc739): https://github.com/orgs/grimmory-tools/discussions/1978  everything else should be updated per advice..

```
org.opentest4j.AssertionFailedError: 
expected: org.booklore.model.dto.BookMetadata@3048ec86
 but was: null
	at org.booklore.service.metadata.BookMetadataServiceTest$LookupByIsbn.skipsFailingProviderAndTriesNext(BookMetadataServiceTest.java:188)

```
this error comes up, and might be the issue with CI above, but is related to google+amazon. not code i have touched.

backend/src/test/java/org/booklore/service/metadata/BookMetadataServiceTest.java
```diff
@@ -178,7 +168,7 @@ void skipsFailingProviderAndTriesNext() {
             AppSettings settings = AppSettings.builder()
                     .defaultMetadataRefreshOptions(MetadataRefreshOptions.builder()
                             .fieldOptions(MetadataRefreshOptions.FieldOptions.builder()
-                                    .title(MetadataRefreshOptions.FieldProvider.builder()
+                                    .isbn13(MetadataRefreshOptions.FieldProvider.builder()
```
imnotjames: Looks like you didn't undo this change?

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/GraphQLResponse.java
```diff
@@ -159,6 +162,9 @@ public static class Document {
 
         @JsonProperty("users_read_count")
         private Integer usersReadCount;
+
+        private Double levenshteinDistanceTitle;
+        private Double levenshteinDistanceAuthor;
```
imnotjames: Are either of these needed?

backend/src/main/java/org/booklore/service/metadata/parser/hardcover/GraphQLResponse.java
```diff
@@ -235,6 +241,12 @@ public static class Edition {
         private String isbn13;
 
         private Language language;
+
+        @JsonProperty("reading_format_id")
+        private int readingFormatId;
+
+        @JsonProperty("score")
+        private int score;
```
imnotjames: I can't see where this `score` property is being read - is it anywhere?  Can it be removed?

backend/src/test/java/org/booklore/service/metadata/BookMetadataServiceTest.java
imnotjames: I added 6fa9cb8

imnotjames (approved):

imnotjames:
Thanks for your contribution

Issue #241, opened 2026-03-27
It's not possible to detach a book file once attached (without deleting)


piffio:
### What Went Wrong?

I attached a book format to the wrong "main" book entry, and there is no way to detach it besides deleting the file and then uploading it again.

### How Can We Reproduce This?

1. Go to the book details of a book you want to associate with another one. e.g. the recently uploaded PDF version of a book you also have in EPUB format
2. Go through the "Attach to Another Book" modal
3. Select the WRONG book because you're tired of having done this for the Nth time with plenty of books
4. Curse, and then go to the WRONG target book hoping to find a button/menu entry that will allow you to "undo" the action
5. Discover in desperation there is no such capability available, and that the only way to fix it is to delete the attached format, upload it again, and go through the whole process hoping you'll not make the same mistake again :) 

### What Should Have Happened?

I expect the interface to offer the possibility to seamlessly detach a file from a book entry, which should create a new standalone book entry for that file format.

### Screenshots or Error Messages (Optional)

<img width="200" height="122" alt="Image" src="https://github.com/user-attachments/assets/10793bb8-f273-438e-ad4e-2d4fe12d9a6f" />
<img width="204" height="47" alt="Image" src="https://github.com/user-attachments/assets/1a94be19-3fd2-4078-a01c-ef552c5f481a" />
<img width="233" height="295" alt="Image" src="https://github.com/user-attachments/assets/ba59cb2d-dcf2-4029-bba1-6884718ed14b" />

### Any Ideas on How to Fix This? (Optional)

We should add an entry in the menu allowing the user to de-merge a file from an existing book. This should create a new standalone book entry as it was prior to the "merge".

Please assign this to me :) 

BTW, I wrongly selected Bug. Sorry, can you re-label it after submission? It should be an `Enhancement`

### Your Setup

- Grimmory version: 2.3.0

### Before Submitting

- [x] I've searched existing issues and confirmed this bug hasn't been reported yet

gaetan-petit:
It's still not possible to detach a file to a new book with the fix from https://github.com/grimmory-tools/grimmory/issues/2084

```ERROR 1 --- [booklore-api] [t-handler-14839] o.b.exception.GlobalExceptionHandler
  : Unhandled exception: Detached entity passed to persist: org.booklore.model.entity.BookFileEntity

org.springframework.dao.InvalidDataAccessApiUsageException: Detached entity passed to persist:
    org.booklore.model.entity.BookFileEntity
[...]

PR #1852, opened 2026-06-26
fix: Defer cover regeneration when detaching to a new book

gaetan-petit:
## Description

Detaching a file to a new book fails.

## Linked Issue

#2084

## Changes

- Move the cover regeneration to AdditionalFileController.detachFile()
- Add tests

## Manual Testing Steps

1. Use detach option on a book file

## Screenshots (Optional)

image

image

I have a very basic Java knowledge, I really hope I'm not wasting your time.

## AI Disclosure

Claude - Asked for a review and writing tests.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Detaching a file now completes successfully even if cover regeneration fails.
  * Cover regeneration is attempted after detachment, without affecting the detachment result.
* **Tests**
  * Added controller tests covering successful detachment, forwarding of the metadata-copy option, and graceful handling when cover refresh throws.
  * Updated service tests to reflect removed cover-regeneration dependency.


imnotjames:
How does this fix #241?  I think ti's unrelated.

Can you open a new bug report describing what you're seeing?

I believe this "fix" is just hiding the actual issue. (That book files aren't actually "attached" to the new book by the time we try to regenerate)

gaetan-petit:
> How does this fix #241? I think ti's unrelated.
> 
> Can you open a new bug report describing what you're seeing?
> 
> I believe this "fix" is just hiding the actual issue. (That book files aren't actually "attached" to the new book by the time we try to regenerate)

@imnotjames Maybe I'm wrong because I don't have a full picture of the process.

When you try to detach the book it fails because it can't generate the cover as the new book doesn't exists yet.

I tried to defer the cover generation after the file is detached to a new book.

imnotjames:
> > How does this fix #241? I think ti's unrelated.
> > Can you open a new bug report describing what you're seeing?
> > I believe this "fix" is just hiding the actual issue. (That book files aren't actually "attached" to the new book by the time we try to regenerate)
> 
> @imnotjames Maybe I'm wrong because I don't have a full picture of the process.
> 
> When you try to detach the book it fails because it can't generate the cover as the new book doesn't exists yet.
> 
> I tried to defer the cover generation after the file is detached to a new book.

The new book does exist.  The file attachment doesn't exist because the changes to the file haven't been flushed, right?

Then it fails because we added a transaction (for some reason) to all cover generation code which causes it to handle the exceptions as a failure.  (This could be avoided by doing new transactions)

Issue #1849, opened 2026-06-26
[DRAFT] Books browse pagination: improve facets to achieve feature parity with the frontend sort/filtering


zachyale:
## Problem

In #1845, we improved pagination support for the book browse endpoints, but did not extend support in that epic to all existing frontend filtering and sorting logic.

## Goal

This issue is the epic that represents driving feature parity between the F/E based filter and sort logic, and the new pagination approach.
This issue will be fleshed out upon completion of #1845.

Issue #1845, opened 2026-06-26
Improve Book API Pagination: cursors, sorting, facets & search (RFC #1580)


zachyale:
## Problem

The frontend loads the entire library through an unpaginated books endpoint- large libraries return responses that could be 1GB+, and take a long time, only to sort/filter all on the FE. This epic targets improvements to the `/api/v1/books/page` endpoint, in a generic way such that it can easily be extended in the future to provide pagination to all other list endpoints within the app (that make sense).

## Goal

Implement the RFC #1580 books pagination surfaces, by breaking the initial work up into 3 stacked PRs. This first ticket represents the work necessary to get the updated endpoint to a place 

## Scope
1.  Paginate the `GET /api/v1/books/page` endpoint w/ cursor + sort + facets + search
2.  Add a new `GET /api/v1/books/facets` endpoint for facet values + live counts
3. Add a new `GET /api/v1/books/ids` endpoint (to allow retrieving a list of IDs to select over all relevant filters- for bulk actions)
 
### Constraints
- All behavioural changes to the request/response structure are additive, to enable backwards compatibility with the existing (limited) endpoint
- Same model shapes
- The new browse primitive structs should be reusable for other future paginated endpoints

## Out of scope

`book_search_data`/related sort key columns, author surname, title stripped sort, file based sorts, random sort, series collapse, etc- this issue is just an MVP, not a big bang all in one drive to feature parity w/ the FE

PR #1853, opened 2026-06-26
refactor(books): paginate /api/v1/books/page with cursor, sort, facet, search

zachyale:
## Description

In RFC https://github.com/orgs/grimmory-tools/discussions/1580, we agreed upon a new abstraction for handling list request and responses within the platform, in the goal of moving towards a pagination-first approach to content. This PR is the first PR in implementing that RFC, and is the first of 3 PRs to get us to an MVP that represents the design agreed upon in that RFC.

## Linked Issue

Implements https://github.com/grimmory-tools/grimmory/issues/1845?issue=grimmory-tools%7Cgrimmory%7C1846

## Changes

This PR changes two areas- first, it adds a new package, `org.booklore.browse`, and implements several reusable abstractions within that package to support the new pagination behaviour within books. Second, it refactors the existing book pagination endpoint (with backwards compatibility), to implement support for those new pagination request/response structures. I've played with breaking this PR up more, but ultimately, I found it's less digestable to have the abstractions in one PR without the implementation, since it's harder to approve/review unused code if you can't yet see how it is actually used. With that in mind:

**`org.booklore.browse`**:
- `BrowsePage` + `Link`; New response struct (content + page/cursor + links)
- `CursorCodec` / `CursorState` / `ParamsHash`; supports the cursor
- `LinksBuilder`; allows for building `self`/`first`/`previous`/`next` links
- `SortRegistry` / `SortParser`; sort whitelist, `-` for desc, `id` tiebreaker
- `FacetLogic`; support defining logic w/ and/or/not
- New `ApiError` codes

**`GET /api/v1/books/page`**:
- Adds optional `sort`, `facet`, `facet_logic`, `query`, `cursor`
- Backwards compatible w/ no params
- Same `Book` DTO/models response structs, cursor/links are additive
- Shared `BookFilterSpecifications.base()` for filtering, reuses `AppBookSpecification` (for now) + magic shelves
- `BookSortRegistry` for current schema sorts; unsupported keys 400
- `BookSearchSpecification` — title/series/author/ISBN/ASIN/genre/tags
- Content restrictions applied for non-admins
- Extracts `BookQueryService.findBooksPaged`
- Tests for cursor, sort, facets, search, legacy parity, scoping

## Manual Testing Steps

1. Verify w/ no params that behaviour supports legacy w/  cursor/links + nothing removed
2. Follow `next`/`previous` cursors, verify response struct/no gaps
3. Verify valid sorting w/ `?sort=title`, `?sort=-title`, `?sort=seriesName,-seriesNumber
4. Verify invalid sorting (`?sort=bogus`throws 400)
5. Verify valid facets w/ `?facet=genre:Horror`, multiple facets, `?facet_logic=or`
6. Verify valid search queries w/ `?query=Sample`, matches title/author/series/etc
7. Verify that reusing a cursor w/ different throws a 400
8. Verify that for non-admins that library/content restrictions are honoured

## Screenshots (Optional)

N/A

This PR does not address the need for a `GET /api/v1/books/facets` endpoint, an ID select endpoint (`GET /api/v1/books/ids`), and it does not aim to achieve feature parity with these facets W.R.T existing F/E sorting/filtering options. Those will be addressed in the future via https://github.com/grimmory-tools/grimmory/issues/1849.

## AI Disclosure

All code was written by me over the past 3 weeks, specs were drafted by Codex

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Upgraded the books browse endpoint to use cursor-based pagination with a dedicated browse response format (items, pagination metadata, and `self/first/previous/next` links).
  * Added cursor-driven sorting, facet filtering, and free-text query search, preserving relevant parameters across navigation; legacy paging is wrapped into the same response shape.
* **Bug Fixes**
  * Added stricter validation and clearer `400 Bad Request` handling for invalid sort, cursor, and facet combinations, including paging edge cases.
* **Documentation**
  * Updated API behavior to reflect the new browse response and link/cursor metadata.

### DIFF
--- a/backend/src/main/java/org/booklore/browse/CursorState.java
+++ b/backend/src/main/java/org/booklore/browse/CursorState.java
@@ -1,0 +1,25 @@
+package org.booklore.browse;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+// Decoded contents of a pagination cursor (compact base64url JSON via CursorCodec)
+// Offset-with-state: absolute offset/limit, sort string, and a fingerprint of the
+// facet/query params, so a cursor with conflicting params is rejected
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record CursorState(
+        @JsonProperty("v") int version,
+        @JsonProperty("o") long offset,
+        @JsonProperty("l") int limit,
+        @JsonProperty("s") String sort,
+        @JsonProperty("f") String paramsHash
+) {
+
+    public CursorState withVersion(int newVersion) {
+        return new CursorState(newVersion, offset, limit, sort, paramsHash);
+    }
+
+    public CursorState withOffset(long newOffset) {
+        return new CursorState(version, newOffset, limit, sort, paramsHash);
+    }
+}
--- a/backend/src/main/java/org/booklore/browse/CursorState.java
+++ b/backend/src/main/java/org/booklore/browse/CursorState.java
@@ -8,18 +8,13 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 // facet/query params, so a cursor with conflicting params is rejected
 @JsonInclude(JsonInclude.Include.NON_NULL)
 public record CursorState(
-        @JsonProperty("v") int version,
         @JsonProperty("o") long offset,
         @JsonProperty("l") int limit,
         @JsonProperty("s") String sort,
         @JsonProperty("f") String paramsHash
 ) {
 
-    public CursorState withVersion(int newVersion) {
-        return new CursorState(newVersion, offset, limit, sort, paramsHash);
-    }
-
     public CursorState withOffset(long newOffset) {
-        return new CursorState(version, newOffset, limit, sort, paramsHash);
+        return new CursorState(newOffset, limit, sort, paramsHash);
     }
 }
--- a/backend/src/main/java/org/booklore/browse/SortParser.java
+++ b/backend/src/main/java/org/booklore/browse/SortParser.java
@@ -1,0 +1,48 @@
+package org.booklore.browse;
+
+import org.booklore.exception.ApiError;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Set;
+
+// Parses the sort param- comma-separated keys, each optionally '-' prefixed for descending
+// Always appends an ascending primary key so pagination is stable
+public final class SortParser {
+
+    public static final String TIEBREAKER_KEY = "id";
+
+    private SortParser() {
+    }
+
+    public static List<SortTerm> parse(String sortString, Set<String> knownKeys) {
+        List<SortTerm> terms = new ArrayList<>();
+        boolean tiebreakerCovered = false;
+
+        if (sortString != null && !sortString.isBlank()) {
+            for (String raw : sortString.split(",", -1)) {
+                String token = raw.trim();
+                boolean descending = false;
+                if (token.startsWith("-")) {
+                    descending = true;
+                    token = token.substring(1).trim();
+                }
+                if (token.isEmpty()) {
+                    throw ApiError.INVALID_SORT.createException("Empty sort key in: " + sortString);
+                }
+                if (!knownKeys.contains(token)) {
+                    throw ApiError.INVALID_SORT.createException("Unknown sort key: " + token);
+                }
+                terms.add(new SortTerm(token, descending));
+                if (token.equals(TIEBREAKER_KEY)) {
+                    tiebreakerCovered = true;
+                }
+            }
+        }
+
+        if (!tiebreakerCovered) {
+            terms.add(new SortTerm(TIEBREAKER_KEY, false));
+        }
+        return terms;
+    }
+}
--- a/backend/src/main/java/org/booklore/browse/SortParser.java
+++ b/backend/src/main/java/org/booklore/browse/SortParser.java
@@ -17,7 +17,6 @@ public final class SortParser {
 
     public static List<SortTerm> parse(String sortString, Set<String> knownKeys) {
         List<SortTerm> terms = new ArrayList<>();
-        boolean tiebreakerCovered = false;
 
         if (sortString != null && !sortString.isBlank()) {
             for (String raw : sortString.split(",", -1)) {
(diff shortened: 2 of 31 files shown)


backend/src/main/java/org/booklore/controller/BookController.java
```diff
@@ -77,12 +80,26 @@ public ResponseEntity<List<Book>> getBooks(
         return ResponseEntity.ok(bookService.getBookDTOs(withDescription, stripForListView));
     }
 
-    @Operation(summary = "Get books (paginated)", description = "Retrieve a paginated list of books. Supports sorting via 'sort' parameter (e.g. sort=metadata.title,asc).")
+    @Operation(summary = "Get books (paginated)", description = "Retrieve a paginated page of books. Supports cursor pagination plus sort, facet, facet_logic, and query parameters; with none of these the legacy page/size behavior is preserved.")
```
imnotjames:
```suggestion
    @Operation(summary = "Get books (paginated)", description = "Retrieve a page of books. Supports cursor pagination plus sort, facet, facet_logic, and query parameters; with none of these the legacy page/size behavior is preserved.")
```

backend/src/main/java/org/booklore/browse/CursorState.java
```diff
@@ -0,0 +1,25 @@
...
+// Decoded contents of a pagination cursor (compact base64url JSON via CursorCodec)
+// Offset-with-state: absolute offset/limit, sort string, and a fingerprint of the
+// facet/query params, so a cursor with conflicting params is rejected
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record CursorState(
+        @JsonProperty("v") int version,
+        @JsonProperty("o") long offset,
+        @JsonProperty("l") int limit,
+        @JsonProperty("s") String sort,
+        @JsonProperty("f") String paramsHash
```
imnotjames: Do we need to keep a hash?

The cursor is mostly just an opaque value that kinda signals to developers "do not try to create this yourself" - we don't need to actually validate it here.

The hash also doesn't use a secret value or anything, so it's mostly just validating that the value was transmitted successfully.

backend/src/main/java/org/booklore/browse/CursorState.java
```diff
@@ -0,0 +1,25 @@
...
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+// Decoded contents of a pagination cursor (compact base64url JSON via CursorCodec)
+// Offset-with-state: absolute offset/limit, sort string, and a fingerprint of the
+// facet/query params, so a cursor with conflicting params is rejected
+@JsonInclude(JsonInclude.Include.NON_NULL)
+public record CursorState(
+        @JsonProperty("v") int version,
```
imnotjames: Given the short-lived nature of cursors, do we actually need a version?

Is the thinking that it could be used for polymorphic cursor states?

backend/src/main/java/org/booklore/browse/SortParser.java
```diff
@@ -0,0 +1,48 @@
...
+                if (token.isEmpty()) {
+                    throw ApiError.INVALID_SORT.createException("Empty sort key in: " + sortString);
+                }
+                if (!knownKeys.contains(token)) {
+                    throw ApiError.INVALID_SORT.createException("Unknown sort key: " + token);
+                }
+                terms.add(new SortTerm(token, descending));
+                if (token.equals(TIEBREAKER_KEY)) {
+                    tiebreakerCovered = true;
+                }
```
imnotjames: Unless the tiebreaker as at the end of the list, is it still the tie breaker?

eg,

If I am sorting by:

* author
* id
* title

Isn't there possible inconsistency if the titles are the same?

backend/src/main/java/org/booklore/service/browse/BookFacetRegistry.java
```diff
@@ -0,0 +1,117 @@
+package org.booklore.service.browse;
+
+import org.booklore.app.specification.AppBookSpecification;
```
imnotjames, line 3: Will we eventually be moving this specification elsewhere?

zachyale: Yes, as part of the draft issue to improve filtering/sorting support (https://github.com/grimmory-tools/grimmory/issues/1849), we'll generalize this specification away from `AppBook`. As we then expand this pattern to other models, we will evaluate as it surfaces if there is a new best practice pattern for a source of truth for filtering/search logic that should be used for all entities moving forward

backend/src/main/java/org/booklore/browse/CursorState.java
zachyale: the hash is used to fail fast if the cursor is followed when query/facet params are changed, instead of using a stale offset into a different set of results. we can kill it if you want, and just ensure in our f/e spec coverage that when a user is scrolling through a set, and calls multiple next cursors, and then changes the filters, that we always ensure we don't reuse a cursor, but I thought it could be a nice fail fast in general for the API. happy to remove it if you think it's overengineering

backend/src/main/java/org/booklore/browse/CursorState.java
zachyale: fair enough, if the version is outdated it'd 400 anyways so no love lost in keeping it/removing it. I'll remove it and mark this resolved once the commit is in

backend/src/main/java/org/booklore/browse/SortParser.java
zachyale: fair- I've shuffled some stuff around to make this make more sense. I've modified the sort parser to treat id as an implicit tiebreaker (more closely aligned to the proposal within the RFC). it's treated as an illegal sort column if provided by a user, and is applied at the end of all of the sorts.

backend/src/main/java/org/booklore/browse/SortParser.java
```diff
@@ -0,0 +1,43 @@
...
+    public static final String TIEBREAKER_KEY = "id";
+
+    private SortParser() {
+    }
+
+    public static List<SortTerm> parse(String sortString, Set<String> knownKeys) {
+        List<SortTerm> terms = new ArrayList<>();
+
+        if (sortString != null && !sortString.isBlank()) {
+            for (String raw : sortString.split(",", -1)) {
```
imnotjames, line 22: I think I see why I had done the sort the way I had originally in the RFC: The original sort uses `,` to separate between the sort key and the direction.

This is a problem - the existing `sort` logic is now broken, and clients such as the koreader plugin will not operate with this new behavior.

Perhaps we can change the split to be an unused character such as `;`?  Then support either a prefix of `-` or a suffix of `,desc` as a "descending" operator /  `,asc` as not descending.  That would keep compat.

zachyale: as per our discussion we're fine with this in the interest of advancing a better pattern moving forward

imnotjames (approved):

PR #1909, opened 2026-07-02
feat(reader): add epub reader tap navigation for mobile

### LINKED ISSUE #1876: [Enhancement] EPUB reader: tap to trigger next/prev page.
### Discussed in https://github.com/orgs/grimmory-tools/discussions/615

<sup>Originally posted by **jolpadgett** March 31, 2026</sup>
### What would you like to improve?

The builtin reader (when Grimmory is run as a PWA) is strong enough to function as my primary reader, but I miss the ability to tap the right side to move to the next page and tap the left side to move back to the previous page. This feature is standard in most reader apps. 

### What's the current friction?

Currently, swiping to move pages is clunky and regularly triggers the top bar. 

### What would make this better?

Add the ability to single tap on the side margins for single page movement. 

### Anything else? (Optional)

_No response_

### Before Submitting

- [x] I've searched existing issues and confirmed this enhancement hasn't been requested yet</div>

pixelsquared:
## Description

The EPUB reader supports tapping the side margins to turn pages, but the zone-tap handler was gated to desktop only. On touch devices / the PWA every tap fell through to the center "toggle toolbars" branch, so pages could only be turned by swiping and sub-threshold swipes registered as taps and toggled the top bar instead.

This PR enables side-margin tap navigation on touch devices too: tap the left margin for the previous page, the right margin for the next page, and the center to toggle the toolbars. The behavior is controlled by a new "Tap to Turn Page" preference (default on). Disabling the option restores the previous touch behavior where taps toggle the toolbars.

## Linked Issue

Fixes #1876

## Changes

- Backend: added `Boolean tapToTurnPage` to `EbookReaderSetting` with a default of `true` in `DefaultUserSettingsProvider`. Stored in the existing JSON settings blob — no Flyway migration; existing users default to enabled.
- Reader state: track `tapToTurnPage` in `ReaderState`, read it from user settings on init (null-safe, defaults on), and expose a setter.
- Event service: gate side-zone navigation on the setting via a new `isTapToTurnEnabled` callback instead of the `isMobile` check.
- Settings UI: new "Navigation -> Tap to Turn Page" toggle in the eBook reader preferences, plus English i18n strings.
- Tests: new Vitest cases for the event service (side taps navigate when enabled; toggle toolbars when disabled) and reader state (reads/defaults/sets the flag); new backend unit test asserting the default is enabled.

## Manual Testing Steps

1. Open an EPUB in the reader on a touch device (or Chrome DevTools with touch/device emulation enabled).
2. Tap the right margin it should advance to the next page.
3. tap the left margin it should goes to the previous page.
4. Tap the center of the page -> the top/bottom toolbars toggle (unchanged).
5. Go to Settings -> Reader Preferences -> eBook -> Navigation and turn "Tap to Turn Page" off.
6. Tap the margins again -> pages no longer turn; margin taps now toggle the toolbars (previous touch behavior).
7. Turn the setting back on -> margin taps turn pages again.
8. On desktop, click the left/right margins -> pages still turn (default behavior unchanged).

## Screenshots (Optional)

image

## Additional Context

The middle tap zone (30%–70%) and the 50px swipe threshold are unchanged.
This feature defaults to "on" if we don't want to change default behavior let me know and I can change the default.

## AI Disclosure

Claude Code (Anthropic, Opus) assisted with Java code and helping me understanding the codebase.

I reviewed all changes.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added a **Tap to turn pages** toggle to eBook reader preferences.
  * The setting is saved and applied during reading, defaulting to enabled when unspecified.
* **Bug Fixes**
  * Side-zone taps now respect the selected tap-to-turn setting.
  * Disabled tap-to-turn taps emit the appropriate single-tap event instead of changing pages.
* **Tests**
  * Added coverage for preference defaults, persistence, state updates, and touch interactions.


imnotjames:
Isn't this fairly standard behavior for an e-reader?  Why would we gate this behind an option?

pixelsquared:
> Isn't this fairly standard behavior for an e-reader? Why would we gate this behind an option?

The default will be the tap to turn page the option allows users to revert to the current method.
I wanted to give an option for users to go back in case they wanted but I have no issues with removing the option and just have tap to turn page be the only method.

SoulFirePhoenix:
If im not mistaken part of the reason why swipe to turn was set, was due to epub scroll mode. I and many others had to send bug report when booklore completely changed how epubs were read, as it basically all but broke scroll mode as it became way to sensitive. If you didn't touch the direct centre of the screen it would jump.

frontend/src/app/features/readers/ebook-reader/state/reader-state.service.ts
```diff
@@ -131,6 +133,7 @@ export class ReaderStateService {
         if (settings.maxInlineSize != null) newState.maxInlineSize = settings.maxInlineSize;
         if (settings.maxBlockSize != null) newState.maxBlockSize = settings.maxBlockSize;
         if (settings.isDark != null) newState.isDark = settings.isDark;
+        if (settings.tapToTurnPage != null) newState.tapToTurnPage = settings.tapToTurnPage;
```
alexhb1:
```suggestion
        if (settings.tapToTurnPage != null) newState.tapToTurnPage = globalSettings.tapToTurnPage ?? true;
```

frontend/src/app/features/readers/ebook-reader/core/event.service.ts
alexhb1, line 440:
```suggestion
      if (touchDuration < this.LONG_HOLD_THRESHOLD_MS && Math.abs(deltaX) < 10 && deltaY < 10) {
        const target = event.target as Element | null;
        if (target?.closest('a[href]')) return;
```

I had trouble when tapping links in the page-turn zone, it'd open the link + turn the page. This should prevent the unwanted page turn.

alexhb1 (changes requested):
Sorry for the inactivity. Just been going through reader issues and fixes last couple days, overall works well, just some small tweaks.

frontend/src/app/features/readers/ebook-reader/state/reader-state.service.ts
```diff
@@ -131,6 +133,7 @@ export class ReaderStateService {
         if (settings.maxInlineSize != null) newState.maxInlineSize = settings.maxInlineSize;
         if (settings.maxBlockSize != null) newState.maxBlockSize = settings.maxBlockSize;
         if (settings.isDark != null) newState.isDark = settings.isDark;
+        if (settings.tapToTurnPage != null) newState.tapToTurnPage = globalSettings.tapToTurnPage ?? true;
```
alexhb1:
```suggestion
       newState.tapToTurnPage = globalSettings.tapToTurnPage ?? true;
```
Sorry, forgot to suggest the removal as well! This should make sure global settings should be respected.

alexhb1 (approved):
Works well, thanks

PR #1928, opened 2026-07-03
refactor(covers): consolidate cover images into app-cover

### LINKED ISSUE #1395: Move missing cover generation to frontend UI
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1371

<sup>Originally posted by **imnotjames** May 17, 2026</sup>

The current way missing covers work is that the backend will read the missing cover jpg and return it as part of the jpg response.  It would be ideal if the API returned back a value that could be read on the frontend and the frontend make a decision on what value to show.

It would be nice to theme the missing cover image & use a cached asset that already exists on the frontend to avoid transferring the file.

### Anything else?

One approach to handle this would be to emit 404 when the image does not exist, and add an `img` callback on failure that swaps the image to the fallback.

### Want to help out?

### Before submitting

alexhb1:
## Description

Consolidates all book cover image use into a new `app-cover` component, including handling of missing covers to use the frontend placeholder. 

## Linked Issue

Fixes #1395

## Changes

- New `app-cover` component
- Migrated image use for book covers to app-cover
- Updated backend to return 404 for missing covers instead of `missing-cover.jpg`
- Fixed inconsistency with book vs audiobook image caching, and changed to use `Cache-Control` on the media endpoints instead of `ImageCachingFilter`
- Removed image lazy loading on the frontend - it made the experience worse for little performance gain given that the virtual grid already manages performance fine. 

## Manual Testing Steps

Check across the UI for book covers (book and audiobook), including books with no cover on the backend which should return the frontend placeholder. 

## Screenshots (Optional)

N/A

Lightbox is preserved by using p-img like before. This works but temporary, I'll look at doing a native angular CDK overlay for the lightbox in a follow-up. 

`missing-cover.jpg` and that error handling will still exist on the backend - it's used for Kobo/OPDS etc. 

## AI Disclosure

N/A

## Checklist

## Summary by CodeRabbit

* **New Features**
  * The app now renders book covers across browsers, tables, series, notebook, reader UI, metadata views, command palette, and duplicate results using a unified cover component with consistent title/authors labeling, sizing, previews, and placeholders.

* **Bug Fixes**
  * Media cover/thumbnail HTTP caching is now handled consistently: successful image responses are cacheable, while missing images return non-cacheable responses to avoid stale/broken visuals.
  * Cover rendering is more resilient when images are missing or fail to load.

* **Tests**
  * Updated cover-related UI tests to target the unified cover component.


frontend/src/app/shared/components/cover/cover.component.ts
```diff
@@ -0,0 +1,77 @@
...
+const COVER_COLORS = [
+  '#1a1a2e', '#2d3436', '#0c3547', '#1e3d59', '#2c2c54', '#1b262c',
+  '#2B2D42', '#3D405B', '#463F3A', '#1B2838', '#2E4057', '#4A3728',
+];
+
+function hashString(str: string): number {
+  let hash = 0;
+  for (let i = 0; i < str.length; i++) {
+    hash = ((hash << 5) - hash) + str.charCodeAt(i);
+    hash = hash & hash;
```
imnotjames: What does this line actually do?  Wouldn't ANDing something with its identity do nothing?

Or is this ensuring it never gets larger than an integer?

If that's the case, I'd just OR with 0.


```suggestion
    // Constrain to an integer
    hash |= 0;
```

frontend/src/app/shared/components/cover/cover.component.ts
```diff
@@ -0,0 +1,77 @@
...
+
+const COVER_COLORS = [
+  '#1a1a2e', '#2d3436', '#0c3547', '#1e3d59', '#2c2c54', '#1b262c',
+  '#2B2D42', '#3D405B', '#463F3A', '#1B2838', '#2E4057', '#4A3728',
+];
+
+function hashString(str: string): number {
+  let hash = 0;
+  for (let i = 0; i < str.length; i++) {
+    hash = ((hash << 5) - hash) + str.charCodeAt(i);
```
imnotjames: This looks like an implementation of java's `hashCode`, but written differently, right?

https://github.com/openjdk/jdk/blob/jdk8-b74/jdk/src/share/classes/java/lang/String.java#L1451-L1463

`n << 5 - n` would be the same as multiplying by 31, but I think multiplying by the prime 31 might be a less confusing implementation.  Not sure.


```suggestion
  // Use the same algorithm as Java's hashcode
  for (const c of str) {
    hash = 31 * hash + c.charCodeAt(0);
```

At the very least the comment would make sense.

imnotjames (approved):

frontend/src/app/shared/components/cover/cover.component.ts
imnotjames: Wait - I just realized this is basically verbatim from the old implementation.

Well, still - same comment applies, I guess.

alexhb1: Yeah there’s not much changed from before here, just moved into the new component. I can take a look at it or leave as-is, up to you :)

imnotjames: We can leave it as is, I guess.

PR #2002, opened 2026-07-14
refactor(kobo): implement Grimmory kepub transformation

### LINKED ISSUE #2001: Find Alternative to Kepubify
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1995

<sup>Originally posted by **imnotjames** July 12, 2026</sup>

Use something other than Kepubify  to create kepubs

Kepubify has a few bugs + security issues.

### Anything else?

_No response_

### Want to help out?

### Before submitting

imnotjames:
## Description

because kepubify is fairly unmaintained, this PR implements logic similar to kepubify directly in Grimmory

## Linked Issue

fixes #2001 

## Changes

* drops kepubify from dockerfile
* implements kepub transformation using jsoup & xml parsing

## Manual Testing Steps

downloaded book and read in kepub
checked that progress pushed back makes sense
did an HTML diff to check for differences in how kobo spans are generated

the exact kobo IDs are different because this implementation uses a single integer for indexing

## AI Disclosure

None.

## Checklist

## Summary by CodeRabbit

- **New Features**
  - EPUB-to-Kepub conversion now runs with Kobo-compatible formatting, optional hyphenation, and cover-image metadata.
  - Conversion supports input files without requiring an `.epub` extension.
  - Added bundled FFprobe support for media processing.

- **Bug Fixes**
  - Improved EPUB resource filtering and limited transformations to eligible HTML/XHTML content.

- **Tests**
  - Added coverage for conversion, resource handling, cover detection, and HTML/XHTML transformation behavior.

### DIFF
--- a/backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
+++ b/backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
@@ -396,12 +396,13 @@ public class KepubConversionService {
         }
     }
 
-    private Resource transformOPF(Resource opfResource, String coverImage) throws IOException {
+    private Resource transformOPF(Resource opfResource, Resource cover) throws IOException {
         // TransformOPF transforms the OPF document for a KEPUB.
         try {
             var builder = SecureXmlUtils.createSecureDocumentBuilder(true);
             var opfDoc = builder.parse(opfResource.asInputStream());
 
+            String coverImage = cover == null ? null : cover.getHref();
             transformOPFCoverImage(opfDoc, coverImage);
 
             Transformer transformer = TransformerFactory.newInstance().newTransformer();
@@ -485,7 +486,7 @@ public class KepubConversionService {
         kepub.setOpfResource(
                 transformOPF(
                         original.getOpfResource(),
-                        original.getCoverImage().getHref()
+                        original.getCoverImage()
                 )
         );
 
--- a/backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
+++ b/backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
@@ -54,13 +54,11 @@ public class KepubConversionService {
     );
 
     private final EpubReader epubReader;
-    private final EpubWriter epubWriter;
     private final KepubHtmlConversionService kepubHtmlConversionService;
 
     public KepubConversionService() {
         this(
                 new EpubReader(),
-                new EpubWriter(),
                 new KepubHtmlConversionService()
         );
     }
@@ -229,7 +227,12 @@ public class KepubConversionService {
         return kepub;
     }
 
-    public void convertEpubToKepub(InputStream inputStream, OutputStream outputStream, boolean forceEnableHyphenation) throws IOException {
+    public void convertEpubToKepub(
+            InputStream inputStream,
+            OutputStream outputStream,
+            boolean forceEnableHyphenation,
+            EpubWriter epubWriter
+    ) throws IOException {
         Book originalBook = epubReader.readEpub(inputStream);
 
         Book kepubBook = convertBookToKepub(originalBook, forceEnableHyphenation);
@@ -237,12 +240,21 @@ public class KepubConversionService {
         epubWriter.write(kepubBook, outputStream);
     }
 
+    public void convertEpubToKepub(InputStream inputStream, OutputStream outputStream, boolean forceEnableHyphenation) throws IOException {
+        convertEpubToKepub(inputStream, outputStream, forceEnableHyphenation, new EpubWriter());
+    }
+
     private void convertEpubToKepub(Path inputPath, Path outputPath, boolean forceEnableHyphenation) throws IOException {
         validateInputs(inputPath);
 
         try (var inputStream = Files.newInputStream(inputPath)) {
             try (var outputStream = Files.newOutputStream(outputPath)) {
-                convertEpubToKepub(inputStream, outputStream, forceEnableHyphenation);
+                convertEpubToKepub(
+                        inputStream,
+                        outputStream,
+                        forceEnableHyphenation,
+                        new EpubWriter()
+                );
             }
         }
 
--- a/backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
+++ b/backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
@@ -3,17 +3,14 @@ package org.booklore.service.kobo;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.booklore.util.SecureXmlUtils;
-import org.jsoup.Jsoup;
-import org.jsoup.nodes.Document;
-import org.jsoup.nodes.Element;
-import org.jsoup.nodes.Entities;
-import org.jsoup.nodes.TextNode;
 import org.springframework.stereotype.Service;
 import org.grimmory.epub4j.domain.Book;
 import org.grimmory.epub4j.domain.MediaTypes;
 import org.grimmory.epub4j.domain.Resource;
 import org.grimmory.epub4j.epub.EpubReader;
 import org.grimmory.epub4j.epub.EpubWriter;
+import org.w3c.dom.Document;
+import org.w3c.dom.Element;
 import org.w3c.dom.NodeList;
 import org.xml.sax.SAXException;
 
@@ -26,53 +23,13 @@ import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Set;
-import java.util.concurrent.atomic.AtomicInteger;
-import java.util.stream.Gatherer;
 import java.util.stream.Stream;
 
 @Slf4j
 @Service
 @RequiredArgsConstructor
 public class KepubConversionService {
-    private static final String CLASSNAME_KOBO_SPAN = "koboSpan";
-    private static final String CLASSNAME_KOBO_STYLES = "kobostylehacks";
-    private static final String CLASSNAME_KOBO_HYPHENATE = "kobostylehyphenate";
-    private static final String ID_FORMAT_KOBO_SPAN = "kobo.%d";
-
     private static final String OPF_NS = "http://www.idpf.org/2007/opf";
-
-    private static final String CSS_KOBO_STYLES = """
-            div#book-inner {
-                margin-top: 0;
-                margin-bottom: 0;
-            }
-            """;
-
-    private static final String CSS_HYPHENATE = """
-            * {
-                -webkit-hyphens: auto;
-                -moz-hyphens: auto;
-                hyphens: auto;
-        
-                -webkit-hyphenate-limit-after: 3;
-                -webkit-hyphenate-limit-before: 3;
-                -webkit-hyphenate-limit-lines: 2;
-            }
-        
-            h1, h2, h3, h4, h5, h6, td {
-                -moz-hyphens: none !important;
-                -webkit-hyphens: none !important;
-                hyphens: none !important;
-            }
-            """;
-
-    private static final String CSS_NO_HYPHENATE = """
-            * {
-                -moz-hyphens: none !important;
-                -webkit-hyphens: none !important;
-                hyphens: none !important;
-            }
-            """;
 
     private final Set<String> HTML_MEDIA_TYPES = Set.of(
             "text/html",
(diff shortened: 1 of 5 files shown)


backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,539 @@
...
+                        .syntax(Document.OutputSettings.Syntax.xml)
+            );
+
+            var newResource = new Resource(
+                    contentResource.getId(),
+                    document.toString().getBytes(StandardCharsets.UTF_8),
+                    contentResource.getHref(),
+                    MediaTypes.XHTML,
+                    "UTF-8"
+            );
```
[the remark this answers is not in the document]
imnotjames: Seems fine to me.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,280 @@
...
+
+        kepub.setOpfResource(
+                transformOPF(
+                        original.getOpfResource(),
+                        original.getCoverImage()
+                )
+        );
+
+        return kepub;
+    }
```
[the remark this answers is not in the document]
imnotjames: They aren't.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
```diff
@@ -0,0 +1,304 @@
...
+                                    element.before(koboSpan);
+                                    koboSpan.appendChild(element);
+
+                                    return NodeFilter.FilterResult.SKIP_ENTIRELY;
+                                }
+                            }
+
+                            return NodeFilter.FilterResult.CONTINUE;
+                        }
+                );
```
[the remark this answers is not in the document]
imnotjames: It's fine.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,276 @@
...
+                outputPath.getFileName(),
+                Files.size(outputPath)
+        );
+    }
+
+    public File convertEpubToKepub(File epubFile, File tempDir, boolean forceEnableHyphenation) throws IOException {
+        var outputPath = Files.createTempFile(tempDir.getPath(), ".kepub.epub");
+        convertEpubToKepub(epubFile.toPath(), outputPath, forceEnableHyphenation);
+        return outputPath.toFile();
     }
```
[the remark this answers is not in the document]
imnotjames: Technically it would work.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,276 @@
...
+    private final EpubReader epubReader;
+    private final KepubHtmlConversionService kepubHtmlConversionService;
 
-        File outputFile = executeKepubifyConversion(epubFile, tempDir, kepubifyBinary, forceEnableHyphenation);
+    public KepubConversionService() {
+        this(
+                new EpubReader(),
+                new KepubHtmlConversionService()
+        );
+    }
```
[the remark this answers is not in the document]
imnotjames: I want it to fall back.

alexhb1 (commented):
I don't think we need to use EpubWriter at all here? It builds a new epub from scratch which kepubify doesn't do. 

I'm going to bed now so not had chance to write up any specifics, but I've put up a test branch i've been using to mess around and verify some things with test epubs. https://github.com/grimmory-tools/grimmory/commit/c68bbdcffbc056eae11dc09a969453a2945f5efc

Firstly includes tests for each of the issues i was facing, verifies their errors on current 2002, plus tweaks to simplify the code that doesn't use EpubWriter and passes the tests. In all, this seems to work identically to kepubify. 

Feel free to poke around if you get chance (:

imnotjames:
> I don't think we need to use EpubWriter at all here? It builds a new epub from scratch which kepubify doesn't do.

The epubwriter is how we should be crafting all epubs, isn't it?  Since that's how we craft them elsewhere within Grimmory.

It looks like the concerns are:

* `encryption.xml` is improperly handled (it's omitted for some reason)
* filenames are URL encoded
* the source filename is not maintained

The first two seem to be problems with the epub4j library not supporting these use cases.  I don't know if the upstream does or not, but after some testing this is just every epub file.

The last I'm not sure I understand why we need to retain the source filename when it's a temporary file that's meant to be read and then erased.  I can make the change, I just don't really understand where the requirement comes from.  I'm going to hold on that one as it requires extra sanitization.

> Firstly includes tests for each of the issues i was facing, verifies their errors on current 2002, plus tweaks to simplify the code that doesn't use EpubWriter and passes the tests. In all, this seems to work identically to kepubify.

I pulled the HTML tests and integrated them, along with fixing the HTML conversion to handle them.

alexhb1:
I couldn't find anywhere that epubwriter is used in grimmory before this? This would be the first. And I'm sure that kepub creation doesn't need it at all. You need the chapter markers and kobo spans and everything that the HTML conversion service does, but not generate an entirely new file at the end. 

All those test scenarios came from running the various EPUB spec test files via both kepubify and the new transformer and seeing what regressions existed. If you avoid creating a brand new epub unnecessarily, most of those issues go away. 

There are some other kepub-specific edge cases (which you've fixed), so i think the other issues I saw from kepubify (Encryption) would be solved by removing the epub writer step. 

The file name isn't temp I don't think? I guess you could argue consistency with downloading the file yourself or converting comic -> EPUB, which both preserve a file name. Either way, it came from a regression vs the old kepubify version, gone from proper filename -> random filename.

imnotjames:
> I couldn't find anywhere that epubwriter is used in grimmory before this? This would be the first. And I'm sure that kepub creation doesn't need it at all. You need the chapter markers and kobo spans and everything that the HTML conversion service does, but not generate an entirely new file at the end.

We are still generating a new file at the end, just in a different way.

> All those test scenarios came from running the various EPUB spec test files via both kepubify and the new transformer and seeing what regressions existed. If you avoid creating a brand new epub unnecessarily, most of those issues go away.
> 
> There are some other kepub-specific edge cases (which you've fixed), so i think the other issues I saw from kepubify (Encryption) would be solved by removing the epub writer step.

I'll switch it to use the archive service then instead of zip file.

> The file name isn't temp I don't think? I guess you could argue consistency with downloading the file yourself or converting comic -> EPUB, which both preserve a file name. Either way, it came from a regression vs the old kepubify version, gone from proper filename -> random filename.

As far as I can tell we create the file with kepubify, stream the file contents (not the file itself), and then immediately delete it.

The kobo device does not seem to see the file name in my testing.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,224 @@
...
-    private void validateInputs(File epubFile) {
-        if (epubFile == null || !epubFile.isFile() || !epubFile.getName().endsWith(".epub")) {
-            throw new IllegalArgumentException("Invalid EPUB file: " + epubFile);
+                        if (properties.isBlank()) {
+                            properties = "cover-image";
+                        } else {
+                            properties += " cover-image";
+                        }
+
+                        item.setAttribute("properties", properties);
```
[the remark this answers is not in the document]
imnotjames: It doesn't matter because kepub.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,224 @@
...
+    private final CoverDetectorService coverDetectorService;
+
+    private void transformHTML(Path path, boolean forceEnableHyphenation) throws IOException {
+        var transformedHtml = kepubHtmlConversionService.transform(
+                Files.readString(path),
+                forceEnableHyphenation
+        );
+
+        Files.writeString(path, transformedHtml, StandardCharsets.UTF_8);
+    }
```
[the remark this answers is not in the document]
imnotjames: I don't think this is what kepubify did.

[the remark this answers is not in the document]
imnotjames: Yeah it's always utf-8 and kepub can only be utf-8

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
```diff
@@ -0,0 +1,352 @@
...
+                        )
+                );
+    }
+
+    /**
+     * Find every image and text node, and add `kobospan` elements where
+     * appropriate - around each image, and around each sentence in the
+     * text node.
+     */
+    private void transformContentAddKoboSpans(Document document) {
```
alexhb1, line 166: These kobo spans won't be compatible with the old kepubify ones (E.g. before they'd be `kobo.X.Y`, now `kobo.Z`). So progress sync would get pretty messed up between updates

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
```diff
@@ -0,0 +1,352 @@
...
+                            .attr("type", "text/css")
+                            .text(CSS_HYPHENATE)
+            );
+        } else {
+            document.head().appendChild(
+                    document.createElement("style")
+                            .addClass(CLASSNAME_KOBO_HYPHENATE)
+                            .attr("type", "text/css")
+                            .text(CSS_NO_HYPHENATE)
+            );
```
alexhb1, line 267: The force hyphenation = off in kobo settings before meant "do nothing", now instead this would always force hyphenation or remove all hyphenation, no alternative. I think the else branch can just get deleted and allow "off" to mean no change.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
```diff
@@ -0,0 +1,352 @@
...
+            koboSpan.addClass(CLASSNAME_KOBO_SPAN);
+            return koboSpan;
+        };
+
+        for (var textNode : wrappableText) {
+            for (var sentence : getSentences(textNode.text()).toList()) {
+                var koboSpan = nextKoboSpan.get();
+                koboSpan.text(sentence);
+                textNode.before(koboSpan);
+            }
```
alexhb1, line 231:
```suggestion
        for (var textNode : wrappableText) {
            for (var sentence : getSentences(textNode.getWholeText()).toList()) {
                var koboSpan = nextKoboSpan.get();
                koboSpan.text(sentence);
                textNode.before(koboSpan);
            }
```

The text() method removes all non-breaking spaces, which looks like a regression from kepubify / current grimmory

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,224 @@
...
-        File outputFile = executeKepubifyConversion(epubFile, tempDir, kepubifyBinary, forceEnableHyphenation);
+        if (manifestList.item(0) instanceof Element manifest) {
+            NodeList itemList = manifest.getElementsByTagNameNS(OPF_NS, "item");
 
-        log.info("Successfully converted {} to {} (size: {} bytes)", epubFile.getName(), outputFile.getName(), outputFile.length());
-        return outputFile;
-    }
+            for (int i = 0; i < itemList.getLength(); i++) {
+                if (itemList.item(i) instanceof Element item) {
+                    if (coverImage.equals(item.getAttribute("href"))) {
```
alexhb1, line 95: This never seems to run for me. The cover file path provided is always in the OPF folder e.g. `OPF/cover.jpg`. This is looking for it in the root folder so never finds it.

alexhb1 (commented):
Just showing one extra example plus the comments below: 

The "world cultures and geography" example EPUB from the sample list https://idpf.github.io/epub3-samples/30/samples.html, I see a formatting regression with the new code vs kepubify before: 

Before: 

Image

After: 

Image

The original epub lines: 
```
        <section epub:type="copyright-page" id="d18656e22">
            <span epub:type="pagebreak" title="iv" id="piv"/>
            <h1 id="d20258e22-0">Senior Consultants</h1>
```

The correct kepubify conversion: 
```
        <section epub:type="copyright-page" id="d18656e22">
            <span epub:type="pagebreak" title="iv" id="piv"></span>
            <h1 id="d20258e22-0"><span class="koboSpan" id="kobo.6.1">Senior Consultants</span></h1>
```

The converted broken one: 
```
    <section epub:type="copyright-page" id="d18656e22">
     <span epub:type="pagebreak" title="iv" id="piv"> 
      <h1 id="d20258e22-0"><span id="kobo.9" class="koboSpan">Senior Consultants</span></h1> 
```

There's no `</span>` added on the pagebreak, so i guess it breaks the formatting.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
imnotjames: Then they'll be messed up: There's no real way to solve this without basically transpiling the entirety of the golang kepubify, right?  Since it uses a hacked up HTML engine and the numbers / ordering will never match up without that.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
imnotjames: Sure, that makes sense.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
imnotjames: I'll try anyway to figure out how to replicate it - maybe we can find some way of getting _close_.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
imnotjames: This isn't too bad - just needs to be updates to be relative.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
imnotjames: Maybe got it.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,231 @@
...
+
+            transformExtractedEpubHtml(tempDir, forceEnableHyphenation);
+
+            try {
+                Path opfPath = EpubContentReader.findOPFInExtractedEpub(tempDir);
+
+                String coverHref = null;
+                if (coverArchivePath != null) {
+                    Path coverPath = tempDir.resolve(coverArchivePath);
+                    coverHref = opfPath.getParent().relativize(coverPath).toString();
```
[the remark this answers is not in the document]
imnotjames: We can't run on Windows.

backend/src/test/java/org/booklore/service/kobo/KepubHtmlConversionServiceTest.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    @Test
+    void transform_ShouldHandleEmptyDocuments() {
+        String actual = service.transform(
+                "<html><body></body></html>",
+                false
+        );
+
+        assertThat(actual).doesNotContain("Remove");
```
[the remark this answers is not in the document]
imnotjames: I am working on the tests still, just let me let the CI run already.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
alexhb1: As long as we keep a good amount of accuracy then should be fine :)

Maybe more granular numbers would be good anyway, helps with improved accuracy.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
```diff
@@ -1,93 +1,231 @@
...
+        try (var files = Files.walk(path)) {
+            var allFiles = files
+                    .filter(Files::isRegularFile)
+                    .toList();
+
+            for (var file : allFiles) {
+                String mediaType = MimeDetector.detect(file);
+
+                if (HTML_MEDIA_TYPES.contains(mediaType)) {
+                    transformHTML(file, forceEnableHyphenation);
```
alexhb1: This is picking up unwanted HTML files in the epub test files for me. e.g. [the "world cultures and geography" sample](https://idpf.github.io/epub3-samples/30/samples.html), it includes a `renditionMapping.html` which isn't included in the opf manifest. Kepubify correctly ignores it but it's being accidentally converted here: 

Epubcheck shows these extra errors from it: 
```
renditionMapping.html: element "style" not allowed
renditionMapping.html: element "div" not allowed; expected "nav"
renditionMapping.html: element "div" not allowed; expected "nav"
renditionMapping.html: body incomplete; missing required "nav"
renditionMapping.html: must contain exactly one "resource-map" nav
```

Like kepubify should this check the opf manifest and only convert what is listed there?

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
alexhb1: Should we just try and avoid replicating kepubify and just go for new more granular consistent numbers? might as well aim for better accuracy and then invalidate the old maps so people get new more precise positioning.

backend/src/main/java/org/booklore/service/kobo/KepubConversionService.java
imnotjames: Probably so.  Updated.  Unlike kepubify, I am only going off the mime types that we can detect rather than trusting the manifest.

imnotjames:
After some testing, while it doesn't negatively affect the reading on the Kobo, the device _does_ see the filename and it gets written to the sqlite.  I don't really think this actually causes issues but I'll make sure the filename is retained.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
imnotjames: Aside from replicating the HTML parser in kepubify, I think this is pretty good.  The intent was `kobo.PARAGRAPH.SEGMENT` - Where paragraph is the element and segment was the sentence in the segment.  The problem is that we may parse the HTML slightly differently leading to minor variations.

As long as they're unique it should be safe?

backend/src/main/java/org/booklore/service/book/BookDownloadService.java
```diff
@@ -279,8 +301,16 @@ public void downloadKoboBook(Long bookId, HttpServletResponse response) {
...
             if (convertEpubToKepub) {
-                fileToSend = kepubConversionService.convertEpubToKepub(inputFile, tempDir.toFile(),
-                    koboSettings.isForceEnableHyphenation());
+                var kepubPath = getKepubPath(inputFile.toPath(), tempDir);
+
+                kepubConversionService.convertEpubToKepub(
+                        inputFile.toPath(),
+                        kepubPath,
+                        koboSettings.isForceEnableHyphenation()
+                );
```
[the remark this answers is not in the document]
imnotjames: Later.  There is a lot of docs changes we need to make.

alexhb1 (commented):
I think we need to give the Kepub converter format some sort of versioning and a compare against the generated span map by version. It needs to know exactly what version of this kepub was generated and only apply the same kobo span map and positioning if they match. If versions do not match, generate a new kobo span map. 

Otherwise, old kepubify generated books will get totally messed up positioning. Future adjustments to this kepub code may leave mappings all out of wack as well. Even if the epub is identical, the changes from kepubify -> this, and this -> future changes, will lead to a bunch of trouble i think.

backend/src/main/java/org/booklore/service/kobo/KepubHtmlConversionService.java
```diff
@@ -0,0 +1,340 @@
...
+                        )
+                );
+    }
+
+    /**
+     * Find every image and text node, and add `kobospan` elements where
+     * appropriate - around each image, and around each sentence in the
+     * text node.
+     */
+    private void transformContentAddKoboSpans(Document document) {
```
imnotjames, line 165: After some testing:

* Only increment element indicator once for tags of P, Ol, Ul, Table, H1, H2, H3, H4, H5, H6
* Ignore emtpy text nodes unless the text node is within a `p`.

backend/src/test/java/org/booklore/service/kobo/KepubHtmlConversionServiceTest.java
```diff
@@ -0,0 +1,256 @@
...
+
+        var koboSpans = getKoboSpans(actual);
+
+        assertThat(koboSpans.get("kobo.0.1")).isEqualTo("First");
+        assertThat(koboSpans.get("kobo.1.1")).isEqualTo("Second");
+        assertThat(koboSpans.get("kobo.1.2")).isEqualTo("Third");
+        assertThat(koboSpans.get("kobo.2.1")).isEqualTo("Fourth");
+        assertThat(koboSpans.get("kobo.3.1")).isEqualTo("Fifth");
+        assertThat(koboSpans.get("kobo.3.2")).isEqualTo("Sixth");
+    }
```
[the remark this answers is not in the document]
imnotjames: I am adding this basically just to stop coderabbit from annoying me over and over, I don't think this drives value.

backend/src/test/java/org/booklore/service/kobo/KepubHtmlConversionServiceTest.java
```diff
@@ -0,0 +1,270 @@
...
+
+        assertThat(actual).contains("class=\"kobostylehyphenate\"");
+    }
+
+    @Test
+    void transform_ShouldOmitHyphenationWhenDisable() {
+        String actual = service.transform("<html><body><p>Hello World.</p></body></html>", false);
+
+        assertThat(actual).doesNotContain("class=\"kobostylehyphenate\"");
+    }
```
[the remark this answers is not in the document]
imnotjames: uhh no

imnotjames:
> I think we need to give the Kepub converter format some sort of versioning and a compare against the generated span map by version. It needs to know exactly what version of this kepub was generated and only apply the same kobo span map and positioning if they match. If versions do not match, generate a new kobo span map.
> 
> Otherwise, old kepubify generated books will get totally messed up positioning. Future adjustments to this kepub code may leave mappings all out of wack as well. Even if the epub is identical, the changes from kepubify -> this, and this -> future changes, will lead to a bunch of trouble i think.

I would prefer to avoid that.  I will say that you have similar problems with the koreader side of things & a few other spots.  We may eventually need to, but I'm aiming for good enough here.

I made some changes to get slightly closer for the span numbering & added a few tests.

Issue #135, opened 2026-03-23
FB2 pictures import


IvanGavrilenko:
### What Went Wrong?

While importing books in fb2 found out that book cover was not imported

<img width="332" height="329" alt="Image" src="https://github.com/user-attachments/assets/87660418-2008-4c09-a4ad-71b3e2bb0800" />

### How Can We Reproduce This?

Take the enclosed files (the files needed to be renamed - the extention must be .fb2 - github does not allow to upload fb2 files) and try to import - the "bad.fb2" would not show the cover, the "good.fb2" would show

[bad.fb2.txt](https://github.com/user-attachments/files/26179704/bad.fb2.txt)
[good.fb2.txt](https://github.com/user-attachments/files/26179703/good.fb2.txt)

### What Should Have Happened?

every .fb2 file has the book cover and thus every file should be imported with cover

### Screenshots or Error Messages (Optional)

<img width="324" height="328" alt="Image" src="https://github.com/user-attachments/assets/4971e95a-a681-4ccd-9067-a44d8d60d239" />

### Any Ideas on How to Fix This? (Optional)

As fb2 files are a simple xml files - comparing "good" and "bad" files has shown that section  <binary id="cover.jpg" content-type="image/jpg"> differs - the good file has a single line for every picture including book cover, while the bad file had line feeds in the section: 

<img width="1084" height="158" alt="Image" src="https://github.com/user-attachments/assets/b6efd2d0-34da-454b-a0df-24dacbc73101" />

Just to check that the line feeds are the issue - I've corrected the "bad" file removing line feeds in the picture section and got the file:

[bad — modif.fb2.txt](https://github.com/user-attachments/files/26179878/bad.modif.fb2.txt)

that could be uploaded correctly:

<img width="235" height="166" alt="Image" src="https://github.com/user-attachments/assets/165ffb30-df3b-4673-99d0-c95e3e1f27de" />

### Your Setup

v2.3.0
Docker

### Before Submitting

- [x] I've searched existing issues and confirmed this bug hasn't been reported yet

PR #2174, opened 2026-07-30
fix(metadata): detect FB2 cover in first body image

prakunin:
## Summary
- fall back to the first image referenced from the FB2 body when explicit cover metadata is absent
- preserve cover-named binaries and `<coverpage>` references as higher-priority candidates
- decode line-wrapped base64 from converter-generated referenced images
- add regression coverage for converter-generated FB2 files, precedence, and missing binaries

## Context
Follow-up to #135.

Some PDF-to-FB2 converters omit `<title-info><coverpage>` and place the scanned cover as the first image in `<body>`. In that shape Grimmory reports that the file has no embedded cover even though the referenced JPEG binary is present.

## Validation
- rebased `upstream/develop`: `just api test-class org.booklore.service.metadata.extractor.Fb2MetadataExtractorTest` — passed
- real affected FB2 with the final commit — extracted a decodable 90,256-byte JPEG (402×552)
- runtime backend health — HTTP 200
- full local `just api check` was attempted earlier; Gradle lost its internal `in-progress-results-generic.bin` file without reporting a failing test, so the final upstream-based validation is the focused extractor suite above

## Scope
Backend only; no endpoint, UI, archive-resolution, cover-lock, or probe-persistence changes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved FB2 cover image extraction, including support for cover images referenced via local `xlink:href`.
  * Added a last-resort fallback to use the first embedded body image when no cover page image is available.
  * Maintains cover page priority and safely returns no cover when referenced image data is missing or undecodable.
* **Tests**
  * Added unit test coverage for the new cover extraction fallback and failure scenarios.

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/extractor/Fb2MetadataExtractor.java
+++ b/backend/src/main/java/org/booklore/service/metadata/extractor/Fb2MetadataExtractor.java
@@ -34,7 +34,6 @@ public class Fb2MetadataExtractor implements FileMetadataExtractor {
     private static final Pattern KEYWORD_SEPARATOR_PATTERN = Pattern.compile("[,;]");
     private static final Pattern ISBN_CLEANER_PATTERN = Pattern.compile("[^0-9Xx]");
     private static final Pattern ISO_DATE_PATTERN = Pattern.compile("\\d{4}-\\d{2}-\\d{2}");
-
     @Override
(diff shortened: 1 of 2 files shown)


imnotjames:
Please open an issue first and follow the PR template next time.  Thanks!

Issue #2031, opened 2026-07-16
New browser UI and introducing frontend pagination


alexhb1:
This will serve as a high level for the introduction of a brand new shared browser UI, its use as the new book / series / author browser that uses the new paginated endpoints, along with related work developing new components, cards, and various other edits needed to reach this milestone. 

**Broad strokes in order:**
1. Native app-menu 
2. Shelf assignment menu implementation
3. Sidebar library and shelf actions migration to `app-menu`
4. Book Query Service
5. Book Command Service
6. Cards (Book/Author/Series)
7. Shared browse layer 
8. Book Browser
9. Advanced filtering and facets

PR #2209, opened 2026-08-01
refactor(books): update both book caches from socket events

alexhb1:
## Description

Part of the browser rework (#2031). Right now backend socket updates only the old / unpaginated book cache. This adds a small adapter to keep both caches in sync while both still exist, and rewrites the socket handling. The adapter is temporary, once the rest of the FE has switched to the paginated service this can be deleted. 

## Changes

- `legacy-book-cache.ts`: updates both the old book cache and the new book query cache when the backend sends a socket update. Useful until everything has been migrated to the paginated service. 
- `book-socket.service.ts`: Rewritten for the new query layer with some enhancements, including bunching multiple socket events within 50ms into a single invalidation (avoids 200 book deletes triggering 200 separate invalidations), checking for missed socket updates and automatically invalidating the cache to ensure reactivity, and fixing `/user/queue/book-update` to handle both the full book object or an array of book IDs. 
- Additional fix in this when rewriting the book socket service: The "update recommendations" background task never actually worked on the frontend I think? The socket would fire but nothing on the FE ever received or updated it. Now, the socket service includes it cleanly. 

## Manual Testing Steps

N/A - part of the browser rework. 

## Screenshots (Optional)

N/A

N/A

## AI Disclosure

Lots of codex testing and updates to keep both sides of the cache consistent. Plus spec additions. 

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Improved real-time book updates, including metadata, cover changes, additions, and removals.
  * Book information now automatically resynchronizes after reconnecting.
  * Task progress updates are reflected more promptly in book data.
  * Recommendation results refresh when related processing tasks finish.
  * Improved consistency between book lists, details, and recommendation displays during rapid updates.

### DIFF
--- a/frontend/src/app/app.component.spec.ts
+++ b/frontend/src/app/app.component.spec.ts
@@ -1,7 +1,8 @@
 import { signal } from "@angular/core";
 import { ComponentFixture, TestBed } from "@angular/core/testing";
 import { TranslocoTestingModule } from "@jsverse/transloco";
-import { Subject } from "rxjs";
+import { BehaviorSubject, Subject } from "rxjs";
+import { RxStompState } from "@stomp/rx-stomp";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { AppComponent } from "./app.component";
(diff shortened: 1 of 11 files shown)


imnotjames (commented):
Instead of some fairly complex logic to modify the cache like this would it make more sense to clear the cache for books that need to be updated so tanstack knows it needs to refresh from the source of truth?

alexhb1:
@imnotjames Not with the legacy cache, no. That approach is the aim with the new query service and the new browser, that just invalidates what is needed or just invalidates the handful of loaded pages, which is cheap. But for now given that the current FE handles most things client side, that behaviour would need a full re-fetch of the all-books endpoint and all the performance cost that goes with it. This just maintains compatibility with the old world (i.e. fetch everything or patch the cache manually in the client, which is how it currently works) and avoids inventing any new behaviour for now. 

The intent is that existing FE works unchanged, knowing it'll be migrated properly in the future. This just keeps it all working during partial migration instead of requiring a big rewrite of everything at once.

imnotjames (approved):

imnotjames (approved):

PR #2232, opened 2026-08-03
refactor(books): shared search normalisation and debounce helpers + migrate command palette to server-side queries

alexhb1:
## Description

Part of the browser work. Creates two shared files for search term debouncing and normalising special characters in a search query, and migrated the command palette's book search to backend queries while using these. 

These shared files replicate what the client side search typically handles, and will be used for various server-side search inputs for the browser UI, shelf assignment, etc. The palette is just the first user. 

## Linked Issue

N/A - Part of #2031 

## Changes

- New `search-terms.ts` - extracted the client side search normalisation for all future server side search queries
- New `debounced-signal.ts` - Single place for all server-side search inputs for consistent debouncing
- Migrated the command palette from client to server-side book search using the new query service. 

## Manual Testing Steps

Search books in the app's main search UI, confirm all books pull through, search debouncing works correctly, no results handled correctly, and special character searches are consistent with the client side version in latest release. 

## Screenshots (Optional)

N/A

The client side search code is still used for the old browser UI, hence why the diff isn't as negative right now. That will get cleared up with the rest of the browser UI. 

## AI Disclosure

Added specs 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Command palette book searches now use debounced remote search results.
  * Search terms are normalized for more consistent matching.
  * Added clear loading, empty-results, and search-failure states.

* **Bug Fixes**
  * Stale search results are cleared when queries change or the palette closes.
  * Search errors reset appropriately when a new search begins.

* **Tests**
  * Expanded coverage for search behavior, cancellation, failures, normalization, and debouncing.

### DIFF
--- a/frontend/src/app/shared/util/search-terms.ts
+++ b/frontend/src/app/shared/util/search-terms.ts
@@ -1,0 +1,18 @@
+export const SEARCH_DEBOUNCE_MS = 200;
+
+export function normalizeRemoteSearchTerm(str: string): string {
+  return str
+    .replace(/[!@$%^&*_=|~`<>?/";']/g, '')
+    .replace(/\s+/g, ' ')
+    .trim();
+}
+
+export function normalizeLocalSearchTerm(str: string): string {
+  const folded = str.normalize('NFD').replace(/[̀-ͯ]/g, '')
+    .replace(/ø/gi, 'o')
+    .replace(/ł/gi, 'l')
+    .replace(/æ/gi, 'ae')
+    .replace(/œ/gi, 'oe')
+    .replace(/ß/g, 'ss');
+  return normalizeRemoteSearchTerm(folded).toLowerCase();
+}
--- a/frontend/src/app/shared/util/search-terms.ts
+++ b/frontend/src/app/shared/util/search-terms.ts
@@ -8,7 +8,7 @@ export function normalizeRemoteSearchTerm(str: string): string {
 }
 
 export function normalizeLocalSearchTerm(str: string): string {
-  const folded = str.normalize('NFD').replace(/[̀-ͯ]/g, '')
+  const folded = str.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
     .replace(/ø/gi, 'o')
     .replace(/ł/gi, 'l')
     .replace(/æ/gi, 'ae')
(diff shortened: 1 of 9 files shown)


frontend/src/app/shared/util/search-terms.ts
```diff
@@ -0,0 +1,18 @@
...
+export function normalizeRemoteSearchTerm(str: string): string {
+  return str
+    .replace(/[!@$%^&*_=|~`<>?/";']/g, '')
+    .replace(/\s+/g, ' ')
+    .trim();
+}
+
+export function normalizeLocalSearchTerm(str: string): string {
+  const folded = str.normalize('NFD').replace(/[\u0300-\u036f]/g, '')
+    .replace(/ø/gi, 'o')
```
imnotjames, line 12: Why are we transliterating just these characters?  Is this pulled from elsewhere?

If this is frontend only search it makes sense.  If this is working with the backend, this could break things.

frontend/src/app/shared/util/search-terms.ts
```diff
@@ -0,0 +1,18 @@
+export const SEARCH_DEBOUNCE_MS = 200;
```
imnotjames, line 1: This move feels premature based on gut feeling.  It's kind of lacking context being in the util like this.

Having this live closer to the component where it's being used will give us a better understanding of what we are changing and why if we have to tweak it.

Why are we moving this?

imnotjames (approved):

frontend/src/app/shared/util/search-terms.ts
alexhb1: Because it'll get use in the palette, the browser UI's search, the facet search, etc. All searching the same book/facet endpoints. Alternative is duplicating this in each spot or moving it into a shared file later.

frontend/src/app/shared/util/search-terms.ts
alexhb1: Yep its basically copy/paste from the old browser's client search code in `HeaderFilter.ts`, which gets deleted eventually.

frontend/src/app/shared/util/search-terms.ts
imnotjames: Okay.  Iff this were react I'd suggest a search term debounce hook that raps the debounce hook with the correct type and default.  Here I care a little less.

PR #2280, opened 2026-08-08
fix: return 404 for EntityNotFoundException

### LINKED ISSUE #1858: Deleting missing annotation  emits HTTP 500 rather than 404
If an annotation is missing and you try to delete it, the expected response is a 404 - but instead, we get a 500.

irenesilvadev:
## Description

Deleting a missing annotation raised `EntityNotFoundException`, which previously fell through to the generic handler and became HTTP 500. Clients expect a 404 for a missing resource.

## Linked Issue

Fixes #1858

## Changes

- Map `EntityNotFoundException` to HTTP 404 in `GlobalExceptionHandler` (covers annotations and other services using the same pattern).
- Add a unit test for the new handler.

## Manual Testing Steps

1. Run the unit test `handleEntityNotFoundException_shouldReturnNotFound`.
2. Start the API and call `DELETE /api/v1/annotations/{missingId}` with a non-existent id.
3. Confirm the response status is 404 (not 500) and the body carries the not-found message.

No UI changes.

## AI Disclosure

Cursor - helped draft the exception mapping, unit test, and this PR description; I reviewed all changes.

## Checklist

### DIFF
--- a/backend/src/main/java/org/booklore/exception/GlobalExceptionHandler.java
+++ b/backend/src/main/java/org/booklore/exception/GlobalExceptionHandler.java
@@ -1,5 +1,6 @@
 package org.booklore.exception;
 
+import jakarta.persistence.EntityNotFoundException;
 import jakarta.validation.ConstraintViolationException;
 import lombok.extern.slf4j.Slf4j;
 import org.apache.catalina.connector.ClientAbortException;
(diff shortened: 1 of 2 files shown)


imnotjames:
Please update the PR description to match the template

imnotjames:
Went with the approach in #2302 instead.  Thanks for your contribution regardless.

PR #2289, opened 2026-08-09
fix(books): prevent duplicate book file entries

### LINKED ISSUE #2288: `/api/v1/books/page` returns duplicate entries for books with multiple files
With the new paginated endpoint, when a book has multiple files and also multiple items in other fields (tags, moods etc), it returns a row for each combination. As a result the backend returns duplicate rows for things like book files, e.g. below

Image

alexhb1:
## Description

Fixes an issue with duplicate info returned from the paginated endpoint. With the findAllWithMetadataByIds as one big SQL query, hibernate couldn't de-duplicate the files list when multiple files were tied to a single book. 

## Linked Issue

Fixes #2288 

## Changes

- Removed collection paths from `findAllWithMetadataByIds` to avoid an eager-join of more than one collection in a single query. Those now load in separate queries afterwards. 

## Manual Testing Steps

Check row output for a book with multiple tags, moods, and specifically multiple files. confirm rows are 1 to 1 with unique files and no duplicates are shown. 

## Screenshots (Optional)

N/A

No issues with general use of the endpoint or performance concerns. Fetch time per-page was approx 35ms before and after the fix. 

Looking at the rest of `BookRepository` this was a similar approach to other queries anyway, so this feels more conventional and fixes a bug. 

## AI Disclosure

N/A

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved book loading performance by fetching only the metadata needed for book details.
  * Reduced unnecessary related data loading when retrieving books.

### DIFF
--- a/backend/src/main/java/org/booklore/repository/BookRepository.java
+++ b/backend/src/main/java/org/booklore/repository/BookRepository.java
@@ -85,12 +85,8 @@ public interface BookRepository extends JpaRepository<BookEntity, Long>, JpaSpec
     @Query("SELECT b FROM BookEntity b WHERE (b.deleted IS NULL OR b.deleted = false)")
     List<BookEntity> findAllWithMetadata();
 
-    @EntityGraph(attributePaths = {
-        "metadata", "metadata.comicMetadata",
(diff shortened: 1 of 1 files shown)


imnotjames (approved):
This fixes the specific case you are having, but I haven't been able to check if it causes issues elsewhere.  Regardless, probably fine.

I'll open an issue to address the overarching issue we are having with joins.

Issue #2296, opened 2026-08-09
Manual author search results aren't ranked by relevance


narayandesai:
### What went wrong?

The manual author-search dialog (used to pick an author match by hand) returns up to 10 results from Audnexus, but they aren't ordered by relevance to the search query. Audnexus's own search isn't ranked, so the top-listed candidates are often irrelevant, which buries a real match further down the list — or excludes it entirely, since only 10 results are shown.

### How can we reproduce it?

1. Open an author and use the manual search dialog.
2. Search for a name such as "Hayao Miyazaki".
3. Compare the returned order to the raw Audnexus response for the same query:
   ```
   GET https://api.audnex.us/authors?name=Hayao%20Miyazaki&region=us
   ```
   ```json
   [{"asin":"B001I7AEI2","name":"Hayao Kawai"},{"asin":"B079XYL3FZ","name":"Yoshifumi Miyazaki"},{"asin":"B079XYL3FZ","name":"Yoshifumi Miyazaki"}]
   ```
   The results shown in Grimmory preserve this arbitrary order rather than sorting by closeness to the query name.

**What happened:** The result list isn't ordered by relevance, so the best (or only reasonable) candidate isn't necessarily near the top.
**What I expected:** Results should be sorted by how closely each candidate's name matches the search query.

### What Build of Grimmory are you on?

Stable

### Your setup

- Grimmory version: 3.3.1
- How you're running it: Docker

### Screenshots or error messages (optional)

(API response above serves as reproduction evidence.)

### Before submitting

- [x] I've searched existing issues and discussions and this hasn't been reported yet
- [x] I've reviewed this report and have completed it as fully as I can

imnotjames:
> Audnexus's own search isn't ranked, so the top-listed candidates are often irrelevant

I can't find this documented anywhere.  Where are you reading this?

---

Given the examples you have, that first one is pretty close?  What relevance algorithm are you suggesting as the best or reasonable candidate?

You're searching for Hayao Miyazaki.  None of those matches from Audnexus are correct - the first being the closest with the given name of Hayao, right?

imnotjames:
This is a feature request, not a bug report, isn't it?

narayandesai:
[...]

PR #2299, opened 2026-08-09
fix(author): require a name match before applying an Audnexus quick-match result

### LINKED ISSUE #2295: Author automatch applies the wrong author with no way to catch it
### What went wrong?

Grimmory's author automatch (and "Quick Match") takes the *first* result from Audnexus's author search unconditionally, with no check that the name is actually close to the author being matched. Audnexus's search isn't ranked by relevance, so it can return a completely unrelated author as the top hit. When that happens, Grimmory silently applies that wrong author's bio/photo/ASIN as if it were a confirmed match.

### How can we reproduce it?

1. Add or open an author whose exact name isn't in Audnexus's catalog — e.g. "Hayao Miyazaki".
2. Query Audnexus's search endpoint directly to see what Grimmory receives:
   ```
   GET https://api.audnex.us/authors?name=Hayao%20Miyazaki&region=us
   ```
   Response (no true match present, unranked):
   ```json
   [{"asin":"B001I7AEI2","name":"Hayao Kawai"},{"asin":"B079XYL3FZ","name":"Yoshifumi Miyazaki"},{"asin":"B079XYL3FZ","name":"Yoshifumi Miyazaki"}]
   ```
3. Click "Quick Match" / trigger automatch for this author — Grimmory applies "Hayao Kawai" (the first, unrelated result) as if it were a confirmed match.

A second real example, same behavior: searching `J. Bradford Delong` returns `J. Bradford Hipps` (asin `B0131MUW2E`) as the top hit — a different person entirely — and that gets applied.

**What happened:** The wrong author is applied automatically with no warning.
**What I expected:** Automatch should refuse to apply a result whose name doesn't actually match the author being matched, rather than guessing.

Stable

### Your setup

- Grimmory version: 3.3.1

(API responses above serve as reproduction evidence.)

### Before submitting

- [x] I've searched existing issues and discussions and this hasn't been reported yet (related but distinct: #106 covers the *lack of an undo* after a bad match, not the matching-quality root cause)

narayandesai:
## Description

Author automatch and the "Quick Match" button take Audnexus's search results and, previously, applied the *first* result unconditionally — even when it clearly wasn't the author being matched (Audnexus's search isn't ranked by relevance). This silently attached the wrong author's bio/photo/ASIN with no indication anything went wrong. This PR makes automatch require the candidate's name to actually match the query before applying it, and correctly report "no match" otherwise.

## Linked Issue

Fixes #2295

## Changes

- `AudnexusAuthorParser.quickSearch()` now iterates search results and only returns a candidate whose normalized name matches the query, instead of blindly taking the first result.
- Name comparison strips punctuation and **all** whitespace (not just collapsing it), since sources disagree on word-boundary placement for compound surnames — e.g. Audnexus spells the real "J. Bradford Delong" match as "J. Bradford De Long," and that difference shouldn't defeat an otherwise-correct match.
- Added `namesMatch()` / `normalizeName()` helpers.
- 5 new unit tests: no-match-found → returns null, skips non-matching candidates to find the real one, case/punctuation insensitivity, whitespace-insensitive compound-surname matching (the real Delong/De Long case), and the existing empty-results case.

## Manual Testing Steps

1. Built the image from this branch and ran it against my own real library/author data (`just image-build && just image-run`).
2. Quick Match on "Hannu Rajaniemi" — correctly applies the real Audnexus match (ASIN `B003VNOJY6`); no photo/bio appears only because Audnexus's own record has empty `description`/`image` fields for this author, not a bug.
3. Quick Match on "J. Bradford Delong" — previously silently applied an unrelated author ("J. Bradford Hipps"); now correctly finds and applies the real entry ("J. Bradford De Long") despite the differently-tokenized surname.
4. Quick Match on "Marshall McLuhan" — correctly reports no match, since Audnexus genuinely has no entry for this author (verified directly against the Audnexus API — no candidate even shares his surname).
5. Ran the backend test suite and `just api check` — all passing.

## Screenshots (Optional)

_(none — backend-only change, no UI diff)_

A related fix (re-ranking the *manual* author-search results by name similarity, addressing #2296) exists on a separate branch but hasn't been validated against real-world data yet, so it's being held for its own PR rather than bundled here.

## AI Disclosure

Claude (Claude Code) investigated the root cause — including live-tracing the actual Audnexus API responses that were producing wrong matches — wrote the fix and tests test-first (TDD), and iterated through a couple of rounds of debugging as real-world edge cases (Hannu Rajaniemi, Marshall McLuhan, the "De Long" spacing issue) surfaced during my own testing. I reviewed the code changes, ran the fix against my own production author data across multiple real cases, and verified the results myself before submitting.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved author search accuracy by requiring an exact name match while ignoring differences in capitalization, punctuation, and spacing.
  * Prevented unnecessary detail lookups for unrelated search results.
  * Searches now safely return no result when no matching author is found.

* **Tests**
  * Added coverage for matching, non-matching, empty, and variably formatted author searches, including incomplete candidate results.

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/parser/AudnexusAuthorParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/AudnexusAuthorParser.java
@@ -130,8 +130,7 @@ public class AudnexusAuthorParser implements AuthorParser {
         // Word-boundary placement varies across sources for compound surnames (e.g. "De Long"
         // vs "Delong"), so whitespace is stripped entirely rather than just collapsed.
         return name.toLowerCase(Locale.ROOT)
-                .replaceAll("[.,]", "")
-                .replaceAll("\\s+", "");
+                .replaceAll("(?U)[\\p{P}\\s]+", "");
     }
 
     @Override
--- a/backend/src/main/java/org/booklore/service/metadata/parser/AudnexusAuthorParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/AudnexusAuthorParser.java
@@ -101,8 +101,13 @@ public class AudnexusAuthorParser implements AuthorParser {
 
             if (response.statusCode() == 200) {
                 AudnexusAuthorResponse[] authors = objectMapper.readValue(response.body(), AudnexusAuthorResponse[].class);
-                if (authors.length == 0) return null;
-                return getAuthorByAsin(authors[0].getAsin(), region);
+                for (AudnexusAuthorResponse candidate : authors) {
+                    if (namesMatch(name, candidate.getName())) {
+                        return getAuthorByAsin(candidate.getAsin(), region);
+                    }
+                }
+                log.warn("Audnexus author quick search found no name match for: {}", name);
+                return null;
             }
 
             log.warn("Audnexus author quick search returned status {}", response.statusCode());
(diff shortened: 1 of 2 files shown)


imnotjames:
I don't think this is the right approach here.  We're adding our own sorting on top of the existing audnexus sorting and this is making assumptions about names that I don't think are going to fit edge cases about names.

Instead this feels like the "Quick match" doesn't particularly work correctly and should be removed for the time being.  I'll follow up with that, until we can resolve what should be done with the Author integration.

However, thank you for your contribution.

PR #2340, opened 2026-08-12
fix(ui): logout needs access token before clearing session

### LINKED ISSUE #2339: Logout fails to properly call `logout` endpoint
### What went wrong?

The logout button clears the user's session so the auth token is removed before we make the request.  This causes the logout request to fail.

### How can we reproduce it?

Logout
View browser console logs
See HTTP 401

Stable

### Your setup

- Grimmory 3.3.1
- Firefox 152.0.6

_No response_

### Before submitting

imnotjames:
## Description

The logout button clears the user's session so the auth token is removed before we make the request. This causes the logout request to fail, because the `logout` request _requires_ the login token.

We can actually call the `logout` endpoint properly by waiting to clear the session until after the request completes

## Linked Issue

fixes #2339

## Changes

moves `clearSession` to after logout request completes

## Manual Testing Steps

1. Logout
2. See request complete with 200

It's possible the endpoint isn't working as expected but a log out request requiring a valid access token makes sense to me.

## AI Disclosure

None.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved logout handling by clearing the local session whether the logout request succeeds or fails.
  * Ensured users are redirected to the login screen after logout.
  * Updated refresh-session behavior to handle logout responses in the correct order.

### DIFF
--- a/frontend/src/app/shared/service/auth.service.spec.ts
+++ b/frontend/src/app/shared/service/auth.service.spec.ts
@@ -185,8 +185,8 @@ describe('AuthService', () => {
     const logoutRequest = httpTestingController.expectOne(`${API_CONFIG.BASE_URL}/api/v1/auth/logout`);
     expect(logoutRequest.request.body).toEqual({ refreshToken: 'refresh-token' });
 
-    refreshRequest.flush({ accessToken: 'late-access', refreshToken: 'late-refresh' });
     logoutRequest.flush({ logoutUrl: null });
+    refreshRequest.flush({ accessToken: 'late-access', refreshToken: 'late-refresh' });
 
     await expect(refresh).rejects.toThrow('Refresh response belongs to a stale session');
     expect(service.getInternalAccessToken()).toBeNull();
--- a/frontend/src/app/shared/service/auth.service.ts
+++ b/frontend/src/app/shared/service/auth.service.ts
@@ -177,14 +177,14 @@ export class AuthService {
     this._logoutInProgress.set(true);
 
     const refreshToken = this.getInternalRefreshToken();
-    this.clearSession();
-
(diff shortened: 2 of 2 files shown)


frontend/src/app/shared/service/auth.service.spec.ts
```diff
@@ -185,8 +185,8 @@ describe('AuthService', () => {
     const logoutRequest = httpTestingController.expectOne(`${API_CONFIG.BASE_URL}/api/v1/auth/logout`);
     expect(logoutRequest.request.body).toEqual({ refreshToken: 'refresh-token' });
 
-    refreshRequest.flush({ accessToken: 'late-access', refreshToken: 'late-refresh' });
     logoutRequest.flush({ logoutUrl: null });
+    refreshRequest.flush({ accessToken: 'late-access', refreshToken: 'late-refresh' });
```
imnotjames, line 189: While the test is:

> does not restore tokens or websocket state when a refresh resolves after logout

this previously was set up so that the logout resolved after the refresh.

alexhb1 (approved):
Would it be worth adding a standard RXJS timeout to trigger clearSession() in the case of a hung / bad connection? Those would have no error state but could leave logout as unresponsive and unable to retry after this change.

imnotjames:
> Would it be worth adding a standard RXJS timeout to trigger clearSession() in the case of a hung / bad connection? Those would have no error state but could leave logout as unresponsive and unable to retry after this change.

They should be able to log out again, shouldn't they?

I feel like an error shouldn't even clear the session because the logout request needs to actually happen for the frontend to know it has to do a redirect to do an IDP logout..  at least if we want to be correct.

alexhb1:
Makes sense yeah. No issues anyway, just a random thought. fine to merge :)

PR #2367, opened 2026-08-14
refactor(amazon): use predefined domains

### LINKED ISSUE #2344: Refactor Amazon Provider to use predefined domains
Currently the domains for the Amazon provider are crafted by taking the `domain` string and applying it directly into a URL.  This is unsafe & kind of ugly.

Instead, it should operate more like the audible provider: a `Map<String, String>` of tld -> domain so that user input is more tightly constrained.

Something like:

https://github.com/grimmory-tools/grimmory/blob/8fa769bfd59c9b297b4fd09cb730180e269edd6d/backend/src/main/java/org/booklore/service/metadata/parser/AudibleParser.java#L52-L63

king-407:
## Description

Refactors the Amazon metadata parser to resolve configured Amazon domains through a predefined base URI map instead of constructing URLs directly from the configured domain string.

Null or blank Amazon domain settings fall back to `amazon.com`. Unsupported non-empty domain values now fail loudly instead of silently using a different marketplace.

## Linked Issue

Fixes #2344

## Changes

- Added predefined Amazon marketplace domain resolution.
- Updated Amazon parser URL construction to use the resolved base URI.
- Made Amazon domain lookup null-safe for parent settings objects.
- Added fallback behavior for null and blank Amazon domain settings.
- Added explicit failure behavior for unsupported non-empty Amazon domain values.
- Added focused tests for supported, unsupported, null, and empty domain values.

## Manual Testing Steps

1. Ran `./gradlew test --tests org.booklore.service.metadata.parser.AmazonBookParserTest`
2. Ran `./gradlew check --no-daemon --parallel --build-cache`
3. Ran `git diff --check`

## Screenshots

No UI changes.

## Additional Context

I could not run `just api check` locally because `just` is not installed, so I ran the equivalent backend Gradle check from the `backend/Justfile`.

## AI Disclosure

Codex - helped implement and validate the parser refactor and draft tests; I reviewed all changes.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Amazon marketplace handling across metadata, search, reviews, requests, and date parsing.
  * Unsupported Amazon domains are now rejected safely with a clear validation error.
  * Missing, empty, or invalid marketplace settings now fall back to Amazon US.
  * Amazon URLs and locale-sensitive metadata now consistently use the validated marketplace, improving reliability across supported regions.
  * Metadata retrieval continues to work when optional review settings are unavailable.

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
@@ -1,5 +1,6 @@
 package org.booklore.service.metadata.parser;
 
+import org.booklore.exception.ApiError;
 import org.booklore.model.dto.Book;
 import org.booklore.model.dto.BookMetadata;
 import org.booklore.model.dto.BookReview;
@@ -898,7 +899,7 @@ public class AmazonBookParser implements BookParser, DetailedMetadataProvider {
         }
 
         if (!BASE_URIS.containsKey(domain)) {
-            throw new IllegalArgumentException("Unsupported Amazon domain: " + domain);
+            throw ApiError.INVALID_INPUT.createException("Unsupported Amazon domain: " + domain);
         }
 
         return domain;
--- a/backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
@@ -72,7 +72,7 @@ public class AmazonBookParser implements BookParser, DetailedMetadataProvider {
             "yyyy/M/d", "yyyy/MM/dd", "yyyy年M月d日"
     };
 
-    private static final String DEFAULT_TLD = "com";
+    private static final String DEFAULT_DOMAIN = "com";
 
     private static final Map<String, String> BASE_URIS = Map.ofEntries(
             Map.entry("com", "https://www.amazon.com"),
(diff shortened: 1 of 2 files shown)


backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
```diff
@@ -861,6 +886,15 @@ private static LocaleInfo getLocaleInfoForDomain(String domain) {
         return DOMAIN_LOCALE_MAP.getOrDefault(domain, DEFAULT_LOCALE_INFO);
     }
 
+    private String getTld() {
+        String domain = appSettingService.getAppSettings().getMetadataProviderSettings().getAmazon().getDomain();
```
imnotjames: Are there any concerns here for any of these being null?

backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
```diff
@@ -861,6 +886,15 @@ private static LocaleInfo getLocaleInfoForDomain(String domain) {
...
+        return domain != null && BASE_URIS.containsKey(domain) ? domain : DEFAULT_TLD;
```
imnotjames: If the domain is null (not set) the fallback to default makes sense. 

However, if someone is asking for a domain and it doesn't exist: does it make more sense from their perspective that the metadata provider would fail loudly rather than silently do something different from what they asked?

imnotjames (commented):
The PR template looks off.  Can you please use our PR template?

backend/src/main/java/org/booklore/service/metadata/parser/AmazonBookParser.java
king-407: Thanks for the review.

I updated the domain handling so null or blank Amazon domain settings still fall back to the default `com`, but unsupported non-empty domains now fail loudly with an `IllegalArgumentException` instead of silently using `amazon.com`.

I also made the settings lookup null-safe for the parent settings objects and updated the unsupported-domain test accordingly.

imnotjames:
Thanks for the contribution.

imnotjames (approved):

PR #2380, opened 2026-08-16
fix(books): remove stale facet cache

### LINKED ISSUE #2304: Stale `/api/v1/books/facets` Caffeine cache
`BookFacetService` (#1898) includes a 30-second TTL cache with max 200 entries and no invalidation (approx line 65). At the moment, it appears to be a cause of stale filter entries in the new browser UI as no book or metadata command can trigger an update in that window. This also affects larger features like magic shelf lists. 

The frontend already has its own caching mechanism from tanstack query. Repeat requests are cached for 30 seconds so the basic time limit is already present in FE requests to the endpoint. Cache invalidation is also set up for the occasions it needs to re-fetch.

Either we keep the 30 seconds but add the ability to clear cache upon FE action, or just remove the cache entirely and let the frontend handle it safely. Imo I think remove is cleaner, but either way makes sense. 

(cc @zachyale @imnotjames )

king-407:
## Description

Removes the backend Caffeine cache from `BookFacetService` so `/api/v1/books/facets` always reflects the latest book and metadata state.

The frontend already caches facet requests through TanStack Query and can invalidate them after user actions. Keeping an additional backend cache could return stale facet/filter values for up to 30 seconds after book or metadata changes.

## Linked Issue

Fixes #2304

## Changes

- Removed the 30-second Caffeine cache from `BookFacetService`.
- Removed the test-only cache clearing helper.
- Updated facet service coverage to verify newly added books appear in facet results immediately.

## Manual Testing Steps

1. Ran `./gradlew test --tests org.booklore.service.browse.BookFacetServiceTest`
2. Ran `./gradlew check --no-daemon --parallel --build-cache`
3. Ran `git diff --check`

## Screenshots

No UI changes.

## Additional Context

I could not run `just api check` locally because `just` is not installed, so I ran the equivalent backend Gradle check from the `backend/Justfile`.

## AI Disclosure

Codex - helped implement and validate the cache removal and regression test; I reviewed all changes.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Book browsing facets now reflect newly added books immediately.
  * Updated genre counts and author values are shown without requiring cached data to expire or be cleared.

### DIFF
--- a/backend/src/main/java/org/booklore/service/browse/BookFacetService.java
+++ b/backend/src/main/java/org/booklore/service/browse/BookFacetService.java
@@ -1,7 +1,5 @@
 package org.booklore.service.browse;
 
-import com.github.benmanes.caffeine.cache.Cache;
-import com.github.benmanes.caffeine.cache.Caffeine;
 import jakarta.persistence.EntityManager;
 import jakarta.persistence.PersistenceContext;
 import jakarta.persistence.Tuple;
@@ -15,7 +13,6 @@ import jakarta.persistence.criteria.Root;
 import lombok.RequiredArgsConstructor;
 import org.booklore.browse.FacetLogic;
 import org.booklore.browse.Link;
-import org.booklore.browse.ParamsHash;
 import org.booklore.config.security.service.AuthenticationService;
 import org.booklore.model.dto.BookLoreUser;
 import org.booklore.model.dto.browse.FacetGroupsResponse;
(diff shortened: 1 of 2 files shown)


imnotjames:
Thanks for the contribution.

However, without having a discussion and talking through the approaches I think you may have jumped to an extreme for this task.  I don't believe that we should remove the caching entirely, so I'll be closing this PR out.

PR #2452, opened 2026-08-24
fix md resolution wrapping

iamntz:
Very small fix of this issue: (ipad, portrait, on libraries/shelves views)

image

## Summary by CodeRabbit

* **Style**
  * Updated the book browser header layout to allow content to wrap onto multiple lines, improving responsiveness on narrower screens.


alexhb1:
Thanks for the contribution, but unfortunately not necessary as next release we'll be introducing a whole new browser UI for books.

PR #2456, opened 2026-08-25
fix(api): allow deleting primary book format when other formats exist

### LINKED ISSUE #2455: Deleting the primary book format fails with 400 when other formats exist
### What went wrong?

When a book has multiple formats (e.g. a PDF and an EPUB) and the PDF is the primary file, trying to delete the PDF from the book's "Delete format" dropdown fails with an error toast:

```
Primary book file cannot be processed as an additional file: 10153
Failed to delete book format: Http failure response for .../api/v1/books/9754/files/10153: 400 OK
```

The UI explicitly offers the primary file in the delete dropdown (labeled `[Primary]`) and `BookFileService.deleteBookFile` even contains cache logic to promote the next format to primary on success — but the backend rejects the request, so deleting a primary format is currently impossible without deleting the whole book.

### How can we reproduce it?

1. Have a book with two book formats, e.g. PDF (primary) and EPUB.
2. Open the book's metadata view → the "..." menu → Delete format.
3. Choose the file marked `[Primary]` and confirm.

What happened: HTTP 400, "Primary book file cannot be processed as an additional file".
What you expected: the PDF is deleted and the EPUB becomes the primary format.

### Root cause (from reading the code)

`DELETE /api/v1/books/{bookId}/files/{fileId}` routes to `AdditionalFileService.deleteAdditionalFile`, which calls `validateAdditionalFile` — a guard added in #463 that throws for the primary file. That guard is correct for the *download* endpoint (the primary has its own download route), but applying it to *delete* broke the UI's delete-primary flow.

Notably, the primary file is **derived**, not stored: `BookEntity.getPrimaryBookFile()` picks by the library's format priority, falling back to lowest file id. So deleting the primary would automatically promote the next format — no promotion logic is needed. Only deleting the *last* book format needs to stay blocked (that should be a whole-book delete instead).

Stable

### Your setup

- Grimmory version: current `develop` (0e675495b)
- Browser: Chrome
- Operating system: Linux

### Before submitting

a10kiloham:
Fixes #2455

## Problem

Deleting a book's primary format via the UI's "Delete format" dropdown fails with HTTP 400:

```
Primary book file cannot be processed as an additional file: <fileId>
```

`AdditionalFileService.deleteAdditionalFile` reuses the primary-file guard introduced in #463. That guard is right for downloads (the primary file has its own download endpoint) but wrong for deletes: the frontend deliberately offers the primary file in the delete dropdown and already handles promoting the next format on success.

## Fix

Since the primary file is derived (`BookEntity.getPrimaryBookFile()`: library format priority, then lowest id), deleting the primary automatically promotes the next book format — no extra promotion logic is needed. The delete path now only refuses to remove the **last remaining book format**, which would leave the book without a readable file (deleting the book itself is the right operation there). Supplementary (non-book) files are unaffected, and the download guard is unchanged.

## Testing

- Updated `deleteAdditionalFile_WhenFileIsPrimaryBookFile_ShouldThrowException` → `deleteAdditionalFile_WhenFileIsLastBookFormat_ShouldThrowException` (new message/behavior).
- Added `deleteAdditionalFile_WhenFileIsPrimaryButOtherFormatsExist_ShouldDeleteSuccessfully`, which also asserts the other format becomes primary afterwards.
- Full backend test suite passes locally (`./gradlew test`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01AuUR58BRNecW8yD5hwRJKy

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved book file deletion behavior: primary book formats can now be removed when alternative formats are available.
  - Prevented deletion of the last remaining book format; books with no formats must be deleted instead.
  - Automatically assigns another available format as the primary file after deletion.

### DIFF
--- a/backend/src/main/java/org/booklore/service/file/AdditionalFileService.java
+++ b/backend/src/main/java/org/booklore/service/file/AdditionalFileService.java
@@ -59,7 +59,7 @@ public class AdditionalFileService {
 
         BookFileEntity file = fileOpt.get();
         BookEntity book = file.getBook();
-        validateAdditionalFile(file, book);
+        validateNotLastBookFormat(file, book);
 
         try {
             monitoringRegistrationService.unregisterSpecificPath(file.getFullFilePath().getParent());
@@ -160,6 +160,17 @@ public class AdditionalFileService {
     private void validateAdditionalFile(BookFileEntity file, BookEntity book) {
         if (book != null && book.getPrimaryBookFile() != null && file.getId().equals(book.getPrimaryBookFile().getId())) {
             throw new IllegalArgumentException("Primary book file cannot be processed as an additional file: " + file.getId());
+        }
+    }
+
+    // Deleting the primary format is allowed as long as another book format remains: the
+    // primary is derived (library format priority, then lowest id), so the next format is
+    // promoted automatically. Only the last book format is protected, since removing it
+    // would leave the book without a readable file — delete the book itself instead.
+    private void validateNotLastBookFormat(BookFileEntity file, BookEntity book) {
+        if (file.isBook() && book != null && book.getBookFiles().stream()
+                .noneMatch(bf -> bf.isBook() && !bf.getId().equals(file.getId()))) {
+            throw new IllegalArgumentException("Cannot delete the last book format file: " + file.getId() + ". Delete the book instead.");
         }
     }
 }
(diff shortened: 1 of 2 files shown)


imnotjames:
Does not match template.

PR #2469, opened 2026-08-27
refactor(api): add missing book facet fields

### LINKED ISSUE #1849: [DRAFT] Books browse pagination: improve facets to achieve feature parity with the frontend sort/filtering

### LINKED ISSUE #2436: Add remaining filter options in `BookFacetService`
Most filter options are readily available but not advertised in `BookFacetService`

alexhb1:
## Description

Adds remaining book facet items to the endpoint. 

## Linked Issue

Fixes #2436 

## Changes

20 new filter categories now visible in the facets endpoint: 
- content rating, the six provider ratings, age rating, page count, match score, published year, library, shelf, shelf status, read status, personal rating, file size, and the four comic ones. 

Three added filter and sort keys that were missing entirely:
- Narrator, Lubimyczytac rating and Audible rating

Misc: 
- Updated numeric filters from a fixed set of buckets to also accept a min, max or both values, allowing for frontend filter UI to show min and max inputs for example. 
- Fixes the file size filter to use the primary file for measurement, like current browser does client-side. Before, it also considered secondary files. 

## Manual Testing Steps

Test the facets endpoint for returned values across these new fields. 

## Screenshots (Optional)

N/A

The numerical range filtering is actually faster in testing. Did a test with a 1M book database and filtering took 1.4s vs the previous buckets taking 2.7s. 

## AI Disclosure

N/A

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Numeric book filters now support exact values and flexible open or closed ranges for ratings, scores, publication years, page counts, and file sizes.
  * Added narrator, Lubimyczytac rating, and Audible rating facets.
  * Expanded browsing facets for content ratings, libraries, shelves, reading status, personal ratings, file size, and comic-related metadata.

* **Bug Fixes**
  * Invalid numeric filter values now return a clear HTTP 400 error instead of being processed incorrectly.
  * File-size filtering now evaluates the book’s primary format file more accurately.

### DIFF
--- a/backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
+++ b/backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
@@ -15,6 +15,7 @@ import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
+import java.util.function.Function;
 
 public class AppBookSpecification {
 
@@ -59,6 +60,67 @@ public class AppBookSpecification {
             throw new APIException("Invalid " + paramName + " values: " + invalid, HttpStatus.BAD_REQUEST);
         }
         return result;
+    }
+
+    private record NumericRange<T>(T min, T max) {
+    }
+
+    private static <T> List<NumericRange<T>> parseNumericRanges(List<String> values, Function<String, T> parser, String paramName) {
+        List<String> invalid = new ArrayList<>();
+        List<NumericRange<T>> ranges = new ArrayList<>();
+        for (String value : values) {
+            if (value == null || value.isBlank()) {
+                continue;
+            }
+            NumericRange<T> range = parseNumericRange(value.trim(), parser);
+            if (range == null) {
+                invalid.add(value.trim());
+            } else {
+                ranges.add(range);
+            }
+        }
+        if (!invalid.isEmpty()) {
+            throw new APIException("Invalid " + paramName + " values: " + invalid
+                    + ". Expected a number, min-max, min-, or -max.", HttpStatus.BAD_REQUEST);
+        }
+        return ranges;
+    }
+
+    private static <T> NumericRange<T> parseNumericRange(String value, Function<String, T> parser) {
+        try {
+            int dash = value.indexOf('-');
+            if (dash < 0) {
+                T exact = parser.apply(value);
+                return new NumericRange<>(exact, exact);
+            }
+            String minPart = value.substring(0, dash).trim();
+            String maxPart = value.substring(dash + 1).trim();
+            if (minPart.isEmpty() && maxPart.isEmpty()) {
+                return null;
+            }
+            return new NumericRange<>(
+                    minPart.isEmpty() ? null : parser.apply(minPart),
+                    maxPart.isEmpty() ? null : parser.apply(maxPart));
+        } catch (NumberFormatException e) {
+            return null;
+        }
+    }
+
+    private static <T extends Comparable<? super T>> Predicate rangesPredicate(
+            CriteriaBuilder cb, Expression<T> field, List<NumericRange<T>> ranges) {
+        List<Predicate> predicates = new ArrayList<>();
+        for (NumericRange<T> range : ranges) {
+            if (range.min() != null && range.max() != null) {
+                predicates.add(cb.and(
+                        cb.greaterThanOrEqualTo(field, range.min()),
+                        cb.lessThanOrEqualTo(field, range.max())));
+            } else if (range.min() != null) {
+                predicates.add(cb.greaterThanOrEqualTo(field, range.min()));
+            } else {
+                predicates.add(cb.lessThanOrEqualTo(field, range.max()));
+            }
+        }
+        return cb.or(predicates.toArray(Predicate[]::new));
     }
 
     @SuppressWarnings("unchecked")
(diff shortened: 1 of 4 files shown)


SergeiAK:
Good to see these exposed. Two small things I noticed while reading the diff, both
about what a facet value looks like to a person rather than to a filter.

`library` and `shelf` group on the row id, and `toGroup()` puts that same string
into both `title` and `value`, so the group comes back with entries labelled `17`
or `23`. `FacetLink` keeps `title` and `value` as separate fields, so the id could
stay as the filter value while the title carries the library or shelf name.

`page_count`, `file_size` and the rating fields are `Integer` and `Double`, so
their distinct values run into the thousands, while `count()` caps the list at
`MAX_VALUES = 100` ordered by frequency. As min/max filters — which this PR adds —
they work well; as a list of values it comes out as the hundred most common
numbers. Might be worth leaving those out of the facet groups, or bucketing them
into ranges.

Nothing blocking, just what stood out. We read this endpoint from outside the web
UI, where the title is what the user ends up seeing.

backend/src/main/java/org/booklore/service/browse/BookSortRegistry.java
```diff
@@ -38,6 +38,7 @@ private static SortRegistry<BookEntity> build() {
                 "title", "seriesName", "seriesNumber", "publisher", "publishedDate",
                 "amazonRating", "amazonReviewCount", "goodreadsRating", "goodreadsReviewCount",
                 "hardcoverRating", "hardcoverReviewCount", "ranobedbRating",
+                "lubimyczytacRating", "audibleRating", "audibleReviewCount",
```
imnotjames: Can this be moved to a separate PR from the facet changes - it muddies the water a bit here.

alexhb1:
@SergeiAK These endpoints are not complete, I would not advise anyone to make use of them yet

backend/src/main/java/org/booklore/service/browse/BookFacetService.java
```diff
@@ -157,7 +197,11 @@ private static String href(String path, String preserved) {
         return root.join("metadata", JoinType.LEFT);
     }
 
-    private record FacetDef(String key, String title, Function<Root<BookEntity>, Expression<?>> value) {
+    private interface FacetValueSource {
+        Expression<?> value(CriteriaBuilder cb, Root<BookEntity> root, Long userId);
```
imnotjames: It would be good to match the `Expression`'s `apply`?

```suggestion
        Expression<?> apply(CriteriaBuilder cb, Root<BookEntity> root, Long userId);
```

backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
```diff
@@ -61,6 +62,67 @@ private static List<Long> parseLongList(List<String> values, String paramName) {
...
+            if (dash < 0) {
+                T exact = parser.apply(value);
+                return new NumericRange<>(exact, exact);
+            }
+            String minPart = value.substring(0, dash).trim();
+            String maxPart = value.substring(dash + 1).trim();
+            if (minPart.isEmpty() && maxPart.isEmpty()) {
+                return null;
+            }
+            return new NumericRange<>(
```
imnotjames: Should we invert the range so the `min`/`max` are always aligned?  eg, what if `max < min`?  Should we flip them?

backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
```diff
@@ -61,6 +62,67 @@ private static List<Long> parseLongList(List<String> values, String paramName) {
...
+            }
+        }
+        if (!invalid.isEmpty()) {
+            throw new APIException("Invalid " + paramName + " values: " + invalid
+                    + ". Expected a number, min-max, min-, or -max.", HttpStatus.BAD_REQUEST);
+        }
+        return ranges;
+    }
+
+    private static <T> NumericRange<T> parseNumericRange(String value, Function<String, T> parser) {
```
imnotjames: Do we need to handle negative numbers in the range anywhere?  I think we're.. probably safe?  If not, do we want to instead use `,` or `;` to separate the values?

backend/src/main/java/org/booklore/service/browse/BookSortRegistry.java
alexhb1: yep will do

backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
alexhb1: I think rejecting it is cleaner? HTTP 400 if the range is not within accepted bounds

backend/src/main/java/org/booklore/app/specification/AppBookSpecification.java
alexhb1: Just looking, github uses `..` for range, we could use that?

imnotjames: Sure, I'd be open to that.  Mostly trying to avoid pain down the line.

alexhb1: That should be added now btw

imnotjames (approved):

PR #2472, opened 2026-08-27
fix: make notebook bookmarks clickable to open reader

warreth:
I made the bookmarks on the notebook clickable. When clicked, it goes right in the reader to the bookmark.

## Changes
- Backend: Added cfi and pageNumber to NotebookEntry DTO
- Frontend: Made bookmark entries clickable with navigation
- Added query param handling for deep linking to bookmark locations

## Testing
- Navigate to /notebook
- Click on a bookmark → reader opens at exact location

## Summary by CodeRabbit

* **New Features**
  * Notebook bookmarks are now clickable and open directly in the appropriate reader.
  * Bookmarks preserve and use saved locations, including ebook positions and page numbers.
  * Ebook readers can open directly to a specified location from bookmark links.
* **Style**
  * Clickable notebook entries now show pointer, hover, and active-state feedback for clearer interaction.

### DIFF
--- a/backend/src/main/java/org/booklore/model/dto/NotebookEntry.java
+++ b/backend/src/main/java/org/booklore/model/dto/NotebookEntry.java
@@ -22,6 +22,8 @@ public class NotebookEntry {
     private String style;
     private String chapterTitle;
     private String primaryBookType;
+    private String cfi;
+    private Integer pageNumber;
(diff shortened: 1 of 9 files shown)


imnotjames:
Doesn't match PR template.

PR #2478, opened 2026-08-28
fix: bookmarks created via API don't show reader ribbon

warreth:
## The issue:
Bookmarks created outside the reader (via POST /api/v1/bookmarks) never
triggered the bookmark ribbon in the ebook reader, while bookmarks made
inside the reader did.

## Root cause: 
the reader compared the current CFI against each bookmark's
CFI with strict string equality. Bookmarks created in the reader store
the exact point CFI the reader emits, so they matched. Bookmarks created
from a selection (which is what some API clients send) store a range CFI
without the idrefs the reader uses on its own CFIs, e.g.

  reader position: epubcfi(/6/122!/4/2[chapter-56],,/2/22/1:394)
  api-created:     epubcfi(/6/122!/4/2/2/22,/1:0,/1:10)

Same section, structurally different strings, so equality failed.

## Fix: 
compare bookmarks by the section part of the CFI (everything before
the `!` indirection step) instead of the full string. That part is stable
across both formats. Changed updateBookmarkIndicator() and toggleBookmark().

Noticed while testing crosspoint-bookmark-sync against the API.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved bookmark detection when navigating within an ebook.
  - Bookmark indicators now remain accurate across equivalent CFI formats.
  - Toggling a bookmark correctly recognizes existing bookmarks created through different navigation or API flows.

### DIFF
--- a/frontend/src/app/features/readers/ebook-reader/ebook-reader.component.ts
+++ b/frontend/src/app/features/readers/ebook-reader/ebook-reader.component.ts
@@ -520,9 +520,26 @@ export class EbookReaderComponent implements OnInit {
   private updateBookmarkIndicator(): void {
     const currentCfi = this.progressService.currentCfi;
     const isBookmarked = currentCfi
-      ? this.sidebarService.bookmarks().some(bookmark => bookmark.cfi === currentCfi)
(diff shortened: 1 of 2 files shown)


imnotjames:
Doesn't match template, doesn't link to issue.

Just because something is created from the API doesn't mean it should be an invalid CFI.

PR #2516, opened 2026-09-03
fix(douban): ignore extraneous data when parsing window data

### LINKED ISSUE #2515: Douban metadata search fails due to greedy window.__DATA__ regex
### What went wrong?

Douban metadata search fails with:

Failed to parse JSON: Unexpected character (';' (code 59)):
expected a valid value

The Douban search page successfully returns results inside:

window.__DATA__ = {...};

However, DoubanBookParser currently uses:

Pattern.compile("window\\.__DATA__\\s*=\\s*(\\{.*\\});", Pattern.DOTALL);

Because `.*` is greedy, the match can extend past `window.__DATA__`
and capture subsequent JavaScript objects/statements on the page.

The captured string therefore contains valid JSON followed by `;`
and additional JavaScript, causing ObjectMapper.readTree() to fail.

Changing the regex to non-greedy fixes the issue:

Pattern.compile("window\\.__DATA__\\s*=\\s*(\\{.*?\\})\\s*;", Pattern.DOTALL);

### How can we reproduce it?

Search metadata for:

高效原力

Douban returns:
https://search.douban.com/book/subject_search?search_text=高效原力

The HTML contains a valid result:
window.__DATA__ = {"count":15,...,"total":3};

but Grimmory returns no Douban search results.

Stable

### Your setup

Grimmory 3.3.3 in Docker

_No response_

### Before submitting

imnotjames:
## Description

douban seems to have started adding extra data to the script tag which contained the `window.__DATA__` we use for parsing searches.  this tightens up our detection of the script tag & also ensures we can still parse the data by reaching for a lower level Jackson parser

## Linked Issue

fixes #2515 

## Changes

* uses JSOUP script / data node parsing rather than accessing HTML directly
* uses the Jackson parser to ignore non-JSON data after `window.__DATA__ =`
* adds tests + makes the metadata parser more testable

## Manual Testing Steps

1. search douban for `高效原力`
2. see search results
3. rejoice

## Screenshots (Optional)

N/A

While we could have used a regex again, this led to a bit of jank - especially in cases where the strings in the JSON object contain text that _looked_ like JSON.

## AI Disclosure

None.

## Checklist

## Bug Fixes
- Improved Douban metadata parsing for more reliable book search and detail results.
- Empty or unsuccessful Douban searches now return empty results instead of causing null-related issues.
- Improved handling of Douban page responses for more consistent metadata retrieval.

## Tests
- Added coverage for empty and temporarily restricted searches.
- Added coverage confirming multiple matching books are parsed successfully.
- Added representative Douban search and book-page data to validate metadata extraction.

### DIFF
--- a/backend/src/main/java/org/booklore/service/metadata/parser/DoubanBookParser.java
+++ b/backend/src/main/java/org/booklore/service/metadata/parser/DoubanBookParser.java
@@ -18,10 +18,13 @@ import org.jsoup.nodes.Node;
 import org.jsoup.nodes.TextNode;
 import org.jsoup.select.Elements;
 import org.springframework.stereotype.Service;
+import tools.jackson.core.JsonToken;
 import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.ObjectMapper;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.time.Instant;
 import java.time.LocalDate;
 import java.time.ZoneOffset;
@@ -92,7 +95,7 @@ public class DoubanBookParser implements BookParser {
     public List<BookMetadata> fetchMetadata(Book book, FetchMetadataRequest fetchMetadataRequest) {
         List<BookMetadata> searchResults = getDoubanSearchResults(book, fetchMetadataRequest);
         if (searchResults == null || searchResults.isEmpty()) {
-            return null;
+            return List.of();
         }
 
         // For detailed metadata, fetch full book information for top results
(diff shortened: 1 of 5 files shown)


backend/src/main/java/org/booklore/service/metadata/parser/DoubanBookParser.java
```diff
@@ -117,6 +120,38 @@ public List<BookMetadata> fetchMetadata(Book book, FetchMetadataRequest fetchMet
...
+                log.debug("Extracted JSON data: {}", jsonData);
+
+                // Parse just the first object we find, so we can ignore the rest
+                // of the data in the script tag.
+                try (
+                      var is = new ByteArrayInputStream(jsonData.getBytes(StandardCharsets.UTF_8));
+                      var parser = objectMapper.createParser(is)
+                ) {
+                    if (JsonToken.START_OBJECT.equals(parser.nextToken())) {
+                        return objectMapper.readTree(parser);
```
alexhb1: Very nitpicky, but i guess if douban ever changes the page and adds another block of code, even an added JS semicolon would break this. seems like a quick robustness fix. 

```suggestion
                        return objectMapper.readerFor(JsonNode.class)
                               .without(DeserializationFeature.FAIL_ON_TRAILING_TOKENS)
                               .readTree(parser);
```

alexhb1 (approved):

backend/src/main/java/org/booklore/service/metadata/parser/DoubanBookParser.java
imnotjames: Not sure I understand: There are extra semicolons afterwards and extra bits of information afterwards.

The data we're parsing is:

```
{"total": 0, "start": 0, "count": 15, "error_info": "\u641c\u7d22\u8bbf\u95ee\u592a\u9891\u7e41\u3002", "items": [], "text": "\u9ad8\u6548\u539f\u529b", "report": {"qtype": "195", "tags": "\u8bfb\u4e66"}};
      window.__USER__ = { }
```

So after the data there's already extra trailing tokens:

```
; window.__USER__ = { }
```

What test case should I add to exhibit the issue you're suggesting?

alexhb1: yeah its handled fine for the actual parsed data, this is about any potential semicolons after the second object (if they're ever added), e.g: 

`window.__USER__ = { };`

then it'll reject the whole thing. doesn't stop it from working now, but could be a very inconspicuous thing that changes and starts failing.

imnotjames: Isn't that all ignored anyway?  Since I'm using the custom `ObjectMapper` parser.

eg, I updated the fixture so it's got an extra semicolon like you suggested and it still parses it correctly?

I am all for making a change, I just want to add a test that shows it's fixing what we're talking about.

alexhb1: I just tried again and this is what I get: 

```
2026-09-12T16:35:33.614+01:00  WARN 94244 --- [booklore-api] [oundedElastic-1] o.b.s.metadata.parser.DoubanBookParser   : Failed to parse JSON: Unexpected character (';' (code 59)): expected a valid value (JSON String, Number, Array, Object or token 'null', 'true' or 'false')
 at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); byte offset: #2808]
2026-09-12T16:35:33.615+01:00  WARN 94244 --- [booklore-api] [oundedElastic-1] o.b.s.metadata.parser.DoubanBookParser   : No JSON data found in Douban search response
```

Just with that change to add a semicolon above. Again, it's nitpicky but a small potential break in the future.

imnotjames: Like in https://github.com/grimmory-tools/grimmory/pull/2516/commits/eb0a31e3a537d9ebd80d959c81e8e8cac7ceb33e ?

imnotjames: Okay - see where the issue is now, thanks.   The regex we were using was hiding the fact that it was erasing the trailing data.

alexhb1 (approved):

PR #2523, opened 2026-09-04
fix(book): stop N+1 SELECT on comicMetadata in findAllWithMetadata*

### LINKED ISSUE #2482: findAllWithMetadata issues one SELECT per book for comic_metadata
### What went wrong?

`GET /api/v1/books` issues one extra `SELECT` against `comic_metadata` per book. On my library of 19,613 books (and zero comic) a single request runs **25,514 statements, 19,613 of them `comic_metadata` selects**, for 4.15 s of database time.

BookRepository.findAllWithMetadata() [declares](https://github.com/alexhb1/grimmory/blob/develop/backend/src/main/java/org/booklore/repository/BookRepository.java#L84):

```java
@EntityGraph(attributePaths = {"metadata", "metadata.comicMetadata", "library"})
```

The nested `metadata.comicMetadata` path produces no join. The emitted SQL contains only the two first-level paths:

```sql
select ... from book be1_0
  join library l1_0 on l1_0.id=be1_0.library_id
  left join book_metadata m1_0 on be1_0.id=m1_0.book_id
where (be1_0.deleted is null or be1_0.deleted=false)
```

So `comicMetadata` stays lazy and `BookMapperV2.mapMetadata` touches it for every book, costing one `SELECT` each. Books that *do* have a `comic_metadata` row cost far less (3 vs 26 statements at N=25 on H2) so the cost falls on the books without one, and a library like mine without comics get hit worst.

I tried to narrow down why the nested path isn't honoured. Statement counts for the same query over 25 books with `comic_metadata` empty:

| fetch specification | statements |
|---|---:|
| `{metadata, metadata.comicMetadata, library}` (current) | 26 |
| `{metadata.comicMetadata, library}` | 26 |
| `{metadata, metadata.authors, library}` | 1 |
| `{libraryPath, libraryPath.library}` | 1 |
| `{comicMetadata}`, first-level, `BookMetadataRepository` root | 1 |
| `LEFT JOIN FETCH b.metadata m` + `LEFT JOIN FETCH m.comicMetadata` | 1 |

`type = LOAD` instead of `FETCH`, and a derived query instead of `@Query`, both also give 26.

The one combination not honoured is a nested to-one hanging off `BookEntity.metadata`, a `mappedBy` `@OneToOne`. I can't separate that from the other thing that makes `comicMetadata` unique — its read-only shared-FK `@JoinColumn(insertable = false, updatable = false)` — because it's the only association in the model mapped that way. Reproduces on MariaDB and H2.

Two things I haven't measured: `findAllWithMetadataByLibraryIds()` carries the same nested path and backs the same endpoint for non admin users, so it looks like the same cost on that branch; and twenty other `@EntityGraph` declarations in `BookRepository` use the nested path too.

Related to the load-approach trade-offs in #2301, but narrower: this is a declared fetch path that is silently ignored. It's tangent to the pagination work in #2031/#1845. 

Replacing the entity graph on this one method with explicit `LEFT JOIN FETCH`es takes the request to 5,899 statements and 1.85 s of database time with an unchanged response. The improvement should transfer to whatever load strategy #2301 settles on.

On a 19,613-book library with no comics, per `GET /api/v1/books`:

| | before | after |
|---|---:|---:|
| statements | 25,514 | 5,899 |
| of which `comic_metadata` selects | 19,613 | 0 |
| DB time | 4.15 s | 1.85 s |
| TTFB | 8.18 s | 3.11–3.63 s warm |

### How can we reproduce it?

It reproduces on H2 in the existing test harness. The books must have **no** `comic_metadata` row

1. Persist N books, each with a `BookMetadataEntity` and no `comic_metadata` row.
2. `entityManager.flush()`, then `clear()`, and clear Hibernate `Statistics`.
3. Call `bookRepository.findAllWithMetadata()`.
4. On each result call `getMetadata().getComicMetadata()` — it returns null, and that access alone is what fires the per-book `SELECT`. (`BookMapperV2.mapMetadata` reads it for every book, with no book-type check.)
5. Read `statistics.getPrepareStatementCount()`, with `hibernate.generate_statistics` enabled.

**What happened:** the count is exactly N+1 — one root query plus one per book. 26 at N=25, 101 at N=100, 401 at N=400. 
**What I expected:** a count that doesn't scale with N.

The production counts above are from a `performance_schema.events_statements_summary_by_digest` diff taken immediately before and after one request.

Stable

### Your setup

- Grimmory version: v3.3.3
- Database: MariaDB 11.8.8
- Library size: 19,613 books, `comic_metadata` empty
- Also reproduced on H2 via the project's own test harness

n/a

### Before submitting

nfvelten:
## Description

`@EntityGraph` doesn't honour the nested to-one path `metadata.comicMetadata`, so no join is produced, `comicMetadata` stays lazy, and every caller that reads it (`BookMapperV2.mapMetadata`) fires one extra SELECT per book. Replacing the entity graph with an explicit `LEFT JOIN FETCH` collapses that back to a single statement.

Per the follow-up measurements in the issue, this affects four repository methods, not only the one originally reported:

* `findAllWithMetadata()`, the case in the issue
* `findAllWithMetadataByLibraryIds()`, non-admin `/api/v1/books`
* `findAllWithMetadataByIds()`, the browse path (`sort/facet/query/cursor`), used widely
* `findAllWithMetadataPage()`, legacy admin listing

## Linked Issue

Fixes #2482

## Changes

* `BookRepository`: replaced `@EntityGraph` with explicit `LEFT JOIN FETCH` on the four methods above. `libraryPath` is fetched with `LEFT JOIN FETCH` too, since it is nullable and an inner join would silently drop books without a library path.
* `BookRepositoryDataJpaTest`: added one statement-count regression test per method. Each persists N books with no `comic_metadata` row, runs the query, touches `getComicMetadata()` on every result, and asserts the Hibernate statement count does not scale with N.
* Fixed a pre-existing test (`findAllWithMetadataByIds_executesAgainstJpaMetamodel`) that hardcoded book id `1L` instead of the persisted id. It only passed when it happened to run first in the class, and broke once IDENTITY allocation moved past 1 because of the new tests.

## Manual Testing Steps

1. `just api test-class test_class=org.booklore.repository.BookRepositoryDataJpaTest` on `main` with only the new tests applied: the statement-count assertions fail, showing N+1 on all four methods.
2. Apply the `BookRepository` change and rerun the same class: all tests pass, one statement per query regardless of N.
3. `just api check` on the branch: green.

## AI Disclosure

Claude Code. Used to explore the repository, write the tests and draft the patch. I reviewed and ran everything myself.

## Checklist

### DIFF
--- a/backend/src/main/java/org/booklore/repository/BookRepository.java
+++ b/backend/src/main/java/org/booklore/repository/BookRepository.java
@@ -80,9 +80,10 @@ public interface BookRepository extends JpaRepository<BookEntity, Long>, JpaSpec
     @Query("SELECT b.id FROM BookEntity b WHERE b.libraryPath.id IN :libraryPathIds AND (b.deleted IS NULL OR b.deleted = false)")
     List<Long> findAllBookIdsByLibraryPathIdIn(@Param("libraryPathIds") Collection<Long> libraryPathIds);
 
-    // Only ToOne paths in EntityGraph; collections (authors, categories, moods, tags, shelves, bookFiles) loaded via @BatchSize.
-    @EntityGraph(attributePaths = {"metadata", "metadata.comicMetadata", "library"})
-    @Query("SELECT b FROM BookEntity b WHERE (b.deleted IS NULL OR b.deleted = false)")
+    // Nested to-one path metadata.comicMetadata is not honoured by @EntityGraph (produces no join,
+    // leaving comicMetadata lazy and causing an N+1 SELECT per book) — fetched explicitly instead.
+    // collections (authors, categories, moods, tags, shelves, bookFiles) loaded via @BatchSize.
+    @Query("SELECT b FROM BookEntity b LEFT JOIN FETCH b.metadata m LEFT JOIN FETCH m.comicMetadata JOIN FETCH b.library WHERE (b.deleted IS NULL OR b.deleted = false)")
     List<BookEntity> findAllWithMetadata();
 
     // Only ToOne paths in EntityGraph; collections (authors, categories, moods, tags, shelves, bookFiles) loaded via @BatchSize.
@@ -116,9 +117,10 @@ public interface BookRepository extends JpaRepository<BookEntity, Long>, JpaSpec
     @Query("SELECT b FROM BookEntity b WHERE b.library.id = :libraryId AND (b.deleted IS NULL OR b.deleted = false)")
     List<BookEntity> findAllForDuplicateDetection(@Param("libraryId") Long libraryId);
 
-    // Only ToOne paths in EntityGraph; collections (authors, categories, moods, tags, shelves, bookFiles) loaded via @BatchSize.
-    @EntityGraph(attributePaths = {"metadata", "metadata.comicMetadata", "library"})
-    @Query("SELECT b FROM BookEntity b WHERE b.library.id IN :libraryIds AND (b.deleted IS NULL OR b.deleted = false)")
+    // Nested to-one path metadata.comicMetadata is not honoured by @EntityGraph (produces no join,
+    // leaving comicMetadata lazy and causing an N+1 SELECT per book) — fetched explicitly instead.
+    // collections (authors, categories, moods, tags, shelves, bookFiles) loaded via @BatchSize.
+    @Query("SELECT b FROM BookEntity b LEFT JOIN FETCH b.metadata m LEFT JOIN FETCH m.comicMetadata JOIN FETCH b.library WHERE b.library.id IN :libraryIds AND (b.deleted IS NULL OR b.deleted = false)")
     List<BookEntity> findAllWithMetadataByLibraryIds(@Param("libraryIds") Collection<Long> libraryIds);
 
     @EntityGraph(attributePaths = {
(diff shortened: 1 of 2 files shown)


nfvelten:
Addressed the test coverage nitpick: added a `findAllWithMetadataByLibraryIds` regression test and switched all four statement-count assertions to a fixed upper bound instead of one derived from `bookCount`.

Skipped the `@Transactional(readOnly = true)` suggestion: Spring Data JPA repository proxies already wrap every query method, custom `@Query` ones included, in a read-only transaction by default (via `TransactionalRepositoryProxyPostProcessor`), and none of the other methods in this file declare it either. Adding it only to these two would be redundant and inconsistent with the rest of the interface.

nfvelten:
Confirmed and fixed the major finding: `libraryPath` is nullable (`@JoinColumn` without `nullable = false`), and `BookEntity.getFilePath()` already treats a null `libraryPath` as valid state, so the prior entity graph must not have dropped those books. Switched both `findAllWithMetadataByIds` and `findAllWithMetadataPage` from `JOIN FETCH b.libraryPath` to `LEFT JOIN FETCH b.libraryPath`, and added a regression test per method persisting a book with `libraryPath = null`.

imnotjames:
Please match the PR template.

nfvelten:
@imnotjames done, description rewritten to the template. Also ran `just api check` and `just ui check` on the branch, both green.

PR #2581, opened 2026-09-11
chore(i18n): update translations from Crowdin

imnotjames:


imnotjames:
Oh no now I'm doing it.

PR #2628, opened 2026-09-13
fix(books): re-allow deleting a primary file when alternatives exist

### LINKED ISSUE #2455: Deleting the primary book format fails with 400 when other formats exist

alexhb1:
## Description

Allows a primary book file to be deleted, removing the unnecessary guard restricting this. 

## Linked Issue

Fixes #2455 

## Changes

- Removes the `validateAdditionalFile` check from deleting a file
- Adds a check that remaining files for a book are a valid book, i.e. making sure a supplementary file isn't accidentally treated as a book file. 
- Added regression test

## Manual Testing Steps

Delete a primary file when alternatives also exist, confirm this works and a valid book file is replaced as primary. 

## Screenshots (Optional)

N/A

N/A

## AI Disclosure

Regression test and validation with codex. 

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved book file selection so fallback behavior chooses the oldest compatible book-format file instead of a non-book file.
  - Deleting a book’s primary file is now supported when another readable book file is available; the readable alternative is promoted automatically.
  - Supplementary file deletion now completes successfully while preserving the readable book file as primary.


backend/src/main/java/org/booklore/service/file/AdditionalFileService.java
```diff
@@ -59,7 +59,6 @@ public void deleteAdditionalFile(Long bookId, Long fileId) {
 
         BookFileEntity file = fileOpt.get();
         BookEntity book = file.getBook();
-        validateAdditionalFile(file, book);
```
imnotjames, line 62: If we are removing this we should move this out of the `AdditionalFileService` & stop using the `additionalFileRepository` for it, right?

alexhb1: I guess so yeah. depends on if it needs to happen here or a separate PR

alexhb1:
Think it's easier to merge #2621 first and then revisit this

PR #2654, opened 2026-09-16
feat(metadata): add Perrypedia metadata provider

hakan42:
## Description

Adds a Perrypedia (perrypedia.de) metadata provider for Perry Rhodan (classic + Neo) and Atlan Heftromane — a long-running German pulp-SF serial that no existing provider covers. Perrypedia has no Semantic MediaWiki / structured query API, and the infobox template differs per series (`Roman Zyklus 42` for classic, `Handlungszusammenfassung Neo Staffel 38` for Neo, `Handlungszusammenfassung Atlan <cycle name>` for Atlan). A single `action=parse` request gets redirect resolution, raw wikitext, and rendered infobox HTML together in one call. Title/subtitle/author/issue number/publication date come from the wikitext params via a brace-depth-aware parser (the template name itself isn't regexable). The story cycle ("Zyklus") isn't a literal wikitext param — it's expanded from the cycle number by the wiki itself — so that one field is read from the rendered infobox HTML instead, via jsoup.

## Linked Issue

Relates to https://github.com/orgs/grimmory-tools/discussions/2643 ("Perrypedia Metadata Provider"). This is a discussion, not yet an accepted issue — opening this PR ahead of that per my own judgment call, not because it's been accepted. Happy to hold or close if the maintainers would rather see traction there first.

## Changes

- `PerrypediaParser` + `InfoboxWikitextParser`, registered the same way every other provider is (enum, parser map, settings).
- `perrypediaId` — an identifier-only field like `comicvineId` (no rating), wired through every place another single-ID provider already touches: save/edit, metadata refresh, bookdrop, duplicate detection, sidecar files, and the matching frontend bits.
- `PerrypediaParser` also implements `DetailedMetadataProvider`, reusing the same `Quelle:<id>` lookup that already backs the primary search path, so the metadata search results dialog's fetch-full-details-on-click flow works for Perrypedia results too (previously only GoodReads/Amazon/Audible/Comicvine had this).
- Migration for `perrypedia_id` / `perrypedia_id_locked`.
- Cover thumbnail fetched from the same infobox HTML (widest `srcset` candidate), no extra request needed.
- An interrupted fetch (e.g. task cancellation) propagates correctly instead of being swallowed and silently retried as a search fallback — matches how most other parsers in this codebase already handle it.
- Tests using real wikitext captured from live Perrypedia articles for all three series families, plus a regression test for a search that matches nothing real.

## Manual Testing Steps

1. Enable Perrypedia in Settings > Metadata, search/fetch metadata for a real PR classic, PR Neo, and Atlan issue, confirm title/author/series/cover all populate correctly for each series family.
2. Edit the Perrypedia ID by hand and via the search-result copy icon, save, reload the book, confirm it persisted (caught and fixed a real bug here — see below).
3. Re-fetch metadata on a book that already has a Perrypedia ID, confirm the refresh path applies it and existing locks are respected.
4. Confirmed against a real database directly (not just the UI) that `perrypedia_id`/`perrypedia_id_locked` round-trip correctly.

This surfaced two real bugs during testing, both fixed in this branch: series name was initially pulling the whole franchise name instead of the actual story cycle (noticed from the value shown in the UI), and editing the Perrypedia ID silently didn't save at all (found by editing, saving, reopening the book, and seeing it gone — traced to a save-path helper with a hardcoded per-field list that had never had this field added to it).

Automated checks run on this exact branch before opening: backend `./gradlew check` — 3967 tests, 0 failures, 0 errors; frontend `typecheck`/`lint`/`lint:styles`/`test` — all clean, 1722 tests passed, 0 failed.

## Screenshots (optional)

image

image

image

## Additional Context

I'll do my best to keep this PR updated as it goes through review.

A few things deliberately deferred/out of scope for this PR, happy to discuss:
- Cover fetching gets a thumbnail-sized image only (widest available `srcset`); fetching the original upload would need a follow-up request.
- Real-world filename/title conventions for Neo and Atlan issues (`PRN389` vs `PR Neo 389`, `A800` vs `Atlan 800`, etc.) haven't been surveyed against a large real-world library — the current match regex covers the documented conventions but could use more field data.
- No per-request HTTP timeout on Perrypedia's requests, beyond the shared client's 10s connect timeout. Real gap, but 12 of this codebase's 14 metadata parsers share it (including `RanobeDbParser`, which this one's rate limiter deliberately mirrors) — a per-parser patch here wouldn't close the actual risk. Better fixed once, centrally, on the shared `HttpClient` bean, as a separate change.

coderabbitai's pre-merge checks flag a "Docstring Coverage" warning (functions touched by this diff are ~13% documented against an 80% threshold). Deliberately not chasing this: sampled per-method javadoc coverage on three comparable, already-merged provider classes (`RanobeDbParser`, `ComicvineBookParser`, `AudibleParser`) and found ~0% between them — the 80% bar isn't this project's actual convention anywhere else, and writing docstrings on ~90 functions just to satisfy a generic bot threshold would be inconsistent with every comparable class already in the codebase.

## AI Disclosure

Built with Claude Code doing most of the actual typing — implementation, tests, and the up-front research into how Perrypedia's API/pages actually work. I reviewed what it wrote, ran everything against a real instance and a real database myself, and both bugs mentioned above were things I found by using the feature, not things Claude Code flagged on its own. I know what this touches and why.

@coderabbitai's automated review suggestions on this PR were also reviewed and addressed across two rounds: fixed 4 real issues (a missing `Perrypedia` case in the per-field provider-enabled check that silently defaulted to always-enabled, request URIs built without percent-encoding that could break on titles containing `&`/`#`/`+`, a missing length-truncation on one code path that every sibling code path already had, and a swallowed interruption that should propagate as cancellation) — skipped the rest after checking each against the actual code and finding they didn't apply, or that they described a systemic gap already present across most of the codebase rather than something specific to this PR (details in the corresponding commit messages).

## Checklist


hakan42:
@coderabbitai rescan

hakan42:
@coderabbit help

alexhb1:
Hi, please make sure there’s an issue created for any feature requests before submitting. We’re not looking for any new metadata providers right now due to how convoluted the process is. Once we have something more suitable in place, e.g a plugin architecture, we can start promoting some suggestions to issues and accepting PRS.

hakan42:
@alexhb1 I actually created a "discussion". Who would then be responsible for creating the issue from it? Would I have any agency in it?

alexhb1:
If the feature request gets enough traction as a discussion, we’ll promote them as issues and open for PRs. Right now we want to massively clean up the metadata provider system before considering any more, adding Apple Books and OpenLibrary in the latest update was way too convoluted for us to keep doing that for new sources.

PR #2684, opened 2026-09-18
fix(audiobook): sum folder track durations

### LINKED ISSUE #2672: Audiobook duration only lists time of first audio file
### Discussed in https://github.com/orgs/grimmory-tools/discussions/1738

<sup>Originally posted by **jeypawlik** June 13, 2026</sup>
### What went wrong?

I've re-made my Audiobook library several times now, trying to get the files to read correctly. When I tried "per folder" many files were out of order, sometimes not added to the root book at all, or added as it's own separate book. Merging books didn't seem to solve this issue.

I re-made the library to be "per file" while still keeping my audiobook files within a folder per book. This seemed to work better and the files are in order and playing correctly. However, the "duration" on the main page for the audiobook is set to only read the first audio file (it doesn't seem to grab chapters in per file for some reason?).

So if I have Example Book, and Example Book has 3 audio files in the folder:
- The correct file size is showing up (and wasn't showing up in the per folder option)
- The first audio file is 7h 26m, but the whole book is 14h 15m
- The "Duration" reads as 7h 26m

Chapters don't seem to generate for per file organization on multi-file books, I wonder if that's part of the issue? The issue only seems to be on those books. I couldn't find anyone else with this issue in the Github here so I thought I'd flag it.

### How can we reproduce it?

1. Make a new audiobook library set to "per file" organization
2. Add book, with multiple audio files (each file labelled 01, 02, 03)
3. Scan library
4. Book shows up but duration reads as the duration time of the first audio file only and not the whole book

Stable

### Your setup

I'm using Grimmory 11.8-noble through Docker on my UGreen NAS and accessing through Firefox on my Windows 11 desktop.

The issue:
image
image

But not on single-file books:
image
image

### Before submitting

wyi184246-creator:
## Description

Fix folder-based multi-file audiobooks reporting only the first track's duration. The persisted audiobook duration now sums every readable track in the folder while retaining the first track's technical metadata.

## Linked Issue

Fixes #2672

## Changes

- Sum precise millisecond durations for every track in folder-based audiobooks, then round once when persisting seconds.
- Add a lightweight duration-only extractor for subsequent tracks.
- Preserve the existing duration for empty folders and skip unreadable or non-positive tracks.
- Safely retain a `null` duration when no track provides a valid value.
- Add regression coverage for fractional-duration aggregation, normal aggregation, and each fallback path.

## Manual Testing Steps

1. Ran `AudiobookProcessorTest` before the original fix and confirmed that the total-duration assertion failed.
2. Added a fractional-duration regression test with 10.75s and 20.75s tracks; confirmed the old per-track truncation produced 31s instead of the expected 32s.
3. Updated duration handling to sum milliseconds and round once, then confirmed the precision regression test passed.
4. Added and ran edge-case tests for a missing first-track duration, an empty folder, unreadable and non-positive later tracks, and no valid duration.
5. Ran the related audiobook processor, extractor, reader metadata, and library rescan test classes successfully.
6. Ran the full backend suite on Windows before the review follow-up: 3,946 tests executed; 8 Windows path/POSIX-specific tests failed outside the changed area. The targeted regression test also passed under WSL/Linux.
7. Ran frontend typecheck, ESLint, and Stylelint successfully. Frontend build and tests were blocked locally by system virtual-memory exhaustion; this PR does not change frontend files. GitHub CI later completed the backend, frontend, CodeQL, and packaging checks successfully.

## Additional Context

Folder-based audiobook playback already calculates track durations dynamically. This change fixes the persisted duration used by list and detail views without summing separate `BookFileEntity` records, which may represent alternative formats rather than audiobook parts.

## AI Disclosure

Codex - assisted with root-cause analysis, implementation, regression-test drafting, verification, review follow-up, and PR preparation. The submitter reviewed the changes before merge.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved duration accuracy for folder-based audiobooks by totaling the precise lengths of all available tracks.
  * Preserved reliable fallback durations when track metadata is missing, invalid, or unavailable.
  * Rounded fractional track durations more accurately when displaying audiobook length.


backend/src/main/java/org/booklore/service/metadata/extractor/AudiobookMetadataExtractor.java
```diff
@@ -167,6 +167,21 @@ public BookMetadata extractMetadata(File audioFile) {
...
     }
 
+    public Long extractDurationSeconds(File audioFile) {
+        try {
+            AudioHeader header = AudioFileIO.read(audioFile).getAudioHeader();
+            if (header == null) {
+                return null;
+            }
+
+            long durationSeconds = (long) header.getPreciseTrackLength();
```
alexhb1: This rounds down to whole seconds. pre-existing from before but it would be good to get this corrected to use precise millisecond data.

wyi184246-creator: Thanks — fixed in 1e1c0f8. Folder track durations are now read and summed in milliseconds, then rounded once when persisted to the existing durationSeconds field. I also updated single-file extraction to round the precise value and added a regression test covering fractional track durations.

alexhb1 (approved):
Looks good, thanks

